PR #3, opened 2021-08-22
run tests in github actions

kd7lxl:


.github/workflows/test.yaml
```diff
@@ -0,0 +1,32 @@
...
+          # the test fails with these present
+          rm -fv \
+            tests/images/AnyTone_778UV.img \
+            tests/images/CRT_Micron_UV.img \
+            tests/images/Kenwood_TK-3180K2.img \
+            tests/images/Kenwood_TK-8180.img \
+            tests/images/Kenwood_TM-D710G_CloneMode.img \
+            tests/images/Kenwood_TM-D710_CloneMode.img \
+            tests/images/Midland_DBR2500.img \
+            tests/images/Retevis_RT95.img
```
kd7lxl, line 31: This is some of my best work.

kk7ds:
This is cool, but if you could just install the future module, you should be able to get those without having to nuke the images. Obviously want to try to avoid that if possible.

Also, ideal to run with tox if possible because it constructs a specific environment and makes it easier to reproduce locally for people. That commit I linked to had a couple of tweaks to make that work.

kk7ds:
If you rebase this on master, I think it'll pass now.

kd7lxl:
> If you rebase this on master, I think it'll pass now.

The test result is here:
https://github.com/kd7lxl/chirp/runs/3395571267?check_suite_focus=true

Since it's a fork, I don't think it'll report build status here until it's approved. After it's merged, fork PRs will show up like this:
https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

kk7ds:
```unit: commands succeeded
  driver: commands succeeded
  style: commands succeeded
  congratulations :)
```

PR #4, opened 2021-08-23
run tests on all platforms

kd7lxl:


.github/workflows/test.yaml
```diff
@@ -8,17 +8,14 @@ jobs:
...
-          curl https://bootstrap.pypa.io/pip/2.7/get-pip.py --output get-pip.py
-          python get-pip.py
-          pip --version
-          python -m pip install --upgrade --no-deps --force-reinstall lxml
           python -m pip install \
-            future \
-            pep8 \
             pyserial \
             tox
+      - run: rm chirp/drivers/generic_xml.py
```
kk7ds, line 24: I would rather just remove it from the tree than have the test stuff silently behave differently here versus locally. I have a commit on the py3 branch to do it, I'll pull it over.

kk7ds:
Running the tests on the master branch on windows requires using the old test runner because the setup.py is specific for py2exe. Never tried to run them on macos, but the same problem may exist with the legacy py2app stuff. Windows tests on the py3 branch should work the same as linux because that's all modernized.

kd7lxl:
> Running the tests on the master branch on windows requires using the old test runner

How do you do that? Use `python tests/run_tests.py` instead of `tox`?

kk7ds:
> How do you do that? Use `python tests/run_tests.py` instead of `tox`?

Yup, or run_tests.bat in windows. In either case, you have to have all the deps installed manually of course. Personally, I'm not sure it's worth running the tests on the other platforms in master, as I'm not sure it's ever caught anything platform-specific (and we don't have good enough coverage anyway). Running them on py3 should be easy though, and if we're going to catch up that branch and keep it close, then we'd get some tangential coverage that way.

kk7ds:
Actually, I've got most of the build stuff pulled out at this point anyway, so give me some time this week and I'll see if I can extricate the rest and make setup.py more normal in the tree. That'll make it possible for people to use tox on windows anyway.

PR #5, opened 2021-09-21
Don't die if GTK Mac integration is missing

mfncooper:
If GTK Mac integration is missing, Chirp dies when it tries to log this
fact. The code is designed to handle the absence of this capability and
simply omit the Mac-specific features, so handle the logging gracefully
and carry on.

Fixes #8845

### DIFF
--- a/chirp/ui/mainapp.py
+++ b/chirp/ui/mainapp.py
@@ -2212,14 +2212,14 @@ of file.
         try:
             import gtk_osxapplication
             macapp = gtk_osxapplication.OSXApplication()
-        except ImportError:
+        except ImportError as e:
             pass
 
         # for gtk-mac-integration >= 2.0.7
         try:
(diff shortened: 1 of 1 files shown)


kk7ds:
Note that I'm going to avoid pushing this to the public hg tree since it's not really a user-facing change and I don't want to emit a build just for this. But it'll go the next time I push something else.

mfncooper:
It was reported in #8845 by a couple of people as preventing them from starting Chirp on Mac, so I'm not sure what you mean by not user-facing. Still, hopefully it won't be too long before something else comes along. Thanks.

kk7ds:
Oh, I didn't even look at the bug, I was assuming you were trying to run unit tests on mac or something. Running without that module won't yield a very successful app on macos, but that's fine I'll push it now.

PR #6, opened 2021-09-26
Fixes to wxPython UI

mfncooper:
A few fixes that get CSV file editing working.

### DIFF
--- a/chirp/wxui/developer.py
+++ b/chirp/wxui/developer.py
@@ -316,7 +316,8 @@ class ChirpRadioBrowser(common.ChirpEditor):
                                          self._radio.MODEL),
                               self._radio._memobj, pd)
         pd.Destroy()
-        self._treebook.ExpandNode(0)
+        if self._treebook.GetPageCount():
+            self._treebook.ExpandNode(0)
 
     def page_selected(self, event):
         page = self._treebook.GetPage(event.GetSelection())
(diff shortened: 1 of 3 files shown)


kk7ds:
Before I go digging into this, you said on the mailing list that you wanted to keep the PR for the UI fixes separate from the icomciv porting (which I think is a good idea) but this seems to have everything squashed into one.

Did you change your mind or accidentally push the latter in here? I'd definitely like to consider them separately and it should be trivial to pull them out of this one. Is that okay or is there some reason they need to be all in here?

mfncooper:
Sorry about that. I pushed the button to create the second PR, then realised the problem, and tried to back out. Only later did I realise that GH had gone ahead and added my second lot of commits anyway. Pah.

There are 2 individual commits:
* "Update menu item accessor method" fixes the breakage under Phoenix with a trivial renaming. It'll still work under wxPython 2.8.6.0 or later.
* "Some small UI fixes" has a few very minor changes that fix a some exceptions I came across.

After that, there's a bunch of commits, but only one is new. The "Migrate icomciv for Python 3" is my changes to that module to get it working under Python 3. The rest of the commits are just moving over the changes to that module from the master branch, to get the py3 branch up to date with respect to master. I wanted to ensure that those commits were retained in their original state, so that we have the proper commit history, and to ensure that they were separate from my changes for Python 3, rather than squashing them into one.

All 3 of my own commits are independent of each other, and all modify different files. Feel free to apply pieces in whichever order you're comfortable with, or let me know if you want me to change anything.

kk7ds:
Okay, what I'd really prefer is for you to force-push this branch with just the wx bits. Is there some reason you can't do that? I can pull this down and cherry-pick, but...

mfncooper:
Done.

kk7ds:
Thanks!

PR #7, opened 2021-09-30
Update icomciv

kk7ds:


kk7ds:
@mfncooper I think this looks right...yeah?

mfncooper:
Yes it does. You're seeing the same thing I was seeing with "authored" and "committed", but yes, those are the right commits. Thanks!

kk7ds:
Okay, authored and committed are how git handles the original author of a patch vs. the person that put it into the tree. In other systems (like gerrit) it's common to have those split if multiple people push a single commit. From what I can see here, this looks right and preserves proper ownership.

PR #14, opened 2021-12-21
update callable interface for 3.10 compatibility

brett:


goldstar611:
This PR fixes DeprecationWarning
```
>>> import collections
>>> collections.Callable
__main__:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
<class 'collections.abc.Callable'>
```

PR #16, opened 2022-03-04
Update es_ES.po

MELERIX:


locale/es_ES.po
```diff
@@ -695,7 +694,7 @@ msgstr "Código DTCS"
 #: ../chirpui/memedit.py:223 ../chirpui/memedit.py:878
 #: ../chirpui/memedit.py:933 ../chirpui/memedit.py:1060
 msgid "DTCS Pol"
-msgstr "DTCS Pol"
+msgstr "Política DTCS"
```
goldstar611, line 697: Hola, mi espanol no es tan bueno, pero en este linea, `Pol` debe estar `polaridad`, no?
https://chirp.danplanet.com/projects/chirp/wiki/MemoryEditorColumns#DTCS-Pol

Si ves pagina 50 de https://www.manualslib.es/manual/114361/Icom-Ic-V80.html, busca `polaridad DTCS`

--

In case you speak English :P, I think that `polaridad` is better. Everything else looks good to me.

MELERIX: yes, fixed, thanks :P

kk7ds:
(moving the conversation here, where it is relevant)

@MELERIX which build (platform) are you looking at that doesn't have the updates?

MELERIX:
in Windows version, for example I have downloaded the installer and zip version to check if there is a diference but the same issue happens in both, and the Spanish translation is still the older one in current Chirp builds.

kk7ds:
Which build exactly? From the .zip file:
```
$ unzip chirp-daily-20220604-win32.zip locale/es_ES/LC_MESSAGES/CHIRP.mo
Archive:  chirp-daily-20220604-win32.zip
  inflating: locale/es_ES/LC_MESSAGES/CHIRP.mo
$ msgunfmt locale/es_ES/LC_MESSAGES/CHIRP.mo | head
msgid ""
msgstr ""
"Project-Id-Version: CHIRP\n"
"Report-Msgid-Bugs-To: \n"
"PO-Revision-Date: 2022-03-04 00:24-0300\n"
"Last-Translator: MELERIX\n"
"Language-Team: \n"
"Language: es_ES\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
```

Which looks to me to be the updated file...

MELERIX:
with chirp-daily-20220604-win32 and for example if you open it and select Spanish file in the settings and restart Chirp to apply changes, then you will notice that Chirp is still is using old Spanish strings or some settings as were not translated and still being showed in english.

MELERIX:
ok, I have verified again and seems some string does not have translation yet, that's way these string are still in english only :P

MELERIX:
done: https://github.com/kk7ds/chirp/pull/24

PR #17, opened 2022-03-04
Update snap/snapcraft.yaml

goldstar611:
- Build on all architectures (works on the pi yay!)
- Update in-app version so it doesn't say 0.3.0dev
- Add desktop entry for Chirp
- Fix a localization build issue where `gettext` was missing

After fact, I've realized I've broken [the guidelines](https://chirp.danplanet.com/projects/chirp/wiki/DevelopersProcess#Patch-Guidelines). Let me know if you need me to fix up the commit message. Thanks!

https://snapcraft.io/chirp-snap

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -9,8 +9,8 @@ grade: stable
 confinement: strict
 base: core18
 
-architectures:
-  - build-on: amd64
+#architectures:
+#  - build-on: amd64
 
 # This is a python2 GTK2 Application:
 # https://snapcraft.io/docs/gtk2-applications
@@ -42,6 +42,7 @@ hooks:
 apps:
   chirp-snap:
     command: bin/chirpw
+    desktop: share/applications/chirp.desktop
     environment:
       # Fallback to XWayland if running in a Wayland session.
       DISABLE_WAYLAND: 1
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -68,11 +69,16 @@ parts:
     source: https://github.com/kk7ds/chirp.git
     override-pull: |
       snapcraftctl pull
-      # invalid snap version "chirp-daily-20220219+git1.3e607109": cannot be longer than 32 characters (got: 34)
+      # snap version cannot be longer than 32 characters
       LATEST_COMMIT_DATE=$(git log -1 --date=format:"%Y%m%d" --format="%ad")
       LATEST_COMMIT_SHORT=$(git rev-parse --short HEAD)
-      version="chirp-daily-${LATEST_COMMIT_DATE}+${LATEST_COMMIT_SHORT}"
-      snapcraftctl set-version "$version"
+      CHIRP_VERSION="daily-${LATEST_COMMIT_DATE}+${LATEST_COMMIT_SHORT}"
```
kk7ds, line 75: I didn't really pay attention to this earlier, but this will break the automatic update checking routines. I'd really like to figure out a better way to hook the building of stuff like this from the main build system so that the versions line up, if possible.

snap/snapcraft.yaml
```diff
@@ -9,8 +9,8 @@ grade: stable
 confinement: strict
 base: core18
 
-architectures:
-  - build-on: amd64
+#architectures:
+#  - build-on: amd64
```
kk7ds, line 13: Maybe best to just remove these to avoid confusion?

snap/snapcraft.yaml
goldstar611: After a few days of thinking on it, I think I got it figured out.
1. I'll set the in app version back to YYYYMMDD of the build date.
2. I'll change the github action that syncs the repo from hourly to on demand.
3. I'll email you a github secret token and github.com url you can use to trigger the sync, which will then trigger the snap build.

I think this should produce matching YYYMMDD versions.

PR #19, opened 2022-03-24
Added requirements.txt and requirements_dev.txt

polyglot-jones:
which represent the modules I had to install in order to get the unit tests to run for me.


goldstar611:
Hi,

Did you generate the requirements.txt using pip freeze? I don't think that `pywinauto` or `pywin32` can be installed on either macOS or Linux operating systems and I have not seen CHIRP reference either package before.

kk7ds:
pywin32 won't be installable on other platforms, so please remove that. Or maybe just document it elsewhere?

polyglot-jones:
Ah, gotchya. Done.

run_all_tests.bat
```diff
@@ -1,3 +1,3 @@
 @echo off
-python tests\run_tests.py
-python tools\cpep8.py
```
kk7ds, line 3: Did you mean to make these changes? I don't see them mentioned in the commit message, so maybe they were accidental? I'm not a windows user, so I don't know what "py -2" does (aside from the obvious guess). Is that definitely a thing for people with python 2.7 installed?

polyglot-jones: Oh, frack. Yeah, that was just for me. (FYI: I think I actually have a better way now, going thru a VM that has only Python2 installed.)

PR #21, opened 2022-04-27
[PATCH] [BF-T1] Rework driver to use shared common code

rozhuk-im:


kk7ds:
Hi, this is nice refactoring, thanks. Can I assume you've tested thoroughly (especially cloning in and out) with real radios for the affected models? I don't own them so I can't do my own testing...

rozhuk-im:
Tested with:
- BF-T1
- UV-82
- UV-3R

PR #22, opened 2022-04-27
Anytone.py -- (Further) Descriptive changes 

polyglot-jones:
- Added more explanatory comments
- Extracted more "magic" numbers and strings into named constants
- Removed the subclass for the Powerwerx DB750X radio, which will shortly have its own driver (based on this one, hence improving the comments here first).

### DIFF
--- a/chirp/drivers/anytone.py
+++ b/chirp/drivers/anytone.py
@@ -30,19 +30,24 @@ from chirp.settings import RadioSettingGroup, RadioSetting, RadioSettings, \
 
 LOG = logging.getLogger(__name__)
 
-BLOCK_SIZE = 0x10
+BLOCK_SIZE = 0x10  # 16 bytes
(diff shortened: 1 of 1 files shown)


chirp/drivers/anytone.py
```diff
@@ -733,14 +753,3 @@ class PolmarDB50MRadio(AnyTone5888UVRadio):
...
-
-@directory.register
-class PowerwerxDB750XRadio(AnyTone5888UVRadio):
-    """Powerwerx DB-750X"""
-    VENDOR = "Powerwerx"
-    MODEL = "DB-750X"
-    _file_ident = ["DB-750X"]
-
-    def get_settings(self):
-        return {}
```
kk7ds, line 746: Why are you removing this? It will break anyone that uses this subclass and has files with that model in the metadata.

polyglot-jones:
This PR is moot.

PR #23, opened 2022-06-03
Add Chinese zh_CN Localization

DuckSoft:
Carefully translated, compiled, tested and reviewed on my laptop.
Thank you for this great project! @kk7ds


kk7ds:
You were missing the entry in mainapp so it was selectable through the UI and not just via environment. Looks okay to me now, but confirm it still works on your system (and that you can select it through View->Change Language) and then I'll merge.

Thanks!

DuckSoft:
Ah yes... I missed that mainapp entry while re-applying changes from mercurial repo to git repo. Thanks for the fix!

After rebuild from git repo it works:
图片

PR #25, opened 2022-07-07
Py3

sv1:


eshattow:
print "print x" statements need to be converted to function calls "print(x)"

sv1:
I know but at least chirp is running and working for some models.

kk7ds:
This causes the relevant test jobs to fail:

```
$ tox -e style,py3driver
   ...
ERROR:   style: commands failed
ERROR:   py3driver: commands failed
```

sv1:
The errors are irrelevant to the patch. Please accept the patch and motivate people the work on py3!

kk7ds:
I really appreciate this work, but it would really be better if you could convert one driver at at a time. Otherwise this PR is going to grow to touch basically everything and be very difficult to review. There are many other details involved in a driver that aren't really well covered by the tests and thus we've had the best success converting one driver, testing it in the real world, and then moving on to the next one. If this was as easy as running 2to3, it would have been done long ago, but it's pretty common to find issues when you actually use the radio.

There are lots of examples in the history from the drivers that do work, but here's one:

https://github.com/kk7ds/chirp/commit/6ae5def58d4e2aa4603296cabd1e8b317e81b7be

chirp/drivers/ap510.py
```diff
@@ -242,7 +242,7 @@ def set_smartbeacon(self, d):
 
 class AP510Memory20141215(AP510Memory):
     """Compatible with firmware version 20141215"""
-    ATTR_MAP = dict(AP510Memory.ATTR_MAP.items() + {
+    ATTR_MAP = dict(AP510Memory.ATTR_MAP.items() | {
```
goldstar611, line 245: I haven't tried this but I don't think this works as expected. Assuming `ATTR_MAP` is a dictionary:

```python
>>> ATTR_MAP = {"a": "b"}

>>> ATTR_MAP.items() | {"b":"c"}
{'b', ('a', 'b')}

>>> dict(ATTR_MAP.items() | {"b":"c"})
  File "<stdin>", line 1, in <module>
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

chirp/drivers/ft450d.py
```diff
@@ -304,8 +306,8 @@ class FT450DRadio(yaesu_clone.YaesuCloneModeRadio):
         struct mem_struct current;
 
     """
-    _CALLSIGN_CHARSET = [chr(x) for x in range(ord("0"), ord("9") + 1) +
-                        range(ord("A"), ord("Z") + 1) + [ord(" ")]]
+    _CALLSIGN_CHARSET = [chr(x) for x in itertools.chain(range(ord("0"), ord("9") + 1),
```
goldstar611, line 309: I don't understand the need for complication here (I know you didn't originate this code). I suggest using a literal here

```python3
_CALLSIGN_CHARSET = [
    '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 
    'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 
    'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 
    ' ']
```

chirp/drivers/kenwood_live.py
```diff
@@ -84,7 +84,10 @@ def command(ser, cmd, *args):
 
     result = ""
     while not result.endswith(LAST_DELIMITER[0]):
-        result += ser.read(COMMAND_RESP_BUFSIZE).decode()
+        try:
+            result += ser.read(COMMAND_RESP_BUFSIZE).decode()
+        except:
```
goldstar611, line 89: I recommend never using "bare except". Something is definitely wrong here.

chirp/drivers/leixen.py
```diff
@@ -226,7 +226,7 @@
                    ]
 
 MODES = ["NFM", "FM"]
-WTFTONES = map(float, xrange(56, 64))
+WTFTONES = list(map(float, range(56, 64)))
```
goldstar611, line 229: No need for complication here, I suggest

```python3
[float(x) for x in range(56, 64)]
```

chirp/drivers/rfinder.py
```diff
@@ -287,7 +287,7 @@ def __init__(self, *args, **kwargs):
 
         self._rfp = None
 
-    def set_params(self, (lat, lon), miles, email, password):
+    def set_params(self, lat, lon, miles, email, password):
```
goldstar611, line 290: This does not work as expected. You've changed the signature from having 4 arguments (Not including `self`) to having 5 arguments. 

I don't see any updated callers reflect this change.

goldstar611 (commented):
There's some good stuff in here, but there's also some thing that need work. 

I'm no authority here but I suggest seeing what Dan says about such tree-wide changes. I think he has some reservations about the code not operating as it did in python 2 which could lead to radio failure.

Have any of these changes been tested? Which radio(s) are you using?

sv1:
Hello Tony,

see inline comments...


On Thu, 2022-09-15 at 06:39 -0700, Tony F wrote:
> @goldstar611 commented on this pull request.
> 
> There's some good stuff in here, but there's also some thing that need work. 
> 
> I'm no authority here but I suggest seeing what Dan says about such tree-wide
> changes. I think he has some reservations about the code not operating as it
> did in python 2 which could lead to radio failure.
> 
> Have any of these changes been tested? Which radio(s) are you using?
> 
> > @@ -242,7 +242,7 @@ def set_smartbeacon(self, d):
>  
>  class AP510Memory20141215(AP510Memory):
>      """Compatible with firmware version 20141215"""
> -    ATTR_MAP = dict(AP510Memory.ATTR_MAP.items() + {
> +    ATTR_MAP = dict(AP510Memory.ATTR_MAP.items() | {
> 
> I haven't tried this but I don't think this works as expected. Assuming
> `ATTR_MAP` is a dictionary:
> 
> ```python
> > > > ATTR_MAP = {"a": "b"}
> 
> > > > ATTR_MAP.items() | {"b":"c"}
> {'b', ('a', 'b')}
> 
> > > > dict(ATTR_MAP.items() | {"b":"c"})
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ValueError: dictionary update sequence element #0 has length 1; 2 is required

The recommended way of joining dicts since python 3.9.0 is " | ". your code
doesn't work because you forgot .items on the second dict.


> ```
> 
> > @@ -304,8 +306,8 @@ class FT450DRadio(yaesu_clone.YaesuCloneModeRadio):
>          struct mem_struct current;
>  
>      """
> -    _CALLSIGN_CHARSET = [chr(x) for x in range(ord("0"), ord("9") + 1) +
> -                        range(ord("A"), ord("Z") + 1) + [ord(" ")]]
> +    _CALLSIGN_CHARSET = [chr(x) for x in itertools.chain(range(ord("0"),
> ord("9") + 1),
> 
> I don't understand the need for complication here (I know you didn't originate
> this code). I suggest using a literal here
> 
> ```python3
> _CALLSIGN_CHARSET = [
>     '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 
>     'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 
>     'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 
>     ' ']
> ```
> 
Agree, but the result is the same.


> > @@ -84,7 +84,10 @@ def command(ser, cmd, *args):
>  
>      result = ""
>      while not result.endswith(LAST_DELIMITER[0]):
> -        result += ser.read(COMMAND_RESP_BUFSIZE).decode()
> +        try:
> +            result += ser.read(COMMAND_RESP_BUFSIZE).decode()
> +        except:
> 
> I recommend never using "bare except". Something is definitely wrong here.
> 

I am open to suggestions. On My kenwood th-f7 during serial port speed
detection, returns some unknow utf-8 character causing chirp to abort and
through an error. With this I just continue to the next serial port speed which
is the correct one.


> > @@ -226,7 +226,7 @@
>                     ]
>  
>  MODES = ["NFM", "FM"]
> -WTFTONES = map(float, xrange(56, 64))
> +WTFTONES = list(map(float, range(56, 64)))
> 
> No need for complication here, I suggest
> 
> ```python3
> [float(x) for x in range(56, 64)]
> ```
> 
Totally agree!!

> > @@ -287,7 +287,7 @@ def __init__(self, *args, **kwargs):
>  
>          self._rfp = None
>  
> -    def set_params(self, (lat, lon), miles, email, password):
> +    def set_params(self, lat, lon, miles, email, password):
> 
> This does not work as expected. You've changed the signature from having 4
> arguments (Not including `self`) to having 5 arguments. 
> 
> I don't see any updated callers reflect this change.
> 
I really do not remember, either way this is WIP since rfinder is not enabled in
py3

Regards
Apostolos

chirp/drivers/ft450d.py
kk7ds: We do this in a lot of places to save the typing, but also to make sure we don't miss any. I know it should be trivial, but I think there's a reasonable argument for having it be generated.

Either way, best not to change fundamentals like this in a patch that purports to just be compatibility-related.

chirp/drivers/kenwood_live.py
kk7ds: Yes, this should be `except Exception` at the least. That said, the bytes/str sort of errors this is obviously meant to cover up is exactly what is tricky about these conversions and why it's not as simple as just 2to3.

IMHO, this sort of thing can't be in anything we merge here.

kk7ds:
> I'm no authority here but I suggest seeing what Dan says about such tree-wide changes.

Yeah, this is really massive at this point, and I really think it would be better to march through the drivers one at a time. If we were to merge some big conversion like this, we basically need to also tag each (affected) driver as "not tested on py3" until it is. The drivers that have already been converted are tested. Making a bunch import that seem to work but don't is just asking for user pain for anyone that tries to use this without proper warning.

> I think he has some reservations about the code not operating as it did in python 2 which could lead to radio failure.

I dunno, I was pretty adamant about keeping py2 working in this branch so that we could hope for a dual-py release until we were ready to cut over. It doesn't seem like people really have much interest or tolerance for that (especially with people proposing 3.9-and-later syntax, like this does). So, yes, it'd be nice, but at the moment, I'm more concerned with keeping the tests working on py3 and testing the drivers we claim to be "fixing" here.

chirp/drivers/th9000.py
```diff
@@ -399,8 +399,13 @@ def _checksum(data):
 def _read(radio, length):
     try:
         data = radio.pipe.read(length)
+<<<<<<< HEAD
```
kk7ds, line 402: Can you please rebase on py3 and not just keep merging into your PR? Things like this are now permanently in your PR until/unless you clean it up.

I would really prefer a PR full of consistent and coherent changes that fix specific things in a specific way. Each commit should pass tests and build on the last one.

sv1:
Let's start over!

PR #32, opened 2022-09-09
[py3] Merge with master

kk7ds:
This catches up py3 with master from over two years ago. Also includes some extra test stuff to make sure we're not regressing py3 progress when we do this.

This looks like a lot of change, but it was pretty trivial. Many
of the conflicts came from https://github.com/kk7ds/chirp/commit/d40963b3edc83440306122d824e91296e57820be
which fixed a bunch of typos. There were some test adapter changes,
some GTK UI fixes, RR canada additions, and the usual driver
maintenance.

I didn't test the GTK UI since it seems like we're abandoning keeping
that working in python2 or python3 with pygobject (which is fine).


kk7ds:
@mfncooper Do you want to review/test this before I merge? Tests all pass (plus new tests to make sure we're not regressing) and I trivially tested with wxui.

PR #35, opened 2022-09-15
[RB23] Add Retevis RB23

kk7ds:


kk7ds:
py3 failure is due to adding a new image without it being on the exclusion list for a driver that does not work on py3.

PR #40, opened 2022-09-16
Python 3 exceptions

sv1:
python 3 exceptions

### DIFF
--- a/chirp/drivers/anytone_ht.py
+++ b/chirp/drivers/anytone_ht.py
@@ -233,7 +233,7 @@ struct memory memory[200];
 def _echo_write(radio, data):
     try:
         radio.pipe.write(data)
-    except Exception, e:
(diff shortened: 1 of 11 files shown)


kk7ds:
I really don't want to give the impression that I'm trying to be difficult here, but slicing this horizontally by topic (i.e. exceptions) is going to end up being mostly the same as the other approach. As you can see, as soon as you make a single file importable by fixing the syntax, then the test harness is going to assert that the rest of the driver needs to work and you end up having to fix everything at once.

I certainly understand the desire to go methodically by theme (i.e. exceptions, print statements, str->bytes, etc) but it doesn't lend itself to incremental improvement in this case. Going driver by driver will work a lot better I think.

sv1:
I 'll be back :)

PR #44, opened 2022-09-16
Split tests and fix py3 PR actions

kk7ds:


kk7ds:
This will look like a force merge because the test names are changing.

PR #48, opened 2022-09-20
Add a tool for tracking real-world driver testing

kk7ds:
Converting a driver to pass tests on python3 is not quite the same as it actually working in the real world. We need to keep track of which drivers have been tested in the wild, separate from those that merely pass tests.

This adds a file tests/py3_driver-testers.txt, which is a simple list of the driver ID key and the person/date when it was tested. Running the "makesupported" tox target will also generate
tests/Python3_Driver_Testing.md, which will be render-able in github for easy viewing.

If you test a driver on python3, please add a line to the tracking file, run the makesupported target, and update the markdown file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,3 +8,5 @@ preferred as PRs here.
 
 When submitting PRs, please squash commits into logical units in a fashion
 similar to what is currently expected by the project on the mailing list.
+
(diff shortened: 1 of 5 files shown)


kk7ds:
@sv1 FYI

@mfncooper Should I add you in here for the ICOM 910 as of your last patch?

PR #50, opened 2022-09-21
2to3 of more radios

sv1:
2to3 of TK-760 an 760G, some BTECH, Baofeng, Wouxoun and KYD radios.

Tests with real radios is more than welcome


kk7ds:
FYI, I added a driver tracking matrix so that we can hopefully keep track of which drivers have been merely converted vs actually tested. The job will require that you update the matrix when you fix a driver, but running the "makesupported" target will do it for you, so just run that before you commit/submit:

```
tox -e makesupported -r
```

This will also give us a list of drivers for people to look at if they're considering trying the py3 branch and want to know if their driver is converted and/or tested.

chirp/drivers/baofeng_wp970i.py
```diff
@@ -73,7 +73,7 @@ def model_match(cls, data):
 
     if len(data) > 0x2008:
         rid = data[0x2008:0x2010]
-        return rid.startswith(cls.MODEL)
+        return rid.startswith(b'cls.MODEL')
```
kk7ds, line 76: This is not right. Instead of passing the definition of MODEL on cls, you're passing the bytestring `cls.MODEL`. I assume this is not failing tests because we don't have a legacy test image that requires actual detection. You need to encode `cls.MODEL` to bytes here to properly handle this, so:

`return rid.startswith(cls.MODEL.encode())`

chirp/drivers/kguv9dplus.py
```diff
@@ -445,7 +445,7 @@ def freq2short(val, min, max):
                                 (chirp_common.format_freq(_freq),
                                  chirp_common.format_freq(min),
                                  chirp_common.format_freq(max)))
-    return _freq/100000 & 0xFFFF
+    return int(_freq/100000) & 0xFFFF
```
kk7ds, line 448: Why not use `//` here like elsewhere?

chirp/drivers/tk760.py
```diff
@@ -739,26 +736,26 @@ def get_settings(self):
         basic.append(clone)
 
         # front keys
-        rs = RadioSettingValueList(KEYS.values(),
-                                   KEYS.values()[KEYS.keys().index(
+        rs = RadioSettingValueList(list(KEYS.values()),
```
kk7ds, line 739: We might should make these actually `list()`ify their parameters so we don't have to do this everywhere.

chirp/drivers/tk760g.py
```diff
@@ -813,7 +813,7 @@ def _prep_data(self):
         # building the data for the memmap
         fdata = ""
 
-        for k, v in data.iteritems():
+        for k, v in list(data.items()):
```
kk7ds, line 816: This is a nit, but I think just `data.items()` is fine here, no need to listify it first.

tests/py3_remaining.txt
```diff
@@ -104,7 +94,7 @@ Wouxun_KG-UV6.img
 Wouxun_KG-UV8D.img
 Wouxun_KG-UV8D_Plus.img
 Wouxun_KG-UV8E.img
-Wouxun_KG-UV9D_Plus.img
+### Wouxun_KG-UV9D_Plus.img
```
kk7ds, line 97: Did you mean to remove this?

kk7ds (changes requested):
Most of these are for consistency, but the cls.MODEL thing definitely needs to be fixed.

sv1:
Running 
> ```
> tox -e makesupported -r
> ```

Gives error below

````
........
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'develop'
ERROR: InvocationError for command /srv/software/sv1-chirp/.tox/makesupported/bin/python3 setup.py develop (exited with code 1)
_______________________________________________________ summary ________________________________________________________
ERROR:   makesupported: commands failed
````

File: chirp/drivers/baofeng_wp970i.py (hunk 1 above)
Comment (sv1) [reply chirp/drivers/baofeng_wp970i.py]: Done

File: chirp/drivers/kguv9dplus.py (hunk 2 above)
Comment (sv1) [reply chirp/drivers/kguv9dplus.py]: Done!

File: chirp/drivers/tk760g.py (hunk 4 above)
Comment (sv1) [reply chirp/drivers/tk760g.py]: Done!!!

File: tests/py3_remaining.txt (hunk 5 above)
Comment (sv1) [reply tests/py3_remaining.txt]: Yeap!

File: chirp/drivers/tk760.py (hunk 3 above)
Comment (sv1) [reply chirp/drivers/tk760.py:739]: I will need some help with this!

**Comment by kk7ds** [comment]
> According to this [https://setuptools.pypa.io/en/latest/deprecated/commands.html](url)
> 
> replace from distutils.core import setup with from setuptools import setup
> 
> in setup.py to work properly!

I think just having setuptools on your machine will do the job, as it works for me (and obviously does in the job). I will add setuptools to the deps in that tox target and hopefully that will fix it for you. Can you tell me what environment (os/distro, python version, etc) you were on?

Thanks for going through the motions for me :)

File: chirp/drivers/tk760.py (hunk 3 above)
Comment (goldstar611) [reply chirp/drivers/tk760.py:739]: I don't usually say this but, I think if we let these slide through, we should take the opportunity to come back around and simplify the logic, it's really going in circles here.

Take for example in the [tk760g.py driver](https://github.com/sv1/chirp/blob/8a7cf8e6a0b5a5ce0a7ab4bab5002216b2e290f4/chirp/drivers/tk760g.py#L1307) where I added in some print statements to see how full circle we actually go:

```python3
            # Here's what happens
            print KEYS
            print keys.kVOL_UP
            print int(keys.kVOL_UP)
            print KEYS.keys().index(int(keys.kVOL_UP))
            print KEYS.values()[KEYS.keys().index(int(keys.kVOL_UP))]
            print type(keys.kVOL_UP)
            # Here's what we could do instead
            print KEYS[int(keys.kVOL_UP)]
            
            vu = RadioSetting("keys.kVOL_UP", "VOL UP",
                              RadioSettingValueList(KEYS.values(),
                              KEYS.values()[KEYS.keys().index(
                                  int(keys.kVOL_UP))]))
```

```console
{161: 'Channel Up/Down', 51: 'Display character', 53: 'Home Channel', 55: 'CH down', 56: 'CH up', 57: 'Key lock', 58: 'Lamp', 59: 'Public address', 60: 'Reverse', 61: 'Horn alert', 62: 'Selectable QT', 63: '2-tone encode', 64: 'Monitor A: open mommentary', 65: 'Monitor B: Open Toggle', 66: 'Monitor C: Carrier mommentary', 67: 'Monitor D: Carrier toggle', 68: 'Operator selectable tone', 69: 'Redial', 70: 'RF Power Low', 71: 'Scan', 72: 'Scan del/add', 74: 'GROUP down', 75: 'GROUP up', 79: 'None', 80: 'VOL down', 81: 'VOL up', 82: 'Talk around', 93: 'AUX'}
0x51
81
25
VOL up
<type 'instance'>
VOL up
```

That line and many, many others would simply to just
```
            vu = RadioSetting("keys.kVOL_UP", "VOL UP",
                              RadioSettingValueList(KEYS.values(),
                              KEYS[int(keys.kVOL_UP)])
```

I'm more than happy to help but wanted to mention this before we "fix" `./chirp/settings.py`

Tony

kk7ds: I'm working on it now and I think I'm addressing this concern. I'll add you to the PR so you can look.

PR #52, opened 2022-09-24
Add updated Turkish translation

TA1AUB:
Turkish language file has been updated.


kk7ds:
Hi, can you explain what's going on here? Looks like the first two commits were accidental translation of the base (CHIRP.po) file and then the third is tweaks to the translation you recently sent, is that right? It seems like there is a lot of change here, so it would be good to understand why.

Also, the "Delete CHIRP.po" commit really can't go into the history, so it would be great if you could squash this into a single update commit to make the history clear.

Thanks!

TA1AUB:
Hello, Yes I made a mistake. But I don't know how to fix it. Please fix it.

kk7ds:
Sorry I accidentally updated this with master which closed and deleted it. New PR here: https://github.com/kk7ds/chirp/pull/55

PR #53, opened 2022-09-25
Fixed UV-3R radio

yarda:
Not the cleanest solution, but worked for me.

### DIFF
--- a/chirp/drivers/baofeng_uv3r.py
+++ b/chirp/drivers/baofeng_uv3r.py
@@ -19,7 +19,7 @@ import time
 import os
 import logging
 
-from wouxun_common import do_download, do_upload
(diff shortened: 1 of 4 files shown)


yarda:
I tested reading radio, adding new memory channel content and writing back to the radio.

yarda:
I know and agree. I just needed to program the radio, so I went the most easy way to get it working. I will try to migrate the driver from compat. But I am afraid the code will require more changes.

kk7ds:
Yeah, it definitely will require more changes, but they shouldn't be too bad. Anyway, that's cool, I just wanted to see if you'd be up for it. We can always do it later.

I'll rebase this for you in a sec and then merge.

PR #54, opened 2022-09-28
2to3 of Anytone family drivers

sv1:
2to3 of Anytone family drivers.

Real radio is more than welcome!

Thanks!

### DIFF
--- a/chirp/drivers/anytone.py
+++ b/chirp/drivers/anytone.py
@@ -381,7 +381,7 @@ class AnyTone5888UVRadio(chirp_common.CloneModeRadio,
     VENDOR = "AnyTone"
     MODEL = "5888UV"
     BAUD_RATE = 9600
-    _file_ident = ["QX588UV".encode(), "588UVN".encode()]
+    _file_ident = [b"QX588UV", b"588UVN"]
 
     # May try to mirror the OEM behavior later
     _ranges = [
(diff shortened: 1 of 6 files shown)


chirp/drivers/anytone.py
```diff
@@ -685,9 +685,10 @@ def filter(s):
                 s_ += (c if c in chirp_common.CHARSET_ASCII else "")
             return s_
 
+        welcome_msg = ''.join(list(filter(_settings.welcome)))
```
kk7ds, line 688: Why the extra list()? I don't think it should be necessary right?

chirp/drivers/anytone.py
```diff
@@ -381,7 +381,7 @@ class AnyTone5888UVRadio(chirp_common.CloneModeRadio,
     VENDOR = "AnyTone"
     MODEL = "5888UV"
     BAUD_RATE = 9600
-    _file_ident = ["QX588UV", "588UVN"]
+    _file_ident = ["QX588UV".encode(), "588UVN".encode()]
```
kk7ds, line 384: Why not use `b"QX588UV"` to be consistent with everything else? Static strings should be defined as byte strings to avoid the runtime overhead of encoding them. There are a lot of examples here, so it would be good if we can get that fixed first.

kk7ds (commented):
Thanks, this will knock out a bunch of radios :)

chirp/drivers/anytone.py
sv1: Right!

chirp/drivers/anytone.py
sv1: Done!

PR #57, opened 2022-10-01
Apply ML patches from 20221001

kk7ds:


kk7ds:
The py3 merge failure is actually just because of the new models added to unconverted files. Waiting on merge for some missing test images.

PR #59, opened 2022-10-02
Retevis radios for python3 and a UI tweak

sv1:
2to3 of Yaesu and Retevis radios

Test with real radios is more than welcome!

Thanks!


chirp/drivers/ft1d.py
```diff
@@ -1131,7 +1131,8 @@ def _get_aprs_beacons(self):
                 checksum = body[-2:]
                 body = ''.join(s for s in body[:-2]
                                if s in string.printable).translate(
-                                   None, "\x09\x0a\x0b\x0c\x0d")
+                                   str.maketrans("", "",
+                                                 "\x09\x0a\x0b\x0c\x0d"))
```
kk7ds, line 1135: I don't know what this is trying to do... Can you explain please?

chirp/drivers/ftm3200d.py
```diff
@@ -124,10 +124,10 @@ def get_features(self):
 
     def _decode_label(self, mem):
         # TODO preserve the unknown \x80-x86 chars?
-        return str(mem.label).rstrip("\xFF").decode('ascii', 'replace')
+        return str(mem.label).rstrip("\xFF")
```
kk7ds, line 127: Aren't you removing code that replaces non-ascii characters? The original use of decode to do this is probably not the best thing here, but I think you're just introducing a likely crash aren't you?

chirp/wxui/clone.py
```diff
@@ -101,7 +102,7 @@ def _add_grid(label, control):
         _add_grid('Vendor', self._vendor)
         self.Bind(wx.EVT_CHOICE, self._selected_vendor, self._vendor)
 
-        self._model = wx.Choice(panel, choices=[], style = wx.CB_SORT)
+        self._model = wx.Choice(panel, choices=[])
```
kk7ds, line 105: Why are you reverting this now?

chirp/wxui/clone.py
```diff
@@ -92,7 +93,7 @@ def _add_grid(label, control):
             ports.insert(0, last_port)
         elif not last_port:
             last_port = ports[0]
-        self._port = wx.ComboBox(panel, choices=ports)
+        self._port = wx.ComboBox(panel, choices=ports, style=wx.CB_DROPDOWN)
```
kk7ds, line 96: I can't tell what this actually changes. Can you explain?

chirp/wxui/clone.py
```diff
@@ -118,12 +119,14 @@ def _add_grid(label, control):
...
-        vbox.Add(grid, proportion=0, flag=wx.ALIGN_CENTER|wx.TOP|wx.BOTTOM,
+        vbox.Add(grid, proportion=0, flag=wx.ALIGN_CENTER_HORIZONTAL|wx.TOP,
                  border=20)
-        vbox.Add(self.gauge, flag=wx.EXPAND, border=10, proportion=1)
+        vbox.Add(self.gauge, flag=wx.EXPAND|wx.LEFT|wx.RIGHT, border=10, proportion=0)
         vbox.Add(wx.StaticLine(panel), flag=wx.EXPAND|wx.TOP, border=10)
         vbox.Add(hbox2, proportion=0, flag=wx.ALIGN_RIGHT|wx.ALL, border=10)
         panel.SetSizer(vbox)
+        self.Layout()
+        self.Center()
```
kk7ds, line 129: I don't know why you're making these changes. The clone dialog looks identical before/after them AFAICT. Can you please write more descriptive commit messages for things like this to explain what you're doing and why?

chirp/drivers/ft1d.py
sv1: As far as I can understand you are using translate() to remove these chars "\x09\x0a\x0b\x0c\x0d". But in python 3 translate needs a table. The recommended way to create the table is with maketrans(). maketrans could work with only one argument ("\x09\x0a\x0b\x0c\x0d") but I though it would be better to use the optional fields too. Am I missing something?

chirp/wxui/clone.py
sv1: This is my mistake! Sorting the models in th wx.Choice causes return of wrong index when trying to load the radio class!

chirp/wxui/clone.py
sv1: Without these changes the widgets in the clone dialog overlapping and it is a little mess, as far as I can see in two different linux systems. Also the wx.gauge is not visible without them. Currently on pyhton 3.9.2, wxpython 4.0.7 and wxwidgets 3.0.5

chirp/drivers/ftm3200d.py
sv1: I do not think you need to replace non-ascii characters in the memory label, since there is no way to put non-ascii characters in the memory label, right? The only problem if you try to encode a memory label, but I am working on this!

kk7ds: Well, mem is a bitwise object and mem.label is a chars. Since you're not converting this to byte-clean mmap, I'm not exactly sure what happens here. What do you think happens when the image, which actually has non-ascii characters in the char field, is loaded and processed here? And why aren't you removing the `\xFF` cleaning as well. Surely the same argument would hold?

This is the problem with converting radios without real testing...

So if I use this PR to load our test image, it does not render the label at all in the regular editor. In the radio browser, memory 1 shows the unfiltered character string, which has the right-filled non-ascii characters in it. I think that means this is not doing the right thing. The tests pass because we never read and assert anything about the existing image, and we probably space-pad the label when we set it so we don't notice. Load the test image locally and check it out.

chirp/drivers/ft1d.py
kk7ds: Roger I see. Not sure this is the best way to remove characters we don't want from the string, but it's what was there before, so, okay.

chirp/wxui/clone.py
kk7ds: Ah, okay. Probably better to revert that as a single commit with an explanation, but okay.

chirp/wxui/clone.py
kk7ds: Hmm, I don't see that at all on my linux machine. Python 3.9, wxwidgets 3.0.4 and wxpython 4.0.7 here. Maybe it's a theming thing? Can you post a screenshot so I can see? But, it looks the same to me so I guess it's okay.

kk7ds:
I did a little digging on the Yaesu thing and something else is going on for sure. I'd like to pull it out of this PR so we can merge the other changes. I also would prefer not to have a revert...revert...revert in in the history when we can avoid it. So with your permission I'd like to push to this PR to remove the yaesu commit, and squash the UI changes.

There's also another use of str.translate in the ft1d file that looks like it needs to be fixed, which I noticed while trying to debug. Github (ridiculously) won't let me comment on unchanged lines, but we can address that with that commit separated out.

Okay?

sv1:
> I did a little digging on the Yaesu thing and something else is going on for sure. I'd like to pull it out of this PR so we can merge the other changes. I also would prefer not to have a revert...revert...revert in in the history when we can avoid it. So with your permission I'd like to push to this PR to remove the yaesu commit, and squash the UI changes.
>

Please do so!
 
> There's also another use of str.translate in the ft1d file that looks like it needs to be fixed, which I noticed while trying to debug. Github (ridiculously) won't let me comment on unchanged lines, but we can address that with that commit separated out.
> 
> Okay?

Its fine by me!

Meanwhile I played a little with the ft1d driver and found out that if I invert the logic of the flag.valid and flag.used everything is working as expected. And I can verify that with what I can see in the .img of ft1d, ft2d, ft3d ftm3200 and ftm7250.

But how is it possible this driver to work in python2.7 ????

PR #64, opened 2022-10-10
[RB17P] Add support for Retevis RB17P radio's new features

kk7ds:
The latest RB17P firmware (v1.31) upgrades this model to 128 channels, customizable channels and adds support for channel names.

This patch adds support for these new features.

Fixes #10071

### DIFF
--- a/chirp/drivers/retevis_rb17p.py
+++ b/chirp/drivers/retevis_rb17p.py
@@ -45,7 +45,7 @@ struct {
        unknown4:1,
        bcl:2;
     u8 unknown6[2];
-} memory[30];
+} memory[128];
 #seekto 0x0810;
 struct {
     u8 unknown01:5,
@@ -67,7 +67,7 @@ struct {
        timeouttimer:2,
        unknown08:1;
     u8 alarmtype:1,
-       unknown09:3
+       voxdelay:3,
        voxlevel:4;
     u8 unknown10:2,
        backlight:2,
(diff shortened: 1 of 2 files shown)


chirp/drivers/retevis_rb17p.py
```diff
@@ -382,6 +353,12 @@ def get_memory(self, number):
             mem.duplex = int(_mem.rxfreq) > int(_mem.txfreq) and "-" or "+"
             mem.offset = abs(int(_mem.rxfreq) - int(_mem.txfreq)) * 10
 
+        for char in _nam.name:
```
goldstar611, line 356: Is `_nam.name` not guaranteed to be a string here?

It feels like it'd be more natural to write something like `mem.name = _nam.name.replace("\xFF", "").rstrip()`

kk7ds: No, it's a memobj of bitwise chars. You can str(_mem.name) here and the right thing will happen, but it's not a string itself. This pattern is typical for other drivers, for better or worse :)

That said, this is Jim's patch from the mailing list and I'm just proxying here...

goldstar611: Thanks for the explanation.

I'm probably showing my ignorance as I post comments here but I'm trying to learn the ins and outs as they show up  :)

kk7ds: No worries, glad someone's paying attention :)

PR #66, opened 2022-10-11
Make drivers python 3 friendly

sv1:
Make icx90 and radtel_t18 drivers python 3 friendly.


chirp/drivers/icx90.py
```diff
@@ -17,7 +17,7 @@
 
 import logging
 
-import icf
+from . import icf
```
kk7ds, line 20: I think this should be:

```
from chirp.drivers import icf
```

To be more consistent.

chirp/drivers/radtel_t18.py
```diff
@@ -376,7 +376,7 @@ def model_match(cls, data):
     """Match the opened/downloaded image to the correct version"""
 
     if len(data) == cls._memsize:
-        rid = data[0x03D0:0x03D8]
+        rid = data[0x03D0:0x03D8].decode('ascii', 'replace')
         return "P558" in rid
```
kk7ds, line 380: I think we should not decode to ascii here and instead check for ```b"P558" in rid``` right? We've done that everywhere else, AFAIK, why be different here?

chirp/drivers/vgc.py
```diff
@@ -345,7 +345,7 @@
                 chirp_common.PowerLevel("High", watts=50)]
 
 # B-TECH UV-50X3 id string
-UV50X3_id = "VGC6600MD"
+UV50X3_id = b"VGC6600MD"
```
kk7ds, line 348: I don't think this is right. If this radio doesn't use byte-native serial and mmap, then it should be using the string version when it inspects the stream, right?

chirp/drivers/vgc.py
```diff
@@ -549,7 +549,7 @@ def _upload(radio):
 
 def model_match(cls, data):
     """Match the opened/downloaded image to the correct version"""
-    rid = data[0x6140:0x6148]
+    rid = data[0x6140:0x6148].decode('ascii', 'replace')
```
kk7ds, line 552: I think the right way to do this is encode the model string, which we know will encode to UTF-8, instead of decode-and-discard unknown binary data. I'll push a change.

PR #72, opened 2022-10-19
More radios for python3

sv1:
This PR adds more radios available for testing in python 3 chirp.

Real radio testing is more than welcome.

Thanks!

### DIFF
--- a/chirp/drivers/kenwood_live.py
+++ b/chirp/drivers/kenwood_live.py
@@ -122,7 +122,12 @@ def get_id(ser):
             ser.baudrate = i
             ser.write(LAST_DELIMITER[0].encode())
             ser.read(25)
-            resp = command(ser, "ID")
+            try:
+                resp = command(ser, "ID")
+            except UnicodeDecodeError:
+                # If we got binary here, we are using the wrong rate
+                # or not talking to a kenwood live radio.
+                continue
 
             # most kenwood radios
             if " " in resp:
--- a/chirp/drivers/kenwood_live.py
+++ b/chirp/drivers/kenwood_live.py
@@ -65,13 +65,11 @@ LAST_DELIMITER = ("\r", " ")
 # fields, but others do.
 
 
-def command(ser, cmd, *args):
+def _command(ser, cmd, *args):
     """Send @cmd to radio via @ser"""
     global LOCK, LAST_DELIMITER, COMMAND_RESP_BUFSIZE
 
     start = time.time()
-
-    LOCK.acquire()
 
     # TODO: This global use of LAST_DELIMITER breaks reentrancy
     # and needs to be fixed.
@@ -95,9 +93,12 @@ def command(ser, cmd, *args):
     else:
         LOG.error("Giving up")
 
-    LOCK.release()
-
     return result.strip()
+
+
+def command(ser, cmd, *args):
+    with LOCK:
+        return _command(ser, cmd, *args)
 
 
 def get_id(ser):
--- a/tests/Python3_Driver_Testing.md
+++ b/tests/Python3_Driver_Testing.md
@@ -98,7 +98,7 @@
 | Kenwood_TH-D74_live_mode | [@kk7ds](https://github.com/kk7ds) | 20-Oct-2022 | Yes |
 | Kenwood_TH-D7G |  |  | Yes |
 | Kenwood_TH-F6 |  |  | Yes |
-| Kenwood_TH-F7 |  |  | Yes |
+| Kenwood_TH-F7 | [@sv1](https://github.com/sv1) | 19-Oct-2022 | Yes |
 | Kenwood_TH-G71 |  |  | Yes |
 | Kenwood_TH-K2 |  |  | Yes |
 | Kenwood_TK-2180 |  |  | Yes |
--- a/tests/Python3_Driver_Testing.md
+++ b/tests/Python3_Driver_Testing.md
@@ -156,6 +156,7 @@
 | Kenwood_TS-590SG_CloneMode |  |  |  |
 | Kenwood_TS-590S_CloneMode |  |  |  |
 | Kenwood_TS-590S_SG_LiveMode |  |  | Yes |
+| Kenwood_TS-850 |  |  | Yes |
 | LUITON_LT-316 |  |  |  |
 | LUITON_LT-588UV |  |  | Yes |
 | MTC_UV-5R-3 |  |  |  |
--- a/tests/Python3_Driver_Testing.md
+++ b/tests/Python3_Driver_Testing.md
@@ -153,6 +153,8 @@
 | Kenwood_TS-2000 |  |  | Yes |
 | Kenwood_TS-480_CloneMode |  |  |  |
 | Kenwood_TS-480_LiveMode |  |  | Yes |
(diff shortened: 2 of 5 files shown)


chirp/drivers/retevis_rt22.py
```diff
@@ -332,7 +332,7 @@ def model_match(cls, data):
 
     if len(data) == 0x0408:
         rid = data[0x0400:0x0408]
-        return rid.startswith(cls.MODEL)
+        return rid.startswith(cls._model)
```
kk7ds, line 335: Again, why duplicate that bytestring when you can easily do:

```
return rid.startswith(cls.MODEL.encode())
```
When we know the models are actual strings, a straight encode seems like the right thing to me. When they've got non-ASCII values in them, then it makes sense to have a different string, but I think in most cases you'll find they already do.

chirp/drivers/kenwood_live.py
```diff
@@ -84,7 +84,7 @@ def command(ser, cmd, *args):
 
     result = ""
     while not result.endswith(LAST_DELIMITER[0]):
-        result += ser.read(COMMAND_RESP_BUFSIZE).decode()
+        result += ser.read(COMMAND_RESP_BUFSIZE).decode('ascii', 'replace')
```
kk7ds, line 87: Why? The kenwood live protocol is entirely ASCII. I think anywhere you do this `decode('ascii', 'replace')` is just a hack to convert a bytestring into a unicode string, but with silent failures for unexpected values. It's much better to avoid this whenever possible.

tests/Python3_Driver_Testing.md
```diff
@@ -96,7 +96,7 @@
 | Kenwood_TH-D72_live_mode |  |  | Yes |
 | Kenwood_TH-D7G |  |  | Yes |
 | Kenwood_TH-F6 |  |  | Yes |
-| Kenwood_TH-F7 |  |  | Yes |
+| Kenwood_TH-F7 | [@sv1](https://github.com/sv1) | 19-Oct-2022 | Yes |
```
kk7ds, line 99: \o/

chirp/drivers/retevis_rt22.py
sv1: I did this because in PR #54 you said that static strings should be defined as byte strings to avoid the runtime overhead of encoding them.

chirp/drivers/kenwood_live.py
sv1: The decode() function by default is very strict. During probing the serial port for the correct speed some unknown character ends up in decode causing an UnicodeError. This stop the probing before reaching the correct radio speed.

Screenshot from 2022-07-04 23-46-08

``[2022-10-20 10:15:07,216] chirp.drivers.kenwood_live - INFO: Trying ID at baud 57600 with delimiter "('\r', ' ')"

[2022-10-20 10:15:07,327] chirp.drivers.kenwood_live - DEBUG: PC->RADIO: ID
[2022-10-20 10:15:07,828] chirp.drivers.kenwood_live - ERROR: Timeout waiting for data
[2022-10-20 10:15:07,828] chirp.drivers.kenwood_live - ERROR: Giving up
[2022-10-20 10:15:07,828] chirp.drivers.kenwood_live - INFO: Trying ID at baud 38400 with delimiter "('\r', ' ')"
[2022-10-20 10:15:07,939] chirp.drivers.kenwood_live - DEBUG: PC->RADIO: ID
[2022-10-20 10:15:08,440] chirp.drivers.kenwood_live - ERROR: Timeout waiting for data
[2022-10-20 10:15:08,440] chirp.drivers.kenwood_live - ERROR: Giving up
[2022-10-20 10:15:08,440] chirp.drivers.kenwood_live - INFO: Trying ID at baud 19200 with delimiter "('\r', ' ')"
[2022-10-20 10:15:08,551] chirp.drivers.kenwood_live - DEBUG: PC->RADIO: ID
``
I think that relaxing the decode() is ok since the protocol is ASCII as you said.

tests/Python3_Driver_Testing.md
sv1: For this to work, we need the decode() relaxing!

chirp/drivers/kenwood_live.py
kk7ds: Right, but this is exactly the point. The error indicates that we're getting something we don't expect. Making `command()` ignore all of these just means we won't be able to tell when something is going wrong, we'll just get mangled (or no) data back from the radio. If our cable is broken or noisy, blindly filtering out the non-ascii bits could cause us to do something bad to the radio, misinterpret a response, etc.

`command()` expects to converse with the radio, so it should not ignore those errors. However, the `get_id()` routine is probing for a radio, trying different rates, etc, so the situation is different. So here:

https://github.com/kk7ds/chirp/blob/py3/chirp/drivers/kenwood_live.py#L124

We're using `command()` but would not be surprised to get a protocol-violating error due to the fact that we don't know if we're speaking the correct baud rate. So you should `try..except UnicodeDecodeError` around the `command()` invocation and if we get one, don't even bother inspecting `resp` as you know it's not going to be what we want. That way, all the other users of `command()` will abort the conversation if we get unexpected non-ASCII bytes, which is what should happen.

chirp/drivers/retevis_rt22.py
kk7ds: That was different though. In that case we never needed to use the value as a string, and always needed it as `bytes`. Here, we've got a string that serves both purposes, so duplicating it is just risking one to get changed and not the other accidentally. The rule should be "if this was a string because py2 does not distinguish but is always used as bytes, then convert. If it's used in both forms, encode (or decode) as necessary to avoid the duplication/synchronization problem".

kk7ds: This is getting fixed in #75 anyway, since a recent change to master fixed the syntax error.

chirp/drivers/kenwood_live.py
```diff
@@ -121,13 +121,18 @@ def get_id(ser):
             ser.baudrate = i
             ser.write(LAST_DELIMITER[0].encode())
             ser.read(25)
-            resp = command(ser, "ID")
+            try:
+                resp = command(ser, "ID")
+            except UnicodeDecodeError:
+                LOG.info("Unknown utf-8 byte received at baud %i with \
+                        delimiter \"%s\"" % (i, repr(delimiter)))
```
kk7ds, line 128: This will just create noise in the logs for the common situation of talking to the radio at the wrong baud rate. I think we should just move on here. We return on success if L133 succeeds, but if not, we don't log "nope, wrong baud rate."

chirp/drivers/kenwood_live.py
```diff
@@ -121,13 +121,18 @@ def get_id(ser):
...
+                LOCK.release()
```
kk7ds, line 129: This is a bad idea, because it makes it easier for the lock to get out of sync. We should use a try..finally in command so that it is always dropped before we exit, even if we raise. This is a latent bug in the current implementation, which could be tickled if we hit other exceptions, so it's good that you highlighted it here.

kk7ds (commented):
This needs rebasing anyway, so I will do that, fix the lock thing, the log thing, squash the try..except patch and then push up the result.

PR #79, opened 2022-10-22
Fix exceptions, integer divisions, and chr() behaviour

sv1:


chirp/drivers/kguv8dplus.py
```diff
@@ -1103,9 +1103,15 @@ def set_settings(self, settings):
...
 
     def _is_freq(self, element):
-        return "rxfreq" in element.get_name() or "txoffset" in element.get_name() or "rx_start" in element.get_name() or "rx_stop" in element.get_name() or "tx_start" in element.get_name() or "tx_stop" in element.get_name()
\ No newline at end of file
+        return "rxfreq" in element.get_name() \
+                or "txoffset" in element.get_name() \
+                or "rx_start" in element.get_name() \
+                or "rx_stop" in element.get_name() \
+                or "tx_start" in element.get_name() \
+                or "tx_stop" in element.get_name()
```
kk7ds, line 1116: Is this related to the py3 conversion in some way? Or are you just reformatting this since you're here?

Obviously this is better than what was there, but the more non-critical changes we make to the py3 branch, the harder it is to keep it sync'd with master as they're both changing. Just something to keep in mind in the future :)

PR #82, opened 2022-10-22
Fix ICF-based radios for clone out

kk7ds:
At some point we apparently started slamming data down the serial line at too high a rate for older radios. When using an ID-800, the clone dialog finishes in a few seconds, and the radio times out because it missed a huge chunk of the image because it overran its buffer. This seemed to work okay for radios that use the direct serial connection to the 2.5mm data jack, but not via the OPC-478 type interface.

This change makes us read our own echo frames before sending the next one, which slows us down a lot. It feels a lot slower, but it means we're at least pacing the physical echo.

Tested with an ID-800, ID-880 and IC-2820 and this fixes things.

### DIFF
--- a/chirp/drivers/icf.py
+++ b/chirp/drivers/icf.py
@@ -115,15 +115,19 @@ class RadioStream:
 
         return frames
 
-    def get_frames(self, nolimit=False):
+    def get_frames(self, nolimit=False, limit=None):
         """Read any pending frames from the stream"""
         while True:
-            _data = self.pipe.read(64)
+            _data = self.pipe.read(1)
             if not _data:
+                if limit:
+                    LOG.warning('Hit timeout before one frame')
                 break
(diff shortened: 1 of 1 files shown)


kk7ds:
@yarda Any chance you could look this over and/or test? It's been a long time since I've used some of my older icom radios with chirp. While testing on the py3 branch I found that these older ones didn't work, but they don't work on master either. This is what I came up with which seems to work. Since you've modified this file in the past, I thought I'd see if you could at least test with whatever radio you have.

kk7ds:
cc @rrhett

yarda:
I can test with IC-E90.

kk7ds:
Oh well, I'm just going to merge this anyway. At worst I think it slows us down a but and at best, it fixes a bunch of icom radio communication.

PR #83, opened 2022-10-22
More wxui improvements and radio testing

kk7ds:


kk7ds:
I'm going to go ahead and merge this with the ICF change in it before it goes to master. I think the worst thing it will do is potentially slow down our clone, although that's the problem it's fixing too (going too fast)...

PR #84, opened 2022-10-23
Fix download for Baofeng UV-5R for py3

lekv:
Downloading from UV-5R using Python3 would fail with this error:

ERROR: Failed to clone: Failed to communicate with radio: can't concat str to bytes

As Tony Fuller pointed out in issue #10082, this happened because we were trying to append the radio model name as a string to the downloaded data, which was a bytes array.

To fix this, we now encode the model name before appending it to the data.

Testing: I tried this with my UV-5R and confirmed that the fix works and I can download data.

### DIFF
--- a/chirp/drivers/uv5r.py
+++ b/chirp/drivers/uv5r.py
@@ -602,7 +602,7 @@ def _do_download(radio):
             data += _read_block(radio, i, 0x40, False)
 
     if append_model:
-        data += radio.MODEL.ljust(8).encode('ascii', 'ignore')
+        data += radio.MODEL.ljust(8).encode()
 
     LOG.debug("done.")
     return memmap.MemoryMapBytes(data)
--- a/chirp/drivers/uv5r.py
+++ b/chirp/drivers/uv5r.py
@@ -602,7 +602,7 @@ def _do_download(radio):
             data += _read_block(radio, i, 0x40, False)
 
     if append_model:
-        data += radio.MODEL.ljust(8)
+        data += radio.MODEL.ljust(8).encode('ascii', 'ignore')
 
     LOG.debug("done.")
     return memmap.MemoryMapBytes(data)


chirp/drivers/uv5r.py
```diff
@@ -602,7 +602,7 @@ def _do_download(radio):
             data += _read_block(radio, i, 0x40, False)
 
     if append_model:
-        data += radio.MODEL.ljust(8)
+        data += radio.MODEL.ljust(8).encode('ascii', 'ignore')
```
kk7ds, line 605: Can you make this just `.encode()`? We shouldn't have any non-ASCII characters in those strings, since they're used for display to the user. This `ignore` flag will prevent us from catching errors if that ever changes.

lekv: Changed it and made sure that it still works.

lekv:
@kk7ds I briefly looked at the automated tests but from the conversation on the bug I figured that there won't be a way to test this without a radio. Let me know if you'd like me to add a unit test for this somewhere.

chirp/drivers/uv5r.py
kk7ds: Thanks, I'll merge this in a minute assuming tests pass.

kk7ds:
We have a little bit of clone simulation in the driver tests, but it likely doesn't make it far enough to hit this anyway. Unfortunately we don't have driver-specific unit tests for this level of stuff, unfortunately.

kk7ds:
Oh @lekv, can you add yourself to `tests/py3_driver_testers.txt`? We're trying to record which drivers have actually be tested on py3, because as you notice, they can pass tests but not ... pass tests. Since you've tested this now, we should capture that.

lekv:
> Oh @lekv, can you add yourself to `tests/py3_driver_testers.txt`? We're trying to record which drivers have actually be tested on py3, because as you notice, they can pass tests but not ... pass tests. Since you've tested this now, we should capture that.

Sent https://github.com/kk7ds/chirp/pull/85

PR #85, opened 2022-10-23
Add lekv to py3_driver_testers for Baofeng-UV5R

lekv:
Tested downloading, adding a channel, and uploading.


kk7ds:
Thanks! To pass the test you need to also regenerate the matrix so github will show it to web viewers. I just pushed that on top, FYI...

PR #86, opened 2022-10-23
Small fixes for py3 drivers 

sv1:


chirp/drivers/ft8100.py
```diff
@@ -175,7 +175,7 @@ def get_memory(self, number):
 
         if _mem.duplex == DUPLEX.index("split"):
             tx_freq = int(_mem.offset) * 1000
-            print self.VARIANT, number, tx_freq, mem.freq
+            print(self.VARIANT, number, tx_freq, mem.freq)
```
kk7ds, line 178: This is not your problem, but we should convert these to `LOG.debug()` calls or just remove them. Just noting this, no need to revise :)

PR #90, opened 2022-10-24
Fix exceptions, lists for python3 drivers

sv1:


chirp/drivers/kguv8e.py
```diff
@@ -552,6 +552,7 @@ def _get_dcs(val):
             return code, pol
 
         tpol = False
+        print(mem.number, _mem.txtone, _mem.rxtone)
```
kk7ds, line 555: Leftover debug? Please remove :)

kk7ds (commented):
If you could actually fix the commit in question instead of pushing another patch on top, that'd be my preference :)

Looks okay otherwise, thanks!

chirp/drivers/kguv8e.py
sv1: Sorry for that. There seems to be a bug with this driver when trying to decode/encode DCS codes which results in some memories not being loaded at all. I tried to look into it but had no success!

PR #97, opened 2022-10-26
Fix exceptions, lists for python3 drivers

sv1:


kk7ds:
In the future, if you convert a file's line endings like that, please put it in a separate commit from functional changes. Otherwise it's difficult to see what you're changing (other than the line endings). I pulled this one down locally and did the comparison, and was able to unconvert to see the few actual changes but github shows the entire file as changed :)

sv1:
OK! But as far as I can see this was the only file with line endings like this!

PR #99, opened 2022-11-05
Fix invalid syntax for python3 drivers

sv1:


chirp/drivers/ft1d.py
```diff
@@ -1131,7 +1132,8 @@ def _get_aprs_beacons(self):
                 checksum = body[-2:]
                 body = ''.join(s for s in body[:-2]
                                if s in string.printable).translate(
-                                   None, "\x09\x0a\x0b\x0c\x0d")
+                                   str.maketrans(
+                                       "", "", "\x09\x0a\x0b\x0c\x0d"))
```
kk7ds, line 1136: Wasn't this the driver we had trouble with loading the memory name from the image? Did you figure out what was going on with that?

sv1:
> 
> Wasn't this the driver we had trouble with loading the memory name from the
> image? Did you figure out what was going on with that?
> 
Yes! But I believe that this has nothing to do with python3 conversion.

I have tried to load the image with python2 chirp-daily-221025 snap and I see
the same problem!

According to MEM_FORMAT the flags for used and valid are here:

#seekto 0x280A;
struct {
  u8 nosubvfo:1,
     unknown:3,
     pskip:1,
     skip:1,
     used:1,
     valid:1;
} flag[%d];


But the image contains:

00002800: ffff ffff ffff ffff ffff 0300 0000 0000  ................
00002810: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002820: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002830: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002840: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002850: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002860: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002870: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002880: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00002890: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000028a0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000028b0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000028c0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000028d0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
000028e0: 0000 0000 0000 0000 0000 0000 0000 0000  ................


So for a used memory, it seems to me that the 'used' and 'valid'  flags is 0 and
empty 1.

I do not know why we use 'not' below:

if not flag.used:
            mem.empty = True
        if not flag.valid:
            mem.empty = True

PR #100, opened 2022-11-05
add Lanchonlh HG-UV98 driver

masenf:
This driver was written in Apr 2022 and I've been testing it intermittently since then. I think it's ready to merge upstream.

This is my first contribution for CHiRP, so I request any functional, stylistic or process suggestions regarding this change.

fixes [#7999](https://chirp.danplanet.com/issues/7999)

related [thread](https://groups.io/g/HG-UV98-users/topic/chirp_support/77354195) on the HG-UV98 groups.io


chirp/drivers/hg_uv98.py
```diff
@@ -0,0 +1,897 @@
...
+    def __init__(self, radio, current_raw):
+        current = int(current_raw)
+        current_mem = radio.get_memory(current)
+        options = [
+            self._format_memory(mem)
+            for mem in [
+                radio.get_memory(n)
+                for n in range(*radio.get_features().memory_bounds)
+            ]
+        ]
```
kk7ds, line 294: This is a nit, but my OCD won't allow me to avoid saying: I *hate* this style, as it's very vertically-wasteful, especially with the parens below.

I don't have very strict auto-style enforcement in chirp, so I shan't hold you up on this, but, I just had to say it :)

isi-mfurer: this is the [psf/black](https://black.readthedocs.io/en/stable/) style that a lot of python projects are adopting these days so their maintainers don't have to think about formatting. lots of trailing commas, and lines with single parens and brackets.

i use it at work, it took some growing on me; and there are definitely a handful of coworkers that _hate_ it with a passion, and won't use it.

since getting used to it, I just tend to write stuff like this, even when I'm not using the formatter. 😬 


thanks for taking the change 👋 👍

kk7ds: Yeah, I'm well aware of where it comes from, and I'm one of those who effing hate it :)

goldstar611:
This driver needs to utilize the "future stanza" instead of a single import builtins.

I'm currently mobile but reading the open issue at https://chirp.danplanet.com/issues/10110

The future stanza I'm referring to is like
```
try:
    from builtins import bytes
    has_future = True
except ImportError:
    has_future = False
    LOG.debug('python-future package is not available; '
              '%s requires it' % __name__)
```

PR #103, opened 2022-11-08
hg_uv98: skip register if `future` is not available

masenf:
this driver currently depends on builtins.bytes

ass try/except ImportError when doing

```
from builtins import bytes
```

since some platforms dont have the `future` package and the driver shouldn't crash the whole program in those cases.

fixes #10110

### DIFF
--- a/chirp/drivers/hg_uv98.py
+++ b/chirp/drivers/hg_uv98.py
@@ -20,7 +20,6 @@
 #  KG7KMV and Bartłomiej Zieliński
 # Based on the implementation of Kenwood TK-8102
 
-from builtins import bytes
 import os
 import logging
 import struct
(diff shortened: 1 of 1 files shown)


chirp/drivers/hg_uv98.py
```diff
@@ -33,6 +32,18 @@
...
 LOG = logging.getLogger(__name__)
 
+# Gross hack to handle missing future module on un-updatable
+# platforms like MacOS. Just avoid registering these radio
+# classes for now.
+try:
+    from builtins import bytes
+    has_future = True
+except ImportError:
+    has_future = False
```
kk7ds, line 42: You need to make the registration conditional on this so we don't register the driver if it will crash later. See the bottom of tk8180.py.

Apologies that you based this on tk8102.py which has a build system exception and also needs fixing, but thanks for following up.

kk7ds (changes requested):

chirp/drivers/hg_uv98.py
masenf: @kk7ds alright, i think the latest change should be sufficient

ultimately i want this driver available everywhere, so i'm going to try to get rid of this import, but for now, lets get unblocked on #10110 with a new build

masenf: in the meantime i've posted #104 which removes the `builtins` import entirely and should also fix the startup bug.

kk7ds:
I'mma just merge this one so we definitely have a fix for tomorrow and it's my preference. We can discuss #104 if you strongly prefer that.

PR #104, opened 2022-11-08
hg_uv98 remove future builtins dependency (windows support)

masenf:
Alternative to #103 

Also fixes bug #10110

This removes the use of `bytes` in the driver entirely.

The only place where `bytes` was material was on line 197, where `data` from `radio.pip.read` was being cast as `bytes`. The data from the pipe should _already_ be bytes, so there is no need to cast here.

The other place we're checking against `bytes` is line 172, which was incorrect anyway... if data was not `bytes`, then the code should `encode` on line 173, NOT `decode`. This was removed entirely

The other spots where `bytes` was used were redundant given the `b` prefix on the string.

### DIFF
--- a/chirp/drivers/hg_uv98.py
+++ b/chirp/drivers/hg_uv98.py
@@ -32,17 +32,6 @@ from chirp.settings import RadioSettingValueString, RadioSettings
 
 LOG = logging.getLogger(__name__)
 
-# Gross hack to handle missing future module on un-updatable
-# platforms like MacOS. Just avoid registering these radio
-# classes for now.
-try:
-    from builtins import bytes
-    has_future = True
-except ImportError:
-    has_future = False
-    LOG.debug('python-future package is not '
-              'available; %s requires it' % __name__)
-
 
 MEM_FORMAT = """
 #seekto 0x0000;
@@ -180,8 +169,6 @@ BEACON_EXIT_DLY_LIST = MENU_DLY_LIST = AUTOLOCK_DLY_LIST
 
 
 def make_frame(cmd, addr, length, data=b""):
-    if not isinstance(data, bytes):
-        data = data.decode("ascii")
     return struct.pack(">BHB", ord(cmd), addr, length) + data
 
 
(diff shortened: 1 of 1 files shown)


kk7ds:
So, the point of the drivers using real bytes is so that they'll behave properly in the py3 branch with minimal changes, where the memmap *is* bytes, serial returns bytes, etc.

Since the missing future lib is just the macos platform, I'd rather take the conditional registration, personally. Are you particularly concerned about the legacy build working on macos? If so, I can take this instead, but it's like 0.5% of users :)

masenf:
> Are you particularly concerned about the legacy build working on macos?

This is actually the build that I use personally... and in my testing, it actually worked just fine on the `chirp-unified-daily-20221108.app.zip` build.

The build that _didn't_ work was the windows build 20221108, so I assumed the missing `future` affected the windows platform, which is important to me.

I also don't think i particularly _need_ to use the name `bytes` anywhere in my driver... I'll do some poking on the py3 branch, and see if it makes a difference either way.

When I'm ready, I'll reopen this PR for review and ping you.

kk7ds:
Okay, the first report I saw was reported against macos, and that's the only platform that should be missing future. We have it on the windows build.

Anyway, yep, let me know.

masenf:
The initial report I saw in #10110 listed Win7 as the platform in the debug log.

When I run the win32 build for 20221109, it runs without crashing, but I don't see my driver in the list. Additionally the Kenwood TK-8180 is also missing.

kk7ds:
Okay, looks like the build system has been failing to install that package for some amount of time:

```
'ConnectTimeoutError(<pip._vendor.requests.packages.urllib3.connection.VerifiedHTTPSConnection object at 0x0131B0B0>, 'Connection to pypi.python.org timed out. (connect timeout=15)')': /simple/future/
  Could not find a version that satisfies the requirement future (from versions: )
No matching distribution found for future
```

The macos build never has/had it. Win32 always did, but the above is getting silently ignored. I'll have to dig into that.

masenf:
> I'll have to dig into that.

Appreciate it. If we can get `future` into the windows build, then we don't need this patch, which sounds like your preference. This would also unblock other drivers that depend on future.

kk7ds:
Yep, fixed and I updated the 20221109 win32 builds on the site. Shouldn't be a problem again.

PR #105, opened 2022-11-09
Test some drivers - btech.py

KC9HI:
This patch makes driver changes that will allow the Python3 Driver Testing matrix status of the BTech UV-25X4 to be shown as "Tested".


kk7ds (commented):
Awesome to see you here Jim! Some comments, but I'll merge this anyway. Thanks!

PR #107, opened 2022-11-10
Try setuptools if present, otherwise use distutils.core.

KC9HI:


kk7ds:
I do want this, I think it's the right thing to do for now. I'll rebase this.

PR #114, opened 2022-11-17
Fix Float properties

kk7ds:
Apparently the ChirpFloatProperty was copied from ChirpIntProperty and never updated. This makes it work as expected.


kk7ds:
CC @KC9HI

PR #149, opened 2022-11-24
Final py3 syntax fixes and a clean sdist

kk7ds:
- Start fresh with setup.py
- Fix up the last few py3 syntax errors
- Remove the safe import and py3_remaining stuff


kk7ds:
CC @sv1 This started as just the sdist stuff, but realized I could do a little work and get us over the edge so it's actually clean. Thanks for your help :)

PR #158, opened 2022-11-29
Fixes and Testing: radtel_t18.py & Update Python3 Driver Testing Matrix

KC9HI:


chirp/drivers/radtel_t18.py
```diff
@@ -316,8 +316,9 @@ def _t18_read_block(radio, block_addr, block_size):
 def _t18_write_block(radio, block_addr, block_size):
     serial = radio.pipe
 
-    cmd = struct.pack(">cHb", 'W', block_addr, block_size)
-    data = radio.get_mmap()[block_addr:block_addr + block_size]
+    cmd = struct.pack(">cHb", b'W', block_addr, block_size)
+    data = radio.get_mmap().get_byte_compatible()[block_addr:block_addr +
```
kk7ds, line 320: Okay, so ... This is technically okay, but... If you set `NEEDS_COMPAT_SERIAL=False` on your radio, then the `radio._mmap` will *always* be a MemoryMapBytes(), you can use `bytes` everywhere and not have to do this conversion here while talking to the radio.

If you're interested, setting that will get all these drivers the coveted "byte clean" designation :)

kk7ds:
I'll go ahead and merge this since it works, but if you want to try the non-compat route, that'd be cool. It'll make the driver perform better too.

I had originally intended to pass CompatSerial to drivers if they set that flag to False, but never did that in the wxui apparently. I've been converting all the drivers I've been testing. If we want to do that, more drivers may incidentally work, but they'll be kinda permanently in the past.

Maybe the thing to do is convert as many radios as we can test, and then when we run out, enable compat mode on the rest so they have the best chance of working.

But for that to work, we need to set `NEEDS_COMPAT_SERIAL=False` on all the ones we fix properly.

KC9HI:
On Mon, Nov 28, 2022 at 10:24 PM Dan Smith ***@***.***> wrote:
>
> @kk7ds commented on this pull request.
>
> ________________________________
>
> In chirp/drivers/radtel_t18.py:
>
> > @@ -316,8 +316,9 @@ def _t18_read_block(radio, block_addr, block_size):
>  def _t18_write_block(radio, block_addr, block_size):
>      serial = radio.pipe
>
> -    cmd = struct.pack(">cHb", 'W', block_addr, block_size)
> -    data = radio.get_mmap()[block_addr:block_addr + block_size]
> +    cmd = struct.pack(">cHb", b'W', block_addr, block_size)
> +    data = radio.get_mmap().get_byte_compatible()[block_addr:block_addr +
>
> Okay, so ... This is technically okay, but... If you set NEEDS_COMPAT_SERIAL=False on your radio, then the radio._mmap will always be a MemoryMapBytes(), you can use bytes everywhere and not have to do this conversion here while talking to the radio.
>
> If you're interested, setting that will get all these drivers the coveted "byte clean" designation :)
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you authored the thread.Message ID: ***@***.***>

Yep. It wasn't working and I did what I usually do, I looked at
something else that was working and did the same thing.

I will make this change before I start anything else.

Jim

KC9HI:
On Mon, Nov 28, 2022 at 10:24 PM Dan Smith ***@***.***> wrote:
>
> @kk7ds commented on this pull request.
>
> ________________________________
>
> In chirp/drivers/radtel_t18.py:
>
> > @@ -316,8 +316,9 @@ def _t18_read_block(radio, block_addr, block_size):
>  def _t18_write_block(radio, block_addr, block_size):
>      serial = radio.pipe
>
> -    cmd = struct.pack(">cHb", 'W', block_addr, block_size)
> -    data = radio.get_mmap()[block_addr:block_addr + block_size]
> +    cmd = struct.pack(">cHb", b'W', block_addr, block_size)
> +    data = radio.get_mmap().get_byte_compatible()[block_addr:block_addr +
>
> Okay, so ... This is technically okay, but... If you set NEEDS_COMPAT_SERIAL=False on your radio, then the radio._mmap will always be a MemoryMapBytes(), you can use bytes everywhere and not have to do this conversion here while talking to the radio.

OK. I'm confused again. Almost every driver that claims to be "Byte
Clean" has ".get_byte_compatible()" in it.

If I remove the ".get_byte_compatible() from the radtel_t18.py driver
and add "NEEDS_COMPAT_SERIAL = False" to the radio, I can successfully
upload a tab that was created by loading a previously saved CHIRP
Radio Images File, however, uploading a tab that was created by
downloading from a radio fails. The error is...

C:\Users\Jim\mychirpgit\chirp>python3 .\chirpwx.py
ERROR: Failed to clone: can't concat str to bytes
  File "C:\Users\Jim\mychirpgit\chirp\chirp\wxui\clone.py", line 48, in run
    self._fn()
  File "C:\Users\Jim\mychirpgit\chirp\chirp\drivers\radtel_t18.py",
line 457, in sync_out
    do_upload(self)
  File "C:\Users\Jim\mychirpgit\chirp\chirp\drivers\radtel_t18.py",
line 376, in do_upload
    _t18_write_block(radio, addr, radio.BLOCK_SIZE)
  File "C:\Users\Jim\mychirpgit\chirp\chirp\drivers\radtel_t18.py",
line 323, in _t18_write_block
    LOG.debug(util.hexprint(cmd + data))
TypeError: can't concat str to bytes

Jim

kk7ds:
Some of that is compatibility... get_byte_compatible() on a MemoryMapBytes is a no-op, it just returns self.

Did you replace this too? https://github.com/kk7ds/chirp/blob/py3/chirp/drivers/radtel_t18.py#L361

When you load a file, you get your memmap from the UI (based on the `NEEDS_COMPAT_SERIAL` flag, but on download, you do it yourself.

PR #160, opened 2022-11-30
py3wxui

kk7ds:
- Fix extra rows on 1-origin memory_bounds radios
- Implement dv_code and dsql_mode for id31, et al
- Implement DVMemory editing via props dialog
- Fix TH-D72 clone driver for py3


kk7ds:
CC @KC9HI This has the extra rows fix in it

PR #163, opened 2022-11-30
Fix mainapp for py3

glv2:


kk7ds:
I can apply this if you want, but we're not supporting the gtk UI in the py3 branch. We're moving on to wxui (chirpwx.py). I haven't removed it from the tree just so that we can keep in sync with master easier...

glv2:
I was making a GNU Guix package for chirp, where the build system compiles all the Python files, and it failed because of the line with bad syntax in "mainapp.py". This patch fixes this, but if the file can be removed from the sources, it works for me too.
Feel free to either apply or close this pull request as you prefer.

kk7ds:
Okay, we can apply just to avoid having unparseable files in the tree for now (that only recently became doable for the drivers anyway). In the future, we'll remove it at some point.

PR #177, opened 2022-12-04
Tweak tab and window close behavior

kk7ds:
- Prompt to save tabs instead of the opposite
- Make sure CloneThread closes the serial when done


kk7ds:
CC @KC9HI

PR #182, opened 2022-12-04
Be graceful about missing stock, icon paths

kk7ds:
This shouldn't happen, but we also don't need to crash if it does.


kk7ds:
@vishwin This will avoid the crash if we don't find the files, since it should not be fatal anyway.  But, still would like to understand why you don't get them installed.

PR #200, opened 2022-12-07
Fix tox.ini for tox 4.0

kk7ds:
- $HOME is not always defined in Unix-like environments
- tox: whitelist -> allowlist


kk7ds:
@vishwin grabbed these to merge sooner since tox 4 will be blocking everything until we do.

Hope that's okay :)

PR #208, opened 2022-12-09
Changes for supporting Alinco DJ-G7 in the py3 branch

mweyland:
Hello

This was done following the [call on the mailing list](http://intrepid.danplanet.com/pipermail/chirp_devel/2022-December/006953.html). I'm a bit rusty so this is probably not following the chirp PR etiquette. I remember the process was somewhat unusual but may have changed since this is now using git rather than mercurial.

In any case, the current py3 branch does not work with the radio in the subject. This is because the driver is expecting strings rather than python `bytes` and is also using strings internally to compile the proper commands that are then sent to the radio. In particular, the `"... %s ..." % foo` idiom is used quite a bit, and would include the the `b` prefix if `foo` is of the `bytes` type. That is, of course, undesired. To make this work, we removed the `b` prefix in all `_model` variables and added `encode()` and `decode()` in `_send()` and `receive()`. We also amended the list of tested radios.

I can only guess that there was some debate among developers how to deal with this very problem. The solution chosen here has the advantage that it leaves most of the driver untouched. One could make the point that now would be a great time to implement this in a cleaner way, but I have no way to test the change for all other Alinco radios; I would probably introduce bugs if I go for a more elaborate change.

This code passes the unit tests and reading from / writing to the Alinco DJ-G7 is successful. I assume that I am allowed to add the "byte clean" flag to the list of tested radios under these conditions (not sure what that column really means...)

Thank you for all your efforts!
Matt


kk7ds:
Hi Matt, thanks for jumping in here. You're almost right, except that the byte conversion was done recently by a developer trying to fix things but without a radio to do it with. The alinco protocol is all actual strings, so it makes sense to do what you did and put it back to all strings until we actually write it to the serial port.

The byte-clean flag is actually set automatically by the thing that generates the matrix based on whether or not the driver has `NEEDS_COMPAT_SERIAL=False`  set on it. In order to get that flag set, you need to set that flag on the driver (the base driver for alincos is fine) and then also change `MemoryMap` to `MemoryMapBytes` in the `_download()` function. When you do this, there will probably be a few other changes. Would you be willing to try to do that and amend this so we can get these drivers straight to byte-clean?

Also, you  need to update the `tests/python3_driver_testers.txt` file and run `tox -emakesupported` to update the `.md` file. I think it would be fine to mark all the other affected alinco drivers as implied by your G7, for what it's worth, since they're all so similar.

Let me know if you can do this, and if not, I'll take a stab for you and see if you can test the result for us.

Thanks!

mweyland:
Hello Dan

Thank you! Regarding `tests/python3_driver_testers.txt`, I was too late in noticing that. I have added another commit to the PR fixing this. github found a conflict with the .md file, probably because I committed a version that was edited manually before I did it properly(?). I'm not sure how to deal with that and I apologize for not following the instructions. 

Regarding the use of `MemoryMapBytes`: I'm happy to give this a try tomorrow. You can leave the PR open if that works for you. I'm confident I can handle the transition for the DJ-G7, but I am afraid of breaking stuff for other Alinco radios...

Regards, Matt

kk7ds:
> Thank you! Regarding `tests/python3_driver_testers.txt`, I was too late in noticing that. I have added another commit to the PR fixing this. github found a conflict with the .md file, probably because I committed a version that was edited manually before I did it properly(?). I'm not sure how to deal with that and I apologize for not following the instructions.

No worries, we're all conflicting with it all the time. Just rebase your commit on top of the current `py3` branch and then re-run `tox -emakesupported` and it will generate a new clean one to add to your commit.

> Regarding the use of `MemoryMapBytes`: I'm happy to give this a try tomorrow. You can leave the PR open if that works for you. I'm confident I can handle the transition for the DJ-G7, but I am afraid of breaking stuff for other Alinco radios...

I think all the alincos use the same protocol (which you're already changing here) so if they pass the tests, I think they're okay. Much of the `MemoryMapBytes` stuff is checked via the automated tests, so I think we're safe there. I promise I won't hold you accountable :)

And yep, I'll just leave this PR until you get a chance to update it. Thanks!

kk7ds:
Run `tox -e makesupported` locally and commit the generated `.md` file as it is. That's what the check is trying to assert, that you committed the updated version. I can also do this for you when I review, no worries. I'll look over this later today, thanks for circling back!

mweyland:
OK I think I have it. In summary, this provides a byte-clean driver for the DJ-G7 and adds encode() and decode() statements for the other radios, such that they remain in a working state (yet not byte-clean). I also fixed a minor bug with the debug output while I was at it. This still passes the unit tests and reading from / writing to the radio works. Is that what you had in mind?

I could try to make the remaining radios byte-clean but without anyone who could test my work, that would probably do more harm than good(?) Let me know what you think...

kk7ds:
Yes, but I think it's okay to do this for all the radios, because (from memory) they all use the same clone routines and the driver tests will make sure that the byte-clean memmap stuff is solid. But, let's just go with what you have here for the G7 and I'll circle back later to refactor the others. That way my name will be on that commit and I can take the blame :)

I gotta run for a bit but will look this over in detail when I get back. Thanks!

mweyland:
No worries. I'll head off for a little bit but I'm happy to re-iterate on this one. Maybe it's best if you have a look whenever you have time (no rush)... The clone routines are mostly the same but not 100%. The overrides are from the DJ175 (minor) and from the G7, which is unfortunately the radio that I have and thus the code that I modified.

That said, I guess the path to make this fully byte-clean would be like this:
- Move `_send()`, `_read()` and `_identify()` from the G7 to the base class and set `NEEDS_COMPAT_SERIAL` there.
- `cmd = b"AL~F%04XR\r\n" % addr` and `data = codecs.decode(_data, "hex")` in `_download_chunk()`. Also check if the processing of `resp` needs some work -- e.g. I believe that `split()` should receive a byte-like argument.
- In `_download()`: `data = b""`, `self._send("AL~E\r\n")`and of course return `MemoryMapBytes`. 
- In `_upload_chunk()`: Use `data = codecs.encode(_data, "hex").upper()` and `cmd = "AL~F%04XW%s\r\n" % (addr, data)`.
- In `_upload()`: self._send(b"AL~E\r\n")`
- Repeat the aforementioned changes for the `_download_chunk()` override in `DJ175Radio`.
- Reintroduce `b"..."` in the `_model` assignment.
- Minor: Check if there is a more elegant way to do `isascii()` and `tohex()`, the latter can probably be handled by `codecs`.

These are just some quick notes while my memory is fresh... Apart from the last, minor point, these are things that I did in the overrides for the G7.

kk7ds:
I'm going to rebase this and squash the iterative commits together, but then it looks okay to me, thanks!

kk7ds:
Er, looks like you have it disabled. If you can turn on "enable edits from maintainers" then I can rebase this for you. If you don't want to, can you do that and also squash this down to one commit? No need to have multiple attempts to fix tests in the git history...

Thanks!

mweyland:
This may be because my repo clone is not a personal one, but stored within an organization... I did not see an option to enable edits for maintainers. But in any case I squashed the branch down to one commited as suggested...

mweyland:
FYI, I rebased the squashed branch once again, and I also started another branch `pr_py3_alinco_style` with the changes mentioned above.

kk7ds:
Okay, could be. I wrote an alinco clone simulator that I also wanted to push in here for you to look at, as I think we can use that to prove that the other drivers work with byte clean fixes. I will push it up to a PR myself and cc you to have a look if you're interested. Thanks!

mweyland:
> I will push it up to a PR myself and cc you to have a look if you're interested. Thanks!

Definitely interested, please let me know about this! Thanks again

PR #214, opened 2022-12-10
Fixes and testing: uv5x3.py & update Python3 Driver Testing matrix

KC9HI:


kk7ds:
This is almost rebased, let me just fix the last little thing for you.

I promise, once we get past this period of everyone trying to update the same two files all the time, things will be *much* easier. I probably should have figured out a different way to do this that didn't require passing a baton around.

PR #221, opened 2022-12-10
Fixes and testing: BF-T1

KC9HI:


kk7ds:
I assume you were holding off on the tested notation for all these until the end right?

KC9HI:
On Sat, Dec 10, 2022 at 9:40 PM Dan Smith ***@***.***> wrote:
>
> I assume you were holding off on the tested notation for all these until the end right?

Yes. I didn't think it would work without issues, so I held back. I
will have it in a few minutes.

Jim

PR #227, opened 2022-12-11
Fix squashed clone dialog on Linux

kk7ds:
Fixes #10173


kk7ds:
@dominickpastore This should fix the clone dialog thing.

PR #230, opened 2022-12-11
Wxui RadioReference

jebbgrenham:
All the radio reference Canada import work from this week. Minor UI tweak to dialog. Prep work for future RadioReference import for USA. The segfaults are resolved, so I think this can be an actual pull request. Flake8 should be ok now.

### DIFF
--- a/chirp/wxui/main.py
+++ b/chirp/wxui/main.py
@@ -527,6 +527,16 @@ class ChirpMain(wx.Frame):
         source_menu = wx.Menu()
         radio_menu.AppendSubMenu(source_menu, _('Query Source'))
 
+        query_rrca_item = wx.MenuItem(source_menu,
+                                      wx.NewId(), 'RadioReference.com Canada')
+        self.Bind(wx.EVT_MENU, self._menu_query_rrca, query_rrca_item)
+        source_menu.Append(query_rrca_item)
+        # Soon to be implemented
+        # query_rrus_item = wx.MenuItem(source_menu,
+        # wx.NewId(), 'RadioReference USA')
+        # self.Bind(wx.EVT_MENU, self._menu_query_rrus, query_rrus_item)
+        # source_menu.Append(query_rrus_item)
+
         query_rb_item = wx.MenuItem(source_menu, wx.NewId(), 'RepeaterBook')
         self.Bind(wx.EVT_MENU, self._menu_query_rb, query_rb_item)
         source_menu.Append(query_rb_item)
@@ -1150,6 +1160,23 @@ class ChirpMain(wx.Frame):
             LOG.info('Selected bandplan: %s' % selected)
             for shortname, name in plans:
                 CONF.set_bool(shortname, shortname == selected, 'bandplan')
+
+    def _menu_query_rrca(self, event):
+        d = query_sources.RRCAQueryDialog(self,
+                                          title='Query RadioReference.com '
+                                                '(Canada)')
+        r = d.ShowModal()
+        if r == wx.ID_OK:
+            LOG.debug('Result file: %s' % d.result_file)
+            self.open_file(d.result_file)
+
+    def _menu_query_rrus(self, event):
+        d = query_sources.RRUSQueryDialog(self,
+                                          title='Query RadioReference (USA)')
+        r = d.ShowModal()
+        if r == wx.ID_OK:
+            LOG.debug('Result file: %s' % d.result_file)
+            self.open_file(d.result_file)
 
     def _menu_query_rb(self, event):
         d = query_sources.RepeaterBookQueryDialog(
--- a/chirp/wxui/query_sources.py
+++ b/chirp/wxui/query_sources.py
@@ -24,6 +24,7 @@ import wx
 
 from chirp import CHIRP_VERSION
 from chirp import dmrmarc
+from chirp import radioreference
 from chirp.drivers import generic_csv
 from chirp.ui import config
 from chirp.ui import fips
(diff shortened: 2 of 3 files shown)


kk7ds:
I just rebased, fixed the tests, and squashed the iterations so far together. Do you want to hold this until you're ready with the rest of the US stuff or what?

requirements.txt
```diff
@@ -5,3 +5,4 @@ pyserial
 future
 requests
 pywin32; platform_system=="Windows"
+suds
```
kk7ds, line 8: Note I added suds here so we get it installed in the build and on people's machines when they do this from source.

chirp/wxui/query_sources.py
```diff
@@ -188,6 +189,70 @@ def do_query(self):
...
+    def do_query(self):
+        global rrloggedin
+        if rrloggedin:
+            rrloggedin = False
+            self.finalRRquery()
+        else:
+            self.do_loginRR()
+
+    def do_loginRR(self):
+        global clist, provinces, default_county, rrloggedin
```
kk7ds, line 205: I guess the goal is to have these all be global so that even after we've run one query, we remain in "logged in" state and don't need to re-list provinces and such?

kk7ds (commented):
Looks okay other than the string untranslation. If you're going to keep working on the US part before we merge, then you can just fix that on top of this. If not, I can add that back into this before we merge.

chirp/wxui/query_sources.py
jebbgrenham: Ah. That was the goal, but in the end I abandoned that quest. It's annoying to have to re-login but it's probably the better way to do it. I'm still a bit on the fence about it, but I think re-logging in is the way to go. Open to hearing opinions though!

requirements.txt
jebbgrenham: Thanks, right!

chirp/wxui/query_sources.py
```diff
@@ -397,3 +462,159 @@ def get_dm_params(self):
...
+    def populateprov(self, event):
+        self.selected_province(event.GetString)
+
+    def populatepc(self):
+        # init and grab conf defaults and populate the selector
+        RRCAQueryDialog.getconfdefaults(self)
+        # Clear the temporary choice dropdown, disable the form bits we don't
+        # need any more & insert provinces as exist in RR database because
+        # they change sometimes.
+        wx.CallAfter(provchoice.Clear)
```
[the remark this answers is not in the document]
jebbgrenham: Much more sensible. Do you want to make that change or should I?

chirp/wxui/main.py
```diff
@@ -525,7 +525,17 @@ def make_menubar(self):
         radio_menu.Append(upload_item)
 
         source_menu = wx.Menu()
-        radio_menu.AppendSubMenu(source_menu, _('Query Source'))
+        radio_menu.AppendSubMenu(source_menu, 'Query Source')
+
+        query_rrca_item = wx.MenuItem(source_menu,
+                                      wx.NewId(), 'RadioReference Canada')
```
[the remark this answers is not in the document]
jebbgrenham: I would leave it untranslated personally, would match the RR web interface? No good answer :)

chirp/wxui/main.py
```diff
@@ -525,7 +525,17 @@ def make_menubar(self):
         radio_menu.Append(upload_item)
 
         source_menu = wx.Menu()
-        radio_menu.AppendSubMenu(source_menu, _('Query Source'))
+        radio_menu.AppendSubMenu(source_menu, 'Query Source')
```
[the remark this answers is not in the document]
jebbgrenham: I think I had pre-translation code initially but yes it should be translated.

chirp/wxui/query_sources.py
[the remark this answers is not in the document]
kk7ds: Up to you, you can clean it up when you finish the US stuff. It's not critical to change, just a minor thing.

chirp/wxui/query_sources.py
jebbgrenham: No, it's not actually retaining a "session" just retaining confirmation that the credentials were valid at the time. It re-authenticates with RR when the county is submitted. If the credentials become invalid in the interim (ie, the premium subscription ends) then it would fail despite reportedly being logged in.

jebbgrenham: The vernacular "logged in" is probably not actually true when it's displayed on the status message...

kk7ds: Ah, okay cool. So the reason for the login step is that you can't pull the provinces, states, or whatever until you're authed?

chirp/wxui/query_sources.py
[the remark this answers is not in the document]
jebbgrenham: Will fix when US implement work starts. Cheers.

jebbgrenham:
Irritatingly yes. On Dec. 11, 2022 15:21, Dan Smith ***@***.***> wrote:
@kk7ds commented on this pull request.



In chirp/wxui/query_sources.py:
> @@ -188,6 +189,70 @@ def do_query(self):
         self.send_end()
 
 
+class RRCALOGINQueryThread(QueryThread):
+    global rrloggedin
+    rrloggedin = False
+
+    def do_query(self):
+        global rrloggedin
+        if rrloggedin:
+            rrloggedin = False
+            self.finalRRquery()
+        else:
+            self.do_loginRR()
+
+    def do_loginRR(self):
+        global clist, provinces, default_county, rrloggedin

Ah, okay cool. So the reason for the login step is that you can't pull the provinces, states, or whatever until you're authed?

chirp/wxui/query_sources.py
jebbgrenham: Irritatingly, yes.

PR #235, opened 2022-12-12
Add alinco clone tester and byte-clean the rest of the alincos

kk7ds:


mweyland:
> Okay, see what you think about this. In the last patch I remove the byte-clean overrides you added for the G7, but no test changes, so that should mean the G7 continues to work as expected. I can hold off on merging this until you test if you want (and are willing).

I went over the changes, all looks great, makes sense and is pretty much in line with what I did. I'm happy to test this when I get home but I would be super surprised to see this fail, given the circumstances.

I'll report how the test went.

kk7ds:
Cool, thanks, I'll hold off until you confirm. Gotta love simple well-behaved radios like this. I wish we had more in-tree clone simulators like this, but it would be a big ask for most people to write one for every radio. Knocking off a lot of alincos with one simulator (similar to my icom simulator) is a big win though.

mweyland:
The `alinco` branch worked perfectly on real G7 hardware. After this, I don't think there is any reason not to proceed with the merge...

kk7ds:
Cool, sounds good, I'll rebase and merge. Thanks!

PR #237, opened 2022-12-12
Allow editing bank names

kk7ds:
This makes it possible to edit bank names by double-clicking on the
column in the bank editor.

Fixes #10177


kk7ds:
@dominickpastore FYI

PR #248, opened 2022-12-14
Bugfix unassigned RR import

jebbgrenham:
When you very kindly moved my EVT_CHOICE method for the province selector to where it should be, it broke the populating of the county Choice dropdown when user changes province Choice selector. Here's a fix. Probably not the "right" fix, but it restores it to a working state.

### DIFF
--- a/chirp/wxui/query_sources.py
+++ b/chirp/wxui/query_sources.py
@@ -515,7 +515,7 @@ class RRCAQueryDialog(QuerySourceDialog):
         return vbox
 
     def populateprov(self, event):
-        self.selected_province(event.GetString)
+        self.selected_province(provchoice.GetStringSelection())
 
     def populatepc(self):
         # init and grab conf defaults and populate the selector


kk7ds:
Eesh, sorry.

chirp/wxui/query_sources.py
```diff
@@ -515,7 +515,7 @@ def build(self):
         return vbox
 
     def populateprov(self, event):
-        self.selected_province(event.GetString)
+        self.selected_province(provchoice.GetStringSelection())
```
kk7ds, line 518: Hmm, provchoice isn't in scope here is it? Ohh, it's global. Hmm, that's why I changed it, thinking that moving it out of the build function would lose sight of it. That's probably not the best plan, to have UI elements be global, especially if they could get re-used or re-parented by a second invocation. Better to store them on self so they live and die with the dialog itself. Did you do that because you need to access them from the other event handlers, or to try to avoid re-building them at some point? Storing them on self makes them accessible later and avoids them needing to be global.

Also, I think `event.GetString()` will make this work (if properly called as a function).

kk7ds:
I'm going to go ahead and apply this because it fixes the bug I introduced, but I'd like us to move anything that doesn't actually need to be global onto self. But I don't want to stop your progress on the US stuff, so we (or I) can circle back.

jebbgrenham:
Yeah it should be on self sorry. Another holdover from me trying to retain the credentials-ok status. I'll get to that in code cleanup too. Thanks for accepting the temp fix. -MarkOn Dec. 14, 2022 07:53, Dan Smith ***@***.***> wrote:
I'm going to go ahead and apply this because it fixes the bug I introduced, but I'd like us to move anything that doesn't actually need to be global onto self. But I don't want to stop your progress on the US stuff, so we (or I) can circle back.

PR #249, opened 2022-12-14
Some more wxUI tweaks

kk7ds:
- Make the minimum column width wider on Linux
- Allow selecting page at startup from CLI
- Allow starting some actions from the CLI

### DIFF
--- a/chirp/wxui/__init__.py
+++ b/chirp/wxui/__init__.py
@@ -53,6 +53,8 @@ def chirpmain():
                         help="Include this driver while loading")
     parser.add_argument("--inspect", action="store_true",
                         help="Show wxPython inspector")
+    parser.add_argument('--page', default=None,
+                        help='Select this page of the default editor at start')
     if sys.platform == 'linux':
         parser.add_argument('--no-linux-gdk-backend', action='store_true',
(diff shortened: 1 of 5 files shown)


kk7ds:
@jebbgrenham Interested in your thoughts on two things here. First, I think you're running linux. On all the (two) linux flavors I've tried, the GTK widgets are huge and barely fix in the cell while editing things like tone/mode on memories. This makes them a little wider by default so that it fits better. Curious what you're seeing and if this looks better.

Second, this allows you to jump straight to query_rrus (for example) from the command-line, which I assume will make it easier to test the RR stuff without so much mouse action. I know it sure would have for the RB stuff.

jebbgrenham:
Screenshot from 2022-12-14 12-36-42

kk7ds:
Actually, that's a different problem. I'm talking about editing fields directly in the spreadsheet.

This is before the change:

image


image

and this is after:

image

image

jebbgrenham:
Screenshot from 2022-12-15 11-51-50

kk7ds:
Cool, looks decent.

jebbgrenham:
Yup!

PR #251, opened 2022-12-15
Fix PosixPath returned from default_dir()

kk7ds:
The default_dir() needs to return a string, but #196 made it return
a PosixPath. This manifested as new users unable to open files, but
established users could, because last_dir was pulled from the config
instead of the platform module.


UtahDave:
This fix works for me on Ubuntu 22.04, python 3.10.6

kk7ds:
Awesome, thanks for the quick check @UtahDave !

PR #261, opened 2022-12-17
Add a super-basic printing implementation

kk7ds:
This is not something I have ever wanted, but it has been requested
plenty of times. Since wx makes this easy, this adds a super bare-
bones implementation that seems to work for me, which I think is
probably worthwhile. It doesn't handle multiple pages super elegantly,
and probably needs to have a "print selection" option, but it works.


kk7ds:
@KC9HI How many bugs do you think I could close with this one commit?

KC9HI:
At least 6

PR #278, opened 2022-12-18
Add TDXone TD-Q8A test report

KC9HI:
Adding DTCS Polarity changed the column ordering, which prevents us from searching the comment field. This fixes that, and the FIXME in that code to make the columns we search referenced by name so it does not happen again.


kk7ds:
Will fix

PR #285, opened 2022-12-19
Bring chirpc up to date with recent py3 changes

mfncooper:
Tested against Icom IC-910, including model detection.


kk7ds:
Appears to work for me with a Kenwood TH-D74 in live and clone mode

PR #289, opened 2022-12-19
Network source improvement

kk7ds:
- Add some FIPS utility functions
- Add NetworkResultRadio to query_sources
- Add MyGMRS as a network source
- Make DMRMARC query use the new pattern
- Make RepeaterBook use new source pattern


kk7ds:
FYI @jebbgrenham I think we need to do this, as CSV doesn't even support everything that usual GMRS repeaters need (which is another problem). But, this makes it so query result tabs don't have save/saveas (but you can export) and don't look like weird temp files, causing the save prompts at exit.

I didn't convert RR because I didn't want to create conflicts for you. I can do that right now if you want, or you can do it with the US stuff, or you can totally just keep doing what you're doing and I'll fix it all up when you're done.

PR #291, opened 2022-12-20
Mark Leixen VV-898S as tested in the py3 test matrix

READ10:


kk7ds:
Thanks Mark! You also need to update the actual matrix file. Let me just fix it for you, hang on.

PR #293, opened 2022-12-22
Add new driver tdh8 tdh8_gmrs tdh8_ham

Sandmann34:
Add three new dirver.
Vandor: TID
Model: TDH8 TDH8_GMRS TDH8_HAM


Sandmann34:
It seems that there is a problem with the unit test...

kk7ds:
Also, you need to put your new file in `tools/cpep8.manifest` so it will run the style checks against it.

Sandmann34:
> 你好，
> 
> 三个驱动文件几乎相同。请将它们合并为一个并子类化三个变体，而不是复制所有通用代码。此外，文件必须是 ASCII 格式，并带有英文注释，因为开发团队的其他成员将无法阅读中文字符。您还需要将测试图像添加到新驱动程序的测试/图像中（也许只有一个就足够了，因为它们非常相似）。使用测试图像，所有测试作业也必须通过。
> 
> 最后，这是一个针对 master 分支的补丁，除了 1 月 1 日的错误修复外，所有分支都将被冻结。如果你能针对 py3 分支提交这个补丁，那就更好了，这是 2023 年以后的所有构建都会做的建立于。如果 py3 分支中没有这些驱动程序的工作版本，我不会将其应用于 master 分支，因为它会产生回归。此处的提交需要在 py3 上工作（请参阅“针对 py3 的测试合并”检查），或者您应该针对 py3 分支本身提交。
> 
> 谢谢！

Hello，
How to use test images and how to add it?
i will commit needs to work on py3 

thanks!

kk7ds:
> How to use test images and how to add it?

Download an image from the radio and save it into `tests/images` formatted exactly as you see the other ones: `VENDOR_MODEL.img`. Add it to your commit. The file being in the right place will cause the **Driver Tests** to include that model and run tests against them.

The existing driver and unit test failures are because your driver detects a UV5R file as its own. Make your `model_match()` method always return `False`. Recent drivers do not need to implement their own matching as metadata in the image does the job. So just make that return False all the time and it will not accidentally match other radio images.

> i will commit needs to work on py3

Great!

PR #295, opened 2022-12-22
Add Retevis H777S

KC9HI:


chirp/drivers/radioddity_r2.py
```diff
@@ -631,3 +658,12 @@ class RT24Alias(chirp_common.Alias):
 @directory.register
 class RadioddityR2Generic(RadioddityR2):
     ALIASES = [RT24Alias]
+
+
+@directory.register
+class RetevisH777S(RadioddityR2):
+    """Retevos H777S"""
```
kk7ds, line 665: "Retevis" is mispelled.

chirp/drivers/radioddity_r2.py
```diff
@@ -502,10 +516,22 @@ def set_memory(self, mem):
...
-        LOG.warning('This driver may be broken for deleted memories')
+        _rsvd = _mem.reserved.get_raw()
+
         if mem.empty:
+            _mem.set_raw("\xFF" * 13 + _rsvd)
             return
 
+        if self._frs16:
+            if _mem.rx_freq.get_raw() == "\xFF\xFF\xFF\xFF":
+                FRS_FREQ = int(FRS16_FREQS[mem.number - 1] * 1000000)
```
kk7ds, line 527: Can we store those in Hz? They're probably okay, but storing frequencies like this in a `float` can be problematic because they don't have infinite resolution. That's why `Memory` stores integers for frequencies (except tune step, I know, but they're small).

chirp/drivers/radioddity_r2.py
KC9HI: Thanks.

chirp/drivers/radioddity_r2.py
KC9HI: I will give it a shot. Thanks.

kk7ds: Okay, not a huge deal as these are known and don't change I guess. But since you have to multiply them anyway, I think you might as well just store them as integers in Hz. I'll put my OCD away now :)

PR #301, opened 2022-12-23
Attempt to detect baud rate for icomciv driver

kk7ds:
This makes us try to detect the baud rate of the radio the first time
we communicate with it by fetching the template memory. If we fail,
we take the default set on the radio class. We also start with the
default on the radio class to speed up detection for the default
case.

This is annoying for users to have to set a specific rate, and by
default, a much slower one than they should need to use.

Tested with an IC-7300

### DIFF
--- a/chirp/drivers/icomciv.py
+++ b/chirp/drivers/icomciv.py
@@ -1,4 +1,5 @@
 # Latest update: April, 2021 Add hasattr test at line 564
+import serial
 import struct
 import logging
 from chirp.drivers import icf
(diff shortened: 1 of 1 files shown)


kk7ds:
@mfncooper Can you test this on the 910 to make sure it doesn't break anything? I assume the 910 is regular serial?

@n8sqt Maybe you want to give it a go as well?

n8sqt:
@kk7ds - Driver detected all baud rates except 4800 on an Icom IC-7300

kk7ds:
Ah, I can't believe it goes down to 4800, I didn't even check. I can add that to the list, although it'll just slow down the detection even more the more option it has to try. I guess we should probably include it, in case someone has some old CIV device they use their rig with. I'll put 4800 at the end of the list.

Thanks for testing!

n8sqt:
I think it's timing out at 4800. I can enter a few edits and it'll appear in the radio memories, then it'll balk (see memory 10).
Screenshot 2022-12-23 at 12 15 19 PM

9600 and 115200 worked. I didn't test the ones in between.

n8sqt:
Screenshot 2022-12-23 at 12 15 19 PM

n8sqt:
[chirp_debug-8rwl3nma.txt.zip](https://github.com/kk7ds/chirp/files/10297211/chirp_debug-8rwl3nma.txt.zip)

kk7ds:
Huh, that's bizarre, your log shows that the simulated echo is different on some of those edits near where it failed. I assume the radio is simulating the echo that a normal cable returns when the PC transmits. It's the right length, but the wrong data.

I guess I'm not sure it's worth really trying to debug that just to get 4800 baud to work, so I lean towards just taking that back out of the list. Maybe @mfncooper has an opinion.

I was mostly trying to get the 7300 to work at 115200 which is its default I think.

n8sqt:
I agree that 4800 isn't worth it. I can't imagine anyone running it that slow, unless you had a network of *very* early rigs (which sounds odd, anyway :-) )

mfncooper:
For the IC-910H, this does what it says on the tin, insofar as it picks 19200, the default, and goes no further. The one quirk on this radio is that the detection is currently not happening on the very first communication with the radio. That's because the very first communication is the detection of the presence or absence of the UX-910 option. If it really matters, you'd want to make a call to `_detect_baudrate()` from `_detect_23cm_unit()`.

kk7ds:
Ah, cool, I'll do that thanks for the confirmation.

n8sqt:
I first tried the 12/24 daily build without loading the module, and it worked only at 19200. I assume this PR wasn't merged yet when noting the times.

I then pulled and loaded the module from git and got an error when connecting to the 7300. Says something about encoding unicode bytes. This is with the 12/24 daily build and loading iclmciv.py pulled from git.

image

kk7ds:
Yep, not in today's build because I just merged it.

That error looks like you've got the `icomciv.py` from master and are trying to load it into a py3 build. Soon the py3 version will be master, but right now it's from the py3 branch, so you'll need the module from there.

n8sqt:
That would be it. When I did `git status` I didn’t see py3, so I assumed that was the right version. My bad. Coffee needs to kick in.

n8sqt:
Just added some WD40 to my rusty git use. 
All supported rates work on the IC-7300 (omitting 4800 as mentioned)

kk7ds:
Sweet, thanks!

PR #307, opened 2022-12-25
Make clone dialog use richer serial port names

kk7ds:
This improves the readability of the items in the box, especially
for nice devices that expose a user-friendly name.

I'm not happy about how this blows out the clone box width with
overly long descriptive names for ports, but I think the benefit
outweighs that, and perhaps something else can be done about it.

### DIFF
--- a/chirp/wxui/clone.py
+++ b/chirp/wxui/clone.py
@@ -19,6 +19,7 @@ import textwrap
 import threading
 
 import serial
+from serial.tools import list_ports
 import wx
 import wx.lib.sized_controls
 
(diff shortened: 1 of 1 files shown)


kk7ds:
@KC9HI Could you poke at this a bit and see if it works for you and looks right?

Absolutely no rush given the timing of course :)

KC9HI:
Tested with Windows 10 Pro & 4 different chip vendors. Looks good. Basically mirrors what is shown in Device Manager. Do you need me to look at Linux?

kk7ds:
Nope, it doesn't do the fancy stuff in Linux, but I've tested that it's (as good) as it was before. Just wanted another windows data point to make sure it worked okay for someone else.

Thanks!

PR #312, opened 2022-12-26
Attempt to parse random pasted text

kk7ds:
This is pretty error-prone, but if we find stuff in the text clipboard
that looks like frequency/tone data, try to parse it out into a
memory and paste it into the grid. Works pretty well for repeaterbook
lines, and a few other local repeater group websites I tried. It's
definitely not something we can do with a high degree of accuracy,
but if you just copy a frequency and tone, it does a decent job.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -20,6 +20,7 @@ import base64
 import json
 import logging
 import math
+import re
 import sys
 
 from chirp import errors, memmap, CHIRP_VERSION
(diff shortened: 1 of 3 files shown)


kk7ds:
@KC9HI FYI, this might be quite useful for helping people. When they post on email or facebook a repeater they're not sure how to enter, a helper can give them a string like this which will copy/paste into chirp like a memory:

```
442.325 +5 107.2/D025
```

even things like this will work:

```
[442.325/+5/107.2/D025]
```

KC9HI:
This is really slick. I was able to copy-and-paste from RepeaterBook to add all of the 2M/70CM repeaters in my county to a tab for a UV-5R.

It didn't work 100% for those from my local radio clubs web page because they are listed in the "Output/Input" format (as shown below).
`146.640 | 146.040 | 136.5`
This provided a 146.640 MHz simplex channel with a TX tone of 136.5 Hz. Which still isn't bad. Too bad it doesn't use the band plan to populate the Duplex/Offset fields when only the RX frequency is given (like CHIRP would do when just the RX frequency is entered directly into the spreadsheet).

But I would not expect to see many ham/gmrs repeaters listed in this manner.

kk7ds:
Ah, that is indeed a bit strange, but I should be able to make it handle that. Will give it a shot.

It does apply the band plan for offset if it finds only a duplex. The problem is that frequencies with no duplex listed that might should not have an offset/duplex are not distinguishable from just some weird format we don't recognize. It'll never be perfect, but I should be able to handle the TX frequency thing.

It'd be cool if we could get people start listing a compact "paste this into chirp" blob on their websites...

kk7ds:
@KC9HI Okay, try that...

KC9HI:
This perfectly parsed from RepeaterBook (frequency/offset/tone/tone) and RadioReference (frequency/frequency/tone/tone). This is so cool!

kk7ds:
Ah, very good, thanks. Hopefully it will work often enough for people not to be frustrated by it, but we'll see.

The change I just pushed will make sure we only pick up on +/- characters after the frequency for sites that list things like "440.350 +" but otherwise unchanged from what you tested (no need to re-test). There's a repeater site here with a dash in the name before the frequency and it would guess that for the duplex instead of the + that came after.

KC9HI:
It seems that currently CHIRP can copy-and-paste between memory rows, and obviously copy from an external app to CHIRP, but it doesn't seem to be able copy anything out to an external app. At some point in the future, would it be possible for CHIRP to copy-and-paste from a memory row to an external app in this (or a similar) format? 

`[442.325/+5/107.2/D025]`

That would allow copying a working memory row out to an email or social media platform to help someone.

kk7ds:
Yeah, I think I can publish both formats to the clipboard such that generic apps will grab the text format. I'll see what I can do.

KC9HI:
All was going well. I tested split frequencies, split tones (CTCSS & DTCS).
Anything that I could program into CHIRP was exported correctly.

Then I copy-and-pasted between tabs to make sure that was working
correctly. Everything looks good there.

Then I ran into a problem. I tried to copy-and-paste from an external
application back to CHIRP and nothing. This is what the console shows me.

<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to
find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to
find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to
find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to
find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[462.675000/+5.000000/136.5/136.5]':
Unable to find a frequency

PR #313, opened 2022-12-26
Copy memories in text format as well

kk7ds:
This allows for pasting into other applications for easy sharing
of memories.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -1661,23 +1661,23 @@ def urlretrieve(url, fn):
 
 def mem_from_text(text):
     m = Memory()
-    freqs = re.findall(r'\b([0-9]{1,3}\.[0-9]{2,5})\b', text)
+    freqs = re.findall(r'\b(\d{1,3}\.\d{2,6})\b', text)
     if not freqs:
         raise ValueError('Unable to find a frequency')
     m.freq = parse_freq(freqs[0])
-    if len(freqs) > 1:
+    offset = re.search(r'([+-])\s*(\d\.\d{1,3}|\d)\b', text)
+    duplex = re.search(r'\W([+-])\W', text[text.index(freqs[0]):])
+    if len(freqs) > 1 and not offset:
         split_to_offset(m, m.freq, parse_freq(freqs[1]))
     else:
-        offset = re.search(r'([+-])\s*\b([0-9](\.[0-9]{0,5})?)\b', text)
         if offset:
             m.duplex = offset.group(1)
             m.offset = parse_freq(offset.group(2))
         # Only look for the first +/- after the frequency, which would be
         # by far the most common arrangement
-        duplex = re.search(r'\W([+-])\W', text[text.index(freqs[0]):])
         if offset is None and duplex:
             m.duplex = duplex.group(1)
-    tones = re.findall(r'\b([0-9]{2,3}\.[0-9]|D[0-9]{3})\b', text)
+    tones = re.findall(r'\b(\d{2,3}\.\d|D\d{3})\b', text)
     if tones and len(tones) <= 2:
         txrx = []
         for val in tones:
@@ -1695,3 +1695,20 @@ def mem_from_text(text):
         split_tone_decode(m, txrx[0], txrx[1])
 
     return m
+
+
+def mem_to_text(mem):
+    pieces = [format_freq(mem.freq)]
+    if mem.duplex == 'split':
+        pieces.append(format_freq(mem.offset))
+    elif mem.duplex in ('-', '+'):
+        pieces.append('%s%i.%3.3s' % (mem.duplex,
+                                      mem.offset / 1000000,
+                                      '%03i' % (mem.offset % 1000000)))
+    txrx = split_tone_encode(mem)
+    for mode, tone, pol in txrx:
+        if mode == 'Tone':
+            pieces.append('%.1f' % tone)
+        elif mode == 'DTCS':
+            pieces.append('D%03i' % tone)
+    return '[%s]' % '/'.join(pieces)
--- a/chirp/wxui/main.py
+++ b/chirp/wxui/main.py
@@ -974,7 +974,7 @@ class ChirpMain(wx.Frame):
             wx.TheClipboard.Close()
         if gotchirpmem:
             self.current_editorset.cb_paste(memdata)
-        if got:
+        elif got:
             self.current_editorset.cb_paste(textdata)
 
     def _menu_selall(self, event):
--- a/chirp/wxui/memedit.py
+++ b/chirp/wxui/memedit.py
@@ -973,10 +973,10 @@ class ChirpMemEdit(common.ChirpEditor, common.ChirpSyncEditor):
             mem.number = self.row2mem(row)
             row += 1
             try:
-                mem = import_logic.import_mem(self._radio, srcrf, mem)
                 if mem.empty:
                     self._radio.erase_memory(mem.number)
                 else:
+                    mem = import_logic.import_mem(self._radio, srcrf, mem)
                     self._radio.set_memory(mem)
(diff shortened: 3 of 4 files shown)


kk7ds:
@KC9HI I haven't done a bunch of testing to make sure this works as expected and doesn't break existing copy/paste, but it seems to be okay. Let me know what you think!

KC9HI:
I can do more testing after supper, but everything I have tested works great.

kk7ds:
Cool, I'll poke a little later too, but I'll wait for your blessing. Since we're relying so heavily on copy/paste we definitely don't want to break it :)

KC9HI:
All was going well. I tested split frequencies, split tones (CTCSS & DTCS). Anything that I could program into CHIRP was exported correctly.

Then I copy-and-pasted between tabs to make sure that was working correctly. Everything looks good there.

Then I ran into a problem. I tried to copy-and-paste from an external application back to CHIRP and nothing. This is what the console shows me.

```
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[400.625000/136.000000]': Unable to find a frequency
<wx._core.TextDataObject object at 0x0000027206CE5480>
WARNING: Failed to parse pasted data '[462.675000/+5.000000/136.5/136.5]': Unable to find a frequency
```

kk7ds:
I was hitting the same thing, because I wasn't catching frequencies with the same number of digits as I was generating. Fixed now, and added tests to validate that our parsing is happy with our generation. Also, another bug was introduced when I moved to validating the pasted memories - empty rows fail the import test and generate errors. So that's fixed in this PR as well.

KC9HI:
Tried to paste this to CHIRP
`[462.675000/+5.000000/136.5/136.5]`
Popup: Tx freq 5.000000 is out of supported range

So apparently parsing +5.000000 as frequency instead of offset

kk7ds:
Ah, yeah, thanks..

kk7ds:
Okay, added some positional awareness so it won't match something after a +/- as a tx frequency, only an offset. I'm adding all these example cases to the unit tests to make sure they're all validated (now and in the future).

KC9HI:
`[450.000000/150.000000]`
Popup: Tx freq 5.000000 is out of supported range
`[400.000000/+9.900000]`
`[500.000000/-9.900000]`
Accepted but only shows offset of 9.000000
`[400.000000/+10.000000]`
`[500.0000
Accepted but only shows offset of 9.000000
`[400.000000/+10.000000]`
`[500.000000/-10.000000]`
Popup: 00/-10.000000]`
Popup: Tx freq 10.000000 is out of supported range

kk7ds:
Okay, I think all of those were fixed in the second to last push except for the +9.9 issue, but I just fixed that. Excellent catches!

KC9HI:
Getting better
'[450.000000/150.000000]'
Popup: Frequency 10.000000 is out of supported range

kk7ds:
`[450.000000/150.000000]` works for me on the latest. Does `git rev-parse --short HEAD` show `37f1f7b3` for you? If not you're not quite on the latest.

KC9HI:
Yes. 

C:\Users\Jim\mychirpgit\chirp>git rev-parse --short HEAD
37f1f7b3

And it still doesn't like offsets >= 10.000000

Could it be due to the radio I am using? I am loading the BTECH UV-25X2 and UV-25X4 images when testing.

kk7ds:
Ah, yep, it's the application of the band plan rules overwriting the offset. Maybe I need to make it only do that if duplex is detected as +/- but not split.

kk7ds:
Okay I think that's better now.

KC9HI:
Ok. All seems to be fixed except where offset [+/-] is >= 10 MHz
`[400.000000/+100.000000]`
`[500.000000/-100.000000]`
Popup: Tx freq 100.000000 is out of supported range

kk7ds:
Okay, but I thought there was a prohibition on splits more than like 70MHz on most radios? So I need to fix like 12MHz, but not 100Mhz, right? I don't think any of my radios can store splits more than 100MHz.

KC9HI:
I don't know about all radio, but using the UV-5R as an example. It was programmed with a 70 (69.990) MHz limit because the VFO had a 70 MHz limit. This meant that the user couldn't use the VFO mode for splits larger than 70 MHz or cross band operation. This was later change to allow the user to set the VFO offset to 999.99 MHz. But the radio channels have always supported a repeater offset as high as the "offset" between the band limits. So the 70 MHz limit programmed into CHIRP is arbitrary and has never matched the capability of the radio. I noticed that there are many driver that contain this 70 MHz limit. I know on at least one driver I set it up to where if the 2 frequencies are in the same band, I have CHIRP convert it to (+/-) offset and if they are in different band split / tx frequency.

kk7ds:
Okay, I'm thinking mostly of the japanese radios that actually store an offset instead of emulating it. I guess because I require the +/- to come before the offset (and not a txfreq) then I should be able to allow for up to 999MHz offsets. In a bit I'll do that and add some tests and push up if so.

KC9HI:
I always wondered what the reason for the 70 MHz limit was. You explanation makes perfect sense.

kk7ds:
Okay, so supporting >10 and especially >100MHz splits makes the matching confuse tone values (like 88.5, 162.2) for offsets in strings like "440.35 + 162.2" which worked before. I think that non-standard splits working in this best-effort copy/paste stuff are probably not that important and so we'll be correct a lot more often if we only support offsets like 0.6, 1.2, 5.0, etc. Do you think that's okay?

I should also probably make the string we generate when we copy for external programs not have so many digits on the offset in that case as that is really where some of this complexity is coming from.

KC9HI:
Yes. Common offsets should be fine. Anything other then the common 0.6, 1.6, 5.0, etc. is probably going to be formatted as rx_freq/tx_freq anyway.

kk7ds:
Yeah, I think the only example I can think of is the 33cm band which uses 25MHz offsets. But that could be noted as tx/rx as you say.

If this catches on, I can parse the `[a/b/c/d]` format specifically and then all these problems go away. I'm just trying to make the "copy and paste generic stuff from a website" parsing code as smart as it can be, but there are limits to how well it can do.

KC9HI:
If I past this faux frequency/offset into CHIRP the offset is change to 1.000000 MHz
`[500.000000/-1.600000]`
However, If I use this valid ham repeater from RepeaterBook, it is entered fine.
`224.6400 | -1.6 MHz | 107.2 / CSQ`

kk7ds:
Yeah, I changed the offsets in that format to only three decimal digits, so `0.6`, `0.60`, `0.600` and `5.000` - anything longer like `0.6000` won't match which helps distinguish between other formats. Everything I've seen on people's websites seems to be short enough that three digits suffices. Copy from chirp should use three decimal digits now too.

Sounds like we should be good now, thanks very much for helping stress test that!

KC9HI:
Yep. Taking that into account. Everything work!

KC9HI:
Well except when you go from CHIRP to external app the result is -1.600000 (which won't cut-and-paste back in)

kk7ds:
Yep, crap. When I said "I shortened offset to three digits" I meant I did that in the parsing code in the commit, and in the generation code in my *mind* :)

KC9HI:
That's got it.

PR #317, opened 2022-12-27
Driver update: bf_t8.py

KC9HI:
* Switch frequency tables to Hz based
* Remove no longer needed validate_memory() code
* Remove no longer needed mem.immutable code


kk7ds:
Rebased with just a click, so should be easy.

KC9HI:
Great! I will get the other 2 projects sent and then switch to using master.

PR #323, opened 2022-12-28
Change executable target in .desktop to the wxui entry point

vishwin:


kk7ds:
...oh yeah, thanks :)

PR #327, opened 2022-12-28
Send users to the ChirpNextBuild page for updates

kk7ds:
Until the proper Download page includes links to the CHIRP-next build,
we should send them to ChirpNextBuild, which includes the download
link and may include updated information over time which would be
good to get in front of early adopters as well.

This also makes us warn the user to quit CHIRP before installing the
next build, as that is a common stumbling block for users.


kk7ds:
CC @KC9HI 

The "be sure to quit first" prompt is just a warning - we could force quit after we send them to the page if you think that would be better, but it might be annoying if they click "ok" and not realize that they're now destined to quit...

PR #329, opened 2022-12-29
BJ-218/318 and IC-2730 with matrix.

Old-Phart:


chirp/drivers/ic2730.py
```diff
@@ -295,6 +296,7 @@ class IC2730Radio(icf.IcomCloneModeRadio):
     _can_hispeed = True
     _raw_frames = True
     _highbit_flip = True
+    _MUNCH_CLONE_RESP = True
```
kk7ds, line 299: This flag was added for one specific radio that the author claimed needed this, but I was always skeptical. Especially with some of the changes I made recently to this module to fully parse the entire stream from the radio, I really don't think this should be necessary. With two other similiar-vintage raw-speaking icom radios working well (5100 and 4100) I'm really not wanting to do this unless there's a really good reason. Can you tell me why you think we need to add this? Like, I want to see logs of failures :)

Is it possible this was left-over from some debugging and isn't actually necessary?

chirp/drivers/lt725uv.py
```diff
@@ -1338,7 +1341,8 @@ def make_dtmf(ary, knt):
             """Generate the DTMF code 1-8, NOT a callback."""
             tmp = ""
             if knt > 0 and knt != 0xff:
-                for val in ary[:knt]:
+                for j in range(0, knt, 1):
+                    val = ary[j]
```
kk7ds, line 1345: Can you explain what's going on here? It looks to me like this change should have no effect, and I think the parent radio is already tested as working, so is this just a cosmetic change or functional?

tests/py3_driver_testers.txt
```diff
@@ -23,6 +23,8 @@ AnyTone_778UV_VOX,+Retevis_RT95,13-Nov-2022
 AnyTone_5888UV,@kk7ds,9-Dec-2022
 AnyTone_OBLTR-8R,@KC9HI,17-Dec-2022
 AnyTone_TERMN-8R,@KC9HI,17-Dec-2022
+Baojie_BJ-218,@AA0RD,28-Dec-2022
```
kk7ds, line 26: The @ thing needs to reference a github ID. So you should put @Old-Phart here so that the generated page links properly.

kk7ds (commented):
Thanks for getting this up Rick!

You need to commit the generated matrix file as well in order to satisfy the tests. So:

```
$ tox -e makesupported
$ git commit -a --amend
```

to include it. I can also amend this commit for you with that if you're not able to run it locally for some reason.

I really want to clarify the icf thing though as I really want to avoid making that change unless there's some reason.

Thanks!

chirp/drivers/lt725uv.py
kk7ds: I'm still curious about this, especially since it changes a driver that we already think is tested. @KC9HI does this look necessary or beneficial for some reason I'm missing?

kk7ds:
Thanks for the update to the matrix, that looks good now, and the 2730 change. I just want to make sure we're on the same page with the `lt725uv.py` change and we should be good.

Thanks Rick!

chirp/drivers/ft450d.py
```diff
@@ -438,7 +481,7 @@ def _clone_in(self):
 
         start = time.time()
 
-        data = ""
+        data = bytes(b"")
```
kk7ds, line 484: Just `b""` is enough here.

chirp/drivers/ft450d.py
```diff
@@ -451,13 +494,14 @@ def _clone_in(self):
             else:
                 repeat = 1
             for _i in range(0, repeat):
-                data += self._read(block, blocks)
-                self.pipe.write(chr(CMD_ACK))
+                chunk = self._read(block, blocks)    # returns bytes()
+                data += chunk
+                self.pipe.write(bytes(chr(CMD_ACK), encoding='utf8'))
```
kk7ds, line 499: This is wrong, you should not encode binary data as a UTF-8 string. What you want is either:

```
CMD_ACK = b'\x06'
self.pipe.write(CMD_ACK)
```
or
```
CMD_ACK = 0x06
self.pipe.write(bytes([CMD_ACK]))
```

chirp/drivers/ft450d.py
```diff
@@ -451,13 +494,14 @@ def _clone_in(self):
...
-                self.pipe.write(chr(CMD_ACK))
+                chunk = self._read(block, blocks)    # returns bytes()
+                data += chunk
+                self.pipe.write(bytes(chr(CMD_ACK), encoding='utf8'))
                 blocks += 1
                 status.cur = blocks
                 self.status_fn(status)
-        data += self.MODEL      # Append ID
-        return memmap.MemoryMap(data)
+        data += bytes(self.MODEL, encoding='utf8')      # Append ID
```
kk7ds, line 503: It would be more conventional to use `self.MODEL.encode()` here. However, you do *not* want a UTF-8 string, you want a direct conversion, so `self.MODEL.enocde('ascii') would be better.

chirp/drivers/ft450d.py
```diff
@@ -451,13 +494,14 @@ def _clone_in(self):
...
+        return memmap.MemoryMapBytes(data)
```
kk7ds, line 504: This is good, but you need to set `NEEDS_COMPAT_SERIAL=False` on your driver if you do this, because the UI uses that flag to determine whether or not to give you a `MemoryMapBytes` or not when loading from a file. The two need to be consistent.

chirp/drivers/ft450d.py
```diff
@@ -479,22 +523,27 @@ def _clone_out(self):
...
+                dbg3 = bytes(chr(xs), encoding='latin_1')
+                self.pipe.write(dbg1)
+                self.pipe.write(dbg2)
+                self.pipe.write(dbg3)
+                buf = bytes(self.pipe.read(1))
+                if not buf or buf[0] != CMD_ACK:
                     time.sleep(delay)
-                    buf = self.pipe.read(1)
-                if not buf or buf[0] != chr(CMD_ACK):
+                    buf = bytes(self.pipe.read(1))
```
kk7ds, line 539: `pipe.read` returns bytes already, no need to convert it...

chirp/drivers/ft450d.py
```diff
@@ -921,7 +984,7 @@ def _set_memory(self, mem, _mem):
 
         for setting in mem.extra:
             if setting.get_name() == "notch_pos":
-                vx = 0          # Override list string with signed value
+                vx = 0          # Overide list string with signed value
```
kk7ds, line 987: Adding a misspelled word

chirp/drivers/ft450d.py
```diff
@@ -939,13 +1002,12 @@ def _set_memory(self, mem, _mem):
...
                 setattr(_mem, setting.get_name(), setting.value)
 
-
     @classmethod
     def match_model(cls, filedata, filename):
         """Match the opened/downloaded image to the correct version"""
         if len(filedata) == cls.MEM_SIZE + 7:    # +7 bytes of model name
             rid = filedata[cls.MEM_SIZE:cls.MEM_SIZE + 7]
-            if rid.startswith(cls.MODEL.encode()):
+            if rid.startswith(cls.MODEL):
```
kk7ds, line 1010: `filedata` is a `bytes` here, so you can't compare it with `cls.MODEL`, which is a string. You're failing the detect test as a result...

chirp/drivers/ft450d.py
```diff
@@ -29,9 +30,16 @@
...
-# That field wiould not be stored in img file, but generated in get_memory
+HAS_FUTURE = True
+try:                         # PY3 compliance
+    from builtins import bytes
+    from past.builtins import range     # RJD 12/18/22
+except ImportError:
+    HAS_FUTURE = False
+    LOG.warning('python-future package is not '
+                'available; %s requires it' % __name__)
+
```
kk7ds, line 41: Why are you adding this to the py3 branch? The behavior of `bytes` in python3 is natively what it is with the future package on python2. Using future to allow a driver to work on both is good, but at this point, these changes are all python3 so adding this code doesn't do anything useful.

chirp/drivers/ft450d.py
```diff
@@ -29,9 +30,16 @@
 
 LOG = logging.getLogger(__name__)
 
-CMD_ACK = 0x06
-# TBD: Enable some form of generated UI field, for the memory tags
-# That field wiould not be stored in img file, but generated in get_memory
+HAS_FUTURE = True
+try:                         # PY3 compliance
+    from builtins import bytes
+    from past.builtins import range     # RJD 12/18/22
```
kk7ds, line 36: Why are you importing this? This makes this code python3-uncompliant. If you're using range below and need a list, then you should make it a list.

chirp/drivers/ft450d.py
```diff
@@ -42,30 +50,65 @@
...
+            if valx > 31 and valx < 127:
+                sx += "%s" % chrx
+            else:
+                sx += "."
+        sx += "]"
+    if lgx & 1:
+        LOG.warning("%s: %s" % (cptn, sx))
+    if lgx & 2:
+        LOG.debug("%s: %s" % (cptn, sx))
+    return
```
kk7ds, line 85: Is this necessary? Just using the existing hexprint function with a slice should be fine:

```
print(util.hexprint(buffer[:32]))
```

chirp/drivers/ft450d.py
```diff
@@ -1493,3 +1637,24 @@ def set_settings(self, settings):
...
+class FT450Alias(chirp_common.Alias):
+    """ ALias for Yaesu FT-450 original radio """
+    VENDOR = "Yaesu"
+    MODEL = "FT-450"
+
+
+class FT450ATAlias(chirp_common.Alias):
+    """ ALias for Yaesu FT-450AT radio with tuner """
+    VENDOR = "Yaesu"
+    MODEL = "FT-450AT"
```
kk7ds, line 1651: We're trying not to add aliases anymore, and using subclasses from now on since the clone box handles those properly.

Are you adding these because the communication or feature set is actually different? If not, I think it would be best to not just clutter the download box with choices that are all equivalent. It's one thing if one radio is listed under two completely different models or vendors, but these are almost identical.

chirp/drivers/ft450d.py
```diff
@@ -1493,3 +1637,24 @@ def set_settings(self, settings):
...
+
+
+if HAS_FUTURE:    # Only register driver if environment is PY3 compliant
```
kk7ds, line 1654: This code is all python3 only so, you don't need to do this.

kk7ds (commented):
Hey Rick, thanks for all the work here. I think it would probably be best to separate the 2730 and BJ-218 stuff as separate PR so we can get that merged, because I think the FT-450 work will require a lot more discussion.

Also, it looks like the FT-450 work is more than just python3 fixes? It would really be much better if we can keep the python3 conversion separate from feature/function stuff like @KC9HI and others have been doing. If we get a bug report on the new version for this driver, it will be a lot harder to narrow down whether or not the problem is related to python3 changes or those other changes if they're all mixed together.

If you can address the question on the BJ-218 thing, I can help split out the 2730 and BJ218 stuff if needed so we can get that merged if you're okay with it.

PR #332, opened 2022-12-29
Python 3.9 fixups

asheplyakov:
As of now CHIRP fails to start with Python 3.9:
- Complains about missing `importlib_resources` module (although `importlib` is available as a part of the standard library)
- Application icon and stock configs can't be found (since `chirp.share` is not a proper python package)

With this patchset I can use CHIRP with Python 3.9.

### DIFF
--- a/chirp/wxui/__init__.py
+++ b/chirp/wxui/__init__.py
@@ -5,9 +5,9 @@ import logging
 import os
 import sys
 
-if sys.version_info < (3, 10):
+try:
     import importlib_resources
-else:
+except ModuleNotFoundError:
     import importlib.resources as importlib_resources
 
 from chirp import directory
--- a/chirp/wxui/main.py
+++ b/chirp/wxui/main.py
@@ -24,9 +24,9 @@ import tempfile
 import time
 import webbrowser
 
-if sys.version_info < (3, 10):
+try:
     import importlib_resources
-else:
+except ModuleNotFoundError:
     import importlib.resources as importlib_resources
 
 import wx
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(name='chirp',
           'pyserial',
           'six',
           'future',
-          'importlib-resources;python_version<"3.10"',
+          'importlib-resources;python_version<"3.9"',
           'yattag',
       ],
       extras_require={


kk7ds:
You need to install the packages in `requirements.txt` to get the importlib-resources compatibility package. The macos build is already using python 3.8 and it works fine that way.

https://github.com/kk7ds/chirp/blob/master/requirements.txt#L9

asheplyakov:
Hi, 29.12.2022, 09:41, "Dan Smith" ***@***.***>:> You need to install the packages in requirements.txt to get the importlib-resources compatibility package. I've tried that (via pip install -r requirements.txt), the application failed to start anyway.I haven't bothered to find out the exact cause, instead I tweaked imports to use theimportlib.resources from the stdlib, and it worked for me. Also setup.py skips chirp/share and chirp/stock_configs directories (they arenot python packages), as a result the application can't find icon and fails to start. > The macos build is already using python 3.8 and it works fine that way.> https://github.com/kk7ds/chirp/blob/master/requirements.txt#L9    —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

kk7ds:
They're good enough for the bundler to pick up - I didn't really think they had to be proper python packages to be considered resources, but I dunno. Maybe @vishwin has some context here, but I can poke with python3.9 a little later. I thought that the thing we specifically needed was only in 3.10...

vishwin:
The new-style API was introduced in Python 3.10, and to simplify maintenance (especially when Python 3.11 formally marked the old-style but imo more concise API deprecated), the package is used for Python < 3.10.

vishwin:
As for package detection itself, setuptools does (or at least did) consider `chirp.share` a package implicitly, but of course explicit is always better than implicit.

kk7ds:
@asheplyakov Is it possible that just the `__init__.py` additions make it work for you? Like @vishwin it surely seems like 3.9 _should_ require `importlib_resources` to work properly. Have you verified that your `pip install` command is actually installing that lib and that it's importable via the python that you're using to run chirp?

kk7ds:
On a clean python 3.9 docker image, `python3 setup.py install` installs  `importlib-resources` as a dep as expected, and I can launch `chirp` from the installed directory just fine and it runs, finds the stock stuff, icons, etc, which I can confirm is installed to the proper location.

So, I'm not sure what the problem is, but I don't think it's specific to 3.9.

goldstar611:
> `python3 setup.py install` installs importlib-resources as a dep as expected ...

Installing via `setup.py` directly is deprecated. `pip3 install [directory]` seems to be the way forward.

You should see an error message like the following when you invoke setup.py directly:
```
SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
```

Sorry for interjecting...

vishwin:
That's not an error, but a warning. And yes PEP-517 compliance is next on the list.

vishwin:
On second look, both of these commits are wrong. The `importlib_resources` package has to be used for Python < 3.10, not only for the new-style `files()` and `as_files()` API introduced in 3.9, but PEP-420 namespace package support only landed in time for 3.10. Both `chirp.share` and `chirp.stock_configs` are namespace packages, which are not to have `__init__.py`.

kk7ds:
> On second look, both of these commits are wrong. The `importlib_resources` package has to be used for Python < 3.10, not only for the new-style `files()` and `as_files()` API introduced in 3.9, but PEP-420 namespace package support only landed in time for 3.10. Both `chirp.share` and `chirp.stock_configs` are namespace packages, which are not to have `__init__.py`.

Yep, this is exactly what I meant when I said "I thought that the thing we specifically needed was only in 3.10...". Further confirmed, I think, by the fact that it works fine for me in 3.9.

asheplyakov:
> > `python3 setup.py install` installs importlib-resources as a dep as expected ...

Yes. But it skips share and stock_configs (unless there is the `__init__.py` file in the directory)


> Installing via `setup.py` directly is deprecated. `pip3 install [directory]` seems to be the way forward.


1. pip calls `setup.py install` anyway, and prints the corresponding warning:
```
  removing build/bdist.linux-x86_64/wheel
  Building wheel for chirp (setup.py) ... done
  Created wheel for chirp: filename=chirp-py3dev-py3-none-any.whl size=1274863 sha256=1ab33209db1015ce3e02b9300a5daeef568b97d1d150da2ed31f8a7a7bf929f3
  Stored in directory: /usr/src/tmp/pip-ephem-wheel-cache-zqphcjaw/wheels/20/54/1a/b3128b303c56af26a8ebbd9f771961fb0e2c24b0734d4c6b09
  WARNING: Built wheel for chirp is invalid: Metadata 1.2 mandates PEP 440 version, but 'py3dev' is not
Failed to build chirp
Installing collected packages: chirp
[skipped]
  Running command Running setup.py install for chirp
  /usr/lib64/python3/site-packages/setuptools/dist.py:499: UserWarning: The version specified ('py3dev') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.

  Running setup.py install for chirp ... done
  DEPRECATION: chirp was installed using the legacy 'setup.py install' method, because a wheel could not be built for it. A possible replacement is to fix the wheel build issue reported above. Discussion can be found at https://github.com/pypa/pip/issues/8368
Successfully installed chirp-py3dev

```

2. I'm building an rpm package. python3 -m `pip install` is not the best tool for the task, so I use `pyproject_installer` instead like this
```
python3 -m pyproject_installer -v build
python3 -m pyproject_installer -v install --destdir=%buildroot
```

This one  successfully builds a wheel and dist-info.

vishwin:
> Yes. But it skips share and stock_configs (unless there is the `__init__.py` file in the directory)

If that was the case, I would not have had implemented this the way it is.

> I'm building an rpm package

Any chance you have some outdated components in your chain by virtue of using RPMs provided by your operating system's repositories?

asheplyakov:
> > Yes. But it skips share and stock_configs (unless there is the `__init__.py` file in the directory)
> 
> If that was the case, I would not have had implemented this the way it is.

Apparently it didn't work as intended.
 
> > I'm building an rpm package
> 
> Any chance you have some outdated components in your chain by virtue of using RPMs provided by your operating system's repositories?

Which components should I check, and what are the correct versions? (the only versioned dependency in requirements.txt is `wxPython>=4.0,<4.2.0 ; platform_system=="Linux"`, and I use wxPython 4.0.7)

kk7ds:
> > > Yes. But it skips share and stock_configs (unless there is the `__init__.py` file in the directory)
> > 
> > 
> > If that was the case, I would not have had implemented this the way it is.
> 
> Apparently it didn't work as intended.

As noted, it does for me, on python 3.8, 3.9, and 3.10.

kooscode:
I am trying to run chirp on Python3 to help test the drivers, but seems the installer doesnt work and is looking for python package `wxPython` which i cant install because somehow its looking for GTK2 ?

Any pointers on getting this running on Python 3.8 ?

kk7ds:
> I am trying to run chirp on Python3 to help test the drivers, but seems the installer doesnt work and is looking for python package `wxPython` which i cant install because somehow its looking for GTK2 ?
> 
> Any pointers on getting this running on Python 3.8 ?

It would help to know what system you're running, but it works fine on python 3.8 without anything specific. See this page: https://chirp.danplanet.com/projects/chirp/wiki/DevelopersPython3Environment

Probably best to discuss further on the devel mailing list, separate from this PR.

kooscode:
> 

Ah - good point. sorry.  I am running Ubuntu 20.04

kooscode:
> It would help to know what system you're running, but it works fine on python 3.8 without anything specific. See this page: https://chirp.danplanet.com/projects/chirp/wiki/DevelopersPython3Environment

Seems there is no branch `py3` for this repo and the above instructions call out to checkout that branch?

`fatal: 'origin/py3' is not a commit and a branch 'py3' cannot be created from it`

vishwin:
> Seems there is no branch `py3` for this repo and the above instructions call out to checkout that branch?
> 
> `fatal: 'origin/py3' is not a commit and a branch 'py3' cannot be created from it`

Because `py3` is now the main branch.

goldstar611:
After a [very lengthy build process](https://www.wxpython.org/blog/2017-08-17-builds-for-linux-with-pip/index.html), I was able to get CHIRP and all of it's dependencies installed on a fresh Ubuntu 20.04 VM using the commands below. (It is probably not necessary to run pip as root and you'll see a message discouraging it.)

```
sudo apt install python3-pip libgtk-3-dev libpng-dev libtiff-dev libjpeg-dev libcurl4-openssl-dev
sudo -H pip3 install attrdict
sudo -H pip3 install "chirp[wx] @ git+https://github.com/kk7ds/chirp.git"
```

kk7ds:
Guys, on linux, don't install wxpython from pip - it builds the entire thing for no reason. Follow the instructions on the build page, install the distro package and then do the pip stuff. Should be two seconds.

goldstar611:
Ah, I knew there should be a faster way. Thanks for the tip Dan. And now I understand why `wx` is listed as an extra in setup.py

From https://chirp.danplanet.com/projects/chirp/wiki/DevelopersPython3Environment:
```
sudo apt install git python3-wxgtk4.0 python3-serial python3-six python3-future python3-requests python3-pip
```

then
```
pip install git+https://github.com/kk7ds/chirp
```

Another confirm that the current tip of master installs on Ubuntu 20.04 with python 3.8.

PR #333, opened 2022-12-29
Fix parsing and formatting frequencies in wxui

kk7ds:
For some reason while writing the wxui I didn't use the existing
methods for parsing and formatting frequencies back and forth between
floating-point MHz strings and integer values of hertz. Without those,
higher frequencies start to be affected by rounding error.


kk7ds:
CC @KC9HI

PR #347, opened 2023-01-01
Fix prompting for offset when duplex=off

kk7ds:
When the user selects a duplex of off, we could have found a
zero offset (either because it was already set on the memory,
or because the band plan gave us a default) and fall through to the
prompt for an offset value. This makes us specifically bail early
if "off" is selected, as no prompt for offset/txfreq would ever be
needed in that case.


kk7ds:
CC @KC9HI 

I finally experienced it (platform-independent). I was testing by tweaking a UHF memory in the repeater range, where I was getting a default offset from the band plan and thus never experienced the behavior. Once (by chance) did it on a VHF one outside the repeater band, I saw it. I should have thought to ask for exact settings on the row in question, but didn't.

Hopefully this should resolve it, thanks for reporting :)

KC9HI:
Nice. Thank you for tracking it down.

PR #361, opened 2023-01-05
wxui improvements

kk7ds:
- Fix printing window parent
- Add separate print and print preview menu items
- Support delete memories from the edit menu
- Disable cell editing for immutable memory fields
- Allow selecting fixed/variable fonts for memedit

### DIFF
--- a/chirp/wxui/bankedit.py
+++ b/chirp/wxui/bankedit.py
@@ -20,7 +20,10 @@ import wx.grid
 
 from chirp import chirp_common
 from chirp.wxui import common
+from chirp.wxui import config
 from chirp.wxui import memedit
+
+CONF = config.get()
 
 
 if platform.system() == 'Linux':
@@ -89,6 +92,8 @@ class ChirpBankEdit(common.ChirpEditor):
                                    family=wx.FONTFAMILY_TELETYPE,
                                    style=wx.FONTSTYLE_NORMAL,
                                    weight=wx.FONTWEIGHT_NORMAL)
+        self._variable_font = self._grid.GetDefaultCellFont()
+        self.update_font(False)
 
         for col, col_def in enumerate(self._col_defs):
             self._grid.SetColLabelValue(col, col_def.label)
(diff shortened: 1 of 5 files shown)


kk7ds:
@KC9HI Do you have a few minutes to try this before I merge? It should fix your printing warning message. I've tested the font options (view menu) on windows, mac and linux now and they look right to me. But if you could confirm just to be extra sure, that would be awesome.

Also, this introduces something I think you might be interested in, which the old UI didn't even do. If you set immutable memory fields, the new UI will just disable the editing of those cells in the spreadsheet. That avoids the user thinking they can edit something only to be slapped on the wrist when doing so. I think we probably have a lot more places where we could/should use that for special models like GMRS-only radios, etc. PR #357 wants to add a bunch of memories with things that can't be changed, which is why I'm adding this now.

KC9HI:
I will be trying this soon.

KC9HI:
Tested printing on Windows 10. Worked and looked good. The driver I chose didn't have a lot of populated memory rows, but the printout (my first one) looks great.

Testing on Linux Mint is the same as it was this morning. The preview looks solid black but if I look very close I can barely see that there are some grid lines.

kk7ds:
Oh I meant to say - I dunno what is up with the black screen, but I see the same. On one of my linux boxes if I resize that preview window it clears up, but it also doesn't seem to affect things (and that trick doesn't work on the other). That's why I isolated the print and print-preview options just to avoid needing to go through that to get to printing. The preview window is provided by wx, not me, so it's somewhat out of my hands. On macos at least the system print dialog has PDF and preview stuff built in so it's not even necessary :)

KC9HI:
I thought there was only Print this morning but I was afraid to say anything. LOL

On my Linux Mint, if I choose Print in CHIRP, the built-in printer driver provides a [Preview] button that works perfectly.

kk7ds:
Okay, maybe I should just remove the preview thing then.

Did you try the view->font options? That's really what I wanted to confirm, since it's such a lightning rod of an issue :P

KC9HI:
Hang on. I just tried on Mint and it looks/works great. Its very nice that it instantly switches. I expected it was going to be a configuration setting that would require a restart of CHIRP like Developer Mode.

KC9HI:
I thought I broke them for a second. Then I realized that they are greyed out when the Memories tab is not selected. Seems to work equally well in Windows.

kk7ds:
Cool, thanks. I disable them on non-memory tabs just so the user isn't flipping them and wondering why the settings tab isn't changing. I might at least enable the size one to affect the settings tab at some point.

PR #363, opened 2023-01-07
Fix crash on linux due to missing print preview

kk7ds:
On Linux we don't add the print preview menu item, but the window
update function tries to en/disable it anyway.

Fixes #10236


goldstar611:
Thanks for the quick fix! I'll open an issue on the redmine tracker next time so there's less leg work to do on your end.

PR #368, opened 2023-01-08
Fix radio browser on Windows

kk7ds:
This makes the browser panels proper children of the TreeBook. Before,
they were children of the TreeBook's parent and thus were fully
overlapping with the TreeBook such that they could not be clicked on.
This only manifested on Windows for some reason, but was wrong
regardless.

Fixes #10243


kk7ds:
@KC9HI Are you able to test this real quick? I was able to reproduce this and this fixes it for me, but... would like confirmation if you can.

KC9HI:
It seems to be working now. Thanks.

PR #374, opened 2023-01-10
kguv9dplus frequency limit editing, brightness bug fix, and python3 compatibility

samwich:
* [kguv9dplus] Fix some python3 integer division and b strings
* [kguv9dplus] make _do_upload() work with python3
* [kguv9dplus] Add frequency limit settings
* [kguv9dplus] fix brighness 5 settings screen crash
* [kguv9dplus] fix code formatting

### DIFF
--- a/chirp/drivers/kguv9dplus.py
+++ b/chirp/drivers/kguv9dplus.py
@@ -738,6 +738,8 @@ WORKMODE_LIST = ["VFO freq", "Channel No.", "Ch. No.+Freq.",
                  "Ch. No.+Name"]
 BACKLIGHT_LIST = ["Off"] + ["%sS" % t for t in range(1, 31)] + \
                  ["Always On"]
+BACKLIGHT_BRIGHT_MIN = 1
+BACKLIGHT_BRIGHT_MAX = 5
 SAVE_MODES = ["Off", "1", "2", "3", "4"]
 LOCK_MODES = ["key-lk", "key+pg", "key+ptt", "all"]
 APO_TIMES = ["Off"] + ["%dm" % t for t in range(15, 151, 15)]
@@ -751,7 +753,6 @@ ID_DLY_LIST = ["%dms" % t for t in range(100, 3001, 100)]
 VOX_GRDS = ["Off"] + ["%dlevel" % l for l in range(1, 11)]
 VOX_DLYS = ["Off"] + ["%ds" % t for t in range(1, 5)]
 RPT_KPTS = ["Off"] + ["%dms" % t for t in range(100, 5001, 100)]
-LIST_1_5 = ["%s" % x for x in range(1, 6)]
 LIST_0_9 = ["%s" % x for x in range(0, 10)]
 LIST_1_20 = ["%s" % x for x in range(1, 21)]
 LIST_OFF_10 = ["Off"] + ["%s" % x for x in range(1, 11)]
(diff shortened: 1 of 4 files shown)


kk7ds:
Hi Sam, thanks for this, and thanks for splitting this into logical pieces :)

I assume you tested this with an actual radio and that it up/downloads properly now? If so, can you add yourself to `tests/py3_driver_testers.txt`? After you do, you need to run `tox -emakesupported` to generate the matrix file, which you can add to your commit.

Also, the conversion to py3 should set `NEEDS_COMPAT_SERIAL=False` on the driver class, and with that, use `MemoryMapBytes` instead of `MemoryMap` (presumably in the download routines). Could you try to do that, or if I do it, could you test?

Thanks!

samwich:
Regarding `tests/py3_driver_testers.txt`:
I have tested this driver with a KG-UV9D **Mate** radio, but the driver is written for KG-UV9D **Plus**. 
Should I add two lines, one for each radio?

```
Wouxun_KG-UV9D_Mate,+Wouxun_KG-UV9D_Plus,10-Jan-2023
Wouxun_KG-UV9D_Plus,@samwich,10-Jan-2023
```

note: This is the first Python I've ever written, so please be patient :-)

I added `NEEDS_COMPAT_SERIAL = False`, and switched to `MemoryMapBytes`. I got uploading to the radio working with a small change. 

Downloading from the radio is failing with:
```
  File "/home/ubuntu/src/chirp/chirp/memmap.py", line 30, in __init__
    assert isinstance(data, bytes)
```
which is being called from the end of `_do_download()`
```
        strmem = "".join([chr(x) for x in mem])
        return memmap.MemoryMapBytes(strmem)
```
Do you have some advice for converting `strmem` to something `MemoryMapBytes`-compatible? 
Let me know if you'd rather give me tech support or just make the change yourself :-)

kk7ds:
> Regarding `tests/py3_driver_testers.txt`: I have tested this driver with a KG-UV9D **Mate** radio, but the driver is written for KG-UV9D **Plus**. Should I add two lines, one for each radio?

Is there any difference between the two? If not, we can add an alias on the website so people know they are the same. There are a ton of rebadged chinese models, so if it works perfectly, it's probably the same :)

@KC9HI or @mrtjr1 might know.

> Wouxun_KG-UV9D_Mate,+Wouxun_KG-UV9D_Plus,10-Jan-2023
> Wouxun_KG-UV9D_Plus,@samwich,10-Jan-2023

You can't do this because the tool will enforce that the model in the first field matches a registered class. There's another way to get an alias on the website so people know. We can handle that afterwards.

> I added `NEEDS_COMPAT_SERIAL = False`, and switched to `MemoryMapBytes`. I got uploading to the radio working with a small change.
> 
> Downloading from the radio is failing with:
> 
> ```
>   File "/home/ubuntu/src/chirp/chirp/memmap.py", line 30, in __init__
>     assert isinstance(data, bytes)
> ```
> 
> which is being called from the end of `_do_download()`
> 
> ```
>         strmem = "".join([chr(x) for x in mem])
>         return memmap.MemoryMapBytes(strmem)

Cool, this is why we need to flip that flag, because it was only working before with janky compat code ;) I think just remove that `strmem=` line, as all that's doing is converting the `bytes` into a string which is the opposite of what we want.

You might need to do this:

```
MemoryMapBytes(bytes(mem))
```

to pass the assert, but hopefully that's it.

kk7ds:
Nice, thanks! That first test failure was an azure problem, not yours.

Give me just a sec and I will rebase this for you and add the alias for your variant.

kk7ds:
Okay, the next build after this is merged will update the Supported Radios list on the website to say "Just use the UV9DPlus".

PR #381, opened 2023-01-12
Test refactor and a bug fix

kk7ds:
- Fix radio identities on some test images
- Remove old test runners
- Massive driver test refactor
- Mock out time functions in clone tests
- Fix delete-and-shift with special channels


kk7ds:
The PR checks check is complaining because it detects the tests are adding new `MemoryMap` code, but it's not, it's just moving it, and it's because some drivers still require it.

PR #384, opened 2023-01-13
Add Alinco DJ-G7T

kk7ds:
Fixes #6801


kk7ds:
@mweyland can you confirm this still works for you?

@sjlongland can you confirm this works for you? If they're really this identical we can forego the test image I guess.

sjlongland:
@kk7ds Looking pretty good so far…

image

sjlongland:
dj-g7t-chirp

DJ-G7T works.  I used `chirpc` via a Bash script to extract the data out of my CSV file and inject it into the memory map, then loaded that in using the GUI tool.

One gotcha is the `chirpc` tool seems to take frequencies in Hz whereas the UI and CSV file are in MHz, but that's a side-issue to supporting the DJ-G7T.

kk7ds:
Okay cool, I'm going to say this is good then because it's pretty simple and works for you. If it broke @mweyland I'll accept the blame and fix it.

mweyland:
I can confirm that current master works with the G7EG. Thanks!

kk7ds:
Sweet, thanks @mweyland

PR #385, opened 2023-01-13
New dirver UV68 and TDH8

Sandmann34:


chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2479 @@
...
+FREQHOP_VALUES = ["Off", "On"]
+SCAN_VALUES = ["Del", "Add"]
+
+
+H8_DTCS = sorted(chirp_common.DTCS_CODES + [645])
+
+
+# AB CHANNEL
+A_OFFSET = ["Off", "-", "+"]
+A_QTDQT_DTCS = ['Off', '67.0', '69.3', '71.9', '74.4', '77.0',
```
kk7ds, line 352: This appears to only be used in commented-out code, so it should be removed

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2479 @@
...
+
+    if len(response) in [8, 12]:
+        # DEBUG
+        LOG.info("Valid response, got this:")
+        LOG.info(util.hexprint(response))
+
+        # Determine the walkie-talkie mode
+        # TDH8 have three mode:ham, gmrs and normal
+        global MODE_RADIO
+        MODE_RADIO = util.hexprint(response)[31:37]
```
kk7ds, line 612: As I mentioned before, you may not use a global for this. Also using the output of hexprint is not appropriate for inspection, it's for debugging.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2479 @@
...
+
+def _write_block(radio, addr, data):
+    serial = radio.pipe
+    cmd = struct.pack(">cHb", b'W', addr, 0x20)
+    data = radio.get_mmap()[addr + 8: addr + 40]
+    cmd_str = hex(sum(data) & 0xff)
+    data_bused = chr(int(cmd_str, 16))
+
+    # This is because when our cmd_str is converted to decimal,
+    # sometimes num will be greater than 128, but chr(num).encode('utf-8') gets
```
kk7ds, line 746: This is why you should not be encoding strings to `bytes` at all. You want something from `struct.pack()` if you're building something up, or use `bytes()` itself.  What you have on L742 and L743 is not the proper way to do this, as `int(hex(value), 16)` is an identity operation - it just wastes cpu. This is all you need for the checksum stuff above:
```
data_bused = bytes([sum(data) & 0xFF])
```

But also, "data_bused" does not mean anything to me. I think you want "checksum" ?

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2479 @@
...
+    # part belongs to Extended ASCII, and the characters in the extended
+    # ascii part greater than 128 are encoded by two bytes, and two
+    # characters will be obtained Bytes, what is needed is an
+    # encoding that supports Extended ASCII characters and uses 1 Byte
+    # fixed-size encoding, such as ISO 8859-1, also known as latin1:
+    # >>> chr(165)
+    # ¥
+    # >>> chr(165).encode('latin1')
+
+    data += data_bused.encode('latin1')
```
kk7ds, line 758: You should pretty much never be doing this kind of encode operation.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2479 @@
...
+LOG = logging.getLogger(__name__)
+
+
+MEM_FORMAT = """
+#seekto 0x0008;
+struct {
+  lbcd rxfreq[4];
+  lbcd txfreq[4];
+  lbcd rxtone[2];
+  lbcd txtone[2];
```
kk7ds, line 38: Thank you :)

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2479 @@
...
+def _firmware_version_from_data(data, version_start, version_stop):
+    version_tag = data[version_start:version_stop]
+    return version_tag
+
+
+def _firmware_version_from_image(radio):
+    version = _firmware_version_from_data(radio.get_mmap(),
+                                          radio._fw_ver_file_start,
+                                          radio._fw_ver_file_stop)
+    # LOG.debug("_firmware_version_from_image: " + util.hexprint(version))
```
kk7ds, line 577: I mentioned this before, but there is a lot of commented-out code in this file. Please remove it all, it makes it very hard to review.

kk7ds (changes requested):
Why do you keep closing the PR and creating a new one? It's better if you can keep using the same one because it keeps the conversation all in one place and it makes it easier to see when things are updated.

There are still several things in this version that I've asked to be fixed, so I have only skimmed over this briefly.

Thanks for making progress on this.

chirp/drivers/tdh8.py
Sandmann34: Yes, a checksum is required here, the driver needs a checksum to verify. I will add a comment

chirp/drivers/tdh8.py
Sandmann34: I deleted the output using hexprint, so I processed the data in L695 do_download() to get the walkie-talkie ID, but it is still a global variable, because I need to check the walkie-talkie ID in L922 sync_in(), but I don't want to re-execute L695 do_download( ) This will re-execute the download, so I use a global variable, how should I deal with this walkie-talkie identification variable?

kk7ds: Set a flag on the `radio` object. The `radio` is created per-use and lives for the lifecycle of one conversation/edit session with a given model. So make it so that once the radio talks to a specific model a flag gets set and then apply the rules for that model when editing, and again when communicating. You also need to be able to detect that flag when you load an image so that you know which model a given image file is for, so you can agree or refuse to talk to that model if the user opens an image from disk.

chirp/drivers/tdh8.py
Sandmann34: If I use encode() directly, it will report an error.

chirp/drivers/tdh8.py
Sandmann34: In the latest version, I don't use the global.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+    raise errors.RadioError("Radio did not respond")
+
+
+def _do_download(radio):
+    data = _ident_radio(radio)
+    append_model = False
+    # HAM OR GMRS
+    # Determine the walkie-talkie mode
+    # TDH8 have three mode:ham, gmrs and normal
+    radio.radio_ident = str(data)[2:8]
```
kk7ds, line 597: The return from `_ident_radio()` is a bytes, you should not be `str()`ing it. Is it supposed to be ASCII characters?

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+    LOG.info("Radio mode is " + radio.radio_ident)
+    LOG.info("Chirp choose mode is " + radio.ident_Mode)
+    # The Ham and Gmrs modes are subclasses of this model tdh8.
+    # When the Ham or Gmrs mode is selected,
+    # a double verification operation will be performed to
+    # obtain the current mode code of the walkie-talkie and
+    # the mode code currently selected by the chirp for comparison.
+    # If the two are consistent, enter This mode,
+    # otherwise an inconsistent error will be prompted.
+    if radio.radio_ident == radio.ident_Mode:
```
kk7ds, line 607: If the result of `_ident_radio()` is the same as `ident_Mode` then you don't need to store it on radio as well, right? Just compare it here:
```
if data != radio.ident_mode:
    raise ...
```
and then you can assume it matched every time after this right? When I told you to store it on the radio, I didn't realize it was identical to the thing you already had there (because of the `hexprint` obfuscation).

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+        LOG.info("Successful match.")
+    else:
+        msg = (
+            "Please check whether the walkie-talkie mode is consistent with "
+            "the selected model mode.")
```
kk7ds, line 612: Our more conventional error message for this would be something like *Unable to talk to this model* or *Model mismatch*. Both "walkie talkie" and "model mode" do not sound very good to native english speakers.

chirp/drivers/tdh8.py
kk7ds: Right, because it's not valid `UTF-8`. It's not valid `latin1` either, it's just that `latin1` is a direct 1:1 byte-for-byte conversion. You should be using `int` and `bytes` types for binary data, not `str`. I gave you the solution above.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+TDH8_CHARSET = chirp_common.CHARSET_UPPER_NUMERIC + \
+    "!@#$%^&*()+-=[]:\";'<>?,./"
+
+
+@directory.register
+class TDH8(chirp_common.CloneModeRadio):
+    """TIDRADIO TD-H8"""
+    VENDOR = "TIDRADIO"
+    MODEL = "TD-H8"
+    ident_Mode = "P31183"
```
kk7ds, line 699: The "M" should not be capitalized, and the convention for model-specific stuff like this is to prefix it with an underscore, as you do elsewhere.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+    _gmrs = False
+    _ham = False
+    _mem_params = (0x1F2F)  # poweron_msg offset
+
+    # offset of fw version in image file
+    _fw_ver_file_start = 0x1838
+    _fw_ver_file_stop = 0x1846
+    NEEDS_COMPAT_SERIAL = False
+    _valid_chars = TDH8_CHARSET
+    radio_ident = ''
```
kk7ds, line 719: Based on the above, I suggest you remove this.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+        return msgs
+
+    def sync_in(self):
+        try:
+            self._mmap = _do_download(self)
+        except Exception as e:
+            raise errors.RadioError("Failed to communicate with radio: %s" % e)
+
+        self.process_mmap()
+        if self.radio_ident == "P31183":
```
kk7ds, line 818: Use your static `ident_mode` here. It will always match what the user selected, and is confirmed by the clone routines, if you do what I suggested above.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+            self._memobj.settings.ham = 0
+            self._memobj.settings.gmrs = 0
+        elif self.radio_ident == "P31184":
+            self._memobj.settings.ham = 0
+            self._memobj.settings.gmrs = 1
+        elif self.radio_ident == "P31185":
+            self._memobj.settings.ham = 1
+            self._memobj.settings.gmrs = 0
```
kk7ds, line 826: In all three of these cases, why are you modifying the image as soon as you download it? These values should already be set in the radio's memory if you've confirmed that the user selected the right model, correct? It is very unconventional for chirp to modify the image after download, as saving the result of a radio download should exactly match what is in the radio.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+
+        # Channel name
+        for char in _nam.name:
+            if str(char) == "\xFF":
+                char = " "
+            mem.name += str(char)
+        mem.name = mem.name.rstrip()
+        if (self.radio_ident == "P31185" or self.radio_ident == "P31184") and \
+                (mem.number >= 189 and mem.number <= 199):
+            mem.name = HAM_NAME[mem.number - 200]
```
kk7ds, line 932: It would really be better to delegate this sort of thing to a handler in the subclasses. Something like:
```
mem.name = self._get_model_name(mem.number, _nam)

def _get_model_name(self, number, _nam):
    return str(_nam.name).replace('\xFF', '').rstrip()
```
and then define this in the ham-specific class:
```
def _get_model_name(self, number, _nam):
    if 189 <= number <= 199:
        mem.name = HAM_NAME[number - 200]
    else:
        mem.name = super()._get_model_name(number, _nam)
```

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+        _nam = self._get_nam(mem.number)
+
+        if mem.empty:
+            _mem.set_raw("\xff" * 16)
+            _nam.set_raw("\xff" * 16)
+            return
+
+        _mem.set_raw("\x00" * 16)
+
+        if self._ham and self.radio_ident == "P31185":
```
kk7ds, line 1058: Why and? When would they ever be different? Anyway, as above, this would be better delegated to a helper method in the ham subclass.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+                _mem.offset = 0
+                msg = (
+                    "The frequency in channels 31-55 must be between "
+                    "462.55000-462.72500 in 0.025 increments.")
+                raise InvalidValueError(msg)
+
+        else:
+            _mem.rxfreq = _mem.txfreq = mem.freq / 10
+
+        # When the channel is empty, you need to set "blockfirm" to 0,
```
kk7ds, line 1131: Again, please choose a different name than "blockfirm".. That name doesn't mean anything to native english speakers. If you give me some more explanation of what you think it means, I may be able to offer some better alternatives.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2222 @@
...
+    except:
+        raise errors.RadioError("Radio refused to exit programming mode")
+
+
+def _write_block(radio, addr, data):
+    serial = radio.pipe
+    cmd = struct.pack(">cHb", b'W', addr, 0x20)
+    data = radio.get_mmap()[addr + 8: addr + 40]
+    cmd_str = hex(sum(data) & 0xff)
+    data_bused = chr(int(cmd_str, 16))
```
kk7ds, line 646: Again, it's not okay to convert ints to hex strings, re-parse as hex ints and then stringify the result. Do the math in integers only, please.

kk7ds (changes requested):
Still improving which is good, but there are still many things that need to be fixed that I have commented on in the past. I did another skim through here, but there is a lot to review, so I'm just doing small chunks at a time.

Now that I understand what your ident string is, I've suggested to just *check* it against `radio.ident_mode` instead of storing it and then checking -- storing it (in duplicate) doesn't accomplish anything if you've already got it stored on the class to begin with.

Thanks!

chirp/drivers/tdh8.py
Sandmann34: So I need to delete the `ident_Mode` stored in the `radio`, right?

kk7ds: No, read what I said above. Keep the static definition of `ident_Mode` and just *check* against it. What you need to do is very common in many other drivers. They store an ident string on the class, and just *check* that it's the expected value before upload/download. Examples:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/icf.py#L339-L351

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/kenwood_live.py#L192-L194

I suggested storing the ident on the radio at clone time (instead of the global) because it wasn't clear to me how you were using it, but now that I see you're just comparing it to the static `ident_Mode` I see there's no reason to store it *again*. Just compare it.

kk7ds: Here's another example:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/tk8180.py#L246

Sandmann34: This is what I think, when I get the ident string of the radio and compare it with the ident_Mode in the static, to determine what mode it is.

Sandmann34: For example, if the user selects TDH8 instead of TDH8-GMRS model, and the radio ident string is "P31184", choosing TDH8 will use the static ident_Mode as "P31183" to compare them. If they are different, an error will be prompted.

chirp/drivers/tdh8.py
Sandmann34: `blockfirm `is to confirm whether the channel is empty, if it is empty, the data will be set to 0, so I choose to modify it to `blockconfirm`.

chirp/drivers/tdh8.py
Sandmann34: Thank you for the method you provided, because he uses this HAM_NAME in both HAM and GMRS, so I should modify the name of this HAM_NAME to avoid misleading, so I should write it like this:

```
mem.name = mem.name.rstrip()
if self.ident_mode != b'P31183\xff\xff' and \
        (mem.number >= 189 and mem.number <= 199):
    mem.name = HAM_NAME[mem.number - 200]
```

chirp/drivers/tdh8.py
Sandmann34: Now I changed to bytes type: `b'P31183\xff\xff'`

Sandmann34:
BTW,My company want to inquire about cooperation in advertising this time. Can you give me your contact information? My colleague will contact you for cooperation.

chirp/drivers/tdh8.py
kk7ds: Okay, then I would recommend "empty flags" or "used flags" or just "used". "blockconfirm" doesn't really mean the right thing either (but it's better than "blockfirm")

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+    @classmethod
+    def get_prompts(cls):
+        rp = chirp_common.RadioPrompts()
+        rp.experimental = \
+            ('Due to the fact that the manufacturer continues to '
+             'release new versions of the firmware with obscure and '
+             'hard-to-track changes, this driver may not work with '
+             'your device. Thus far and to the best knowledge of the '
+             'author, no TD-H8 radios have been harmed by using CHIRP. '
+             'However, proceed at your own risk!')
```
kk7ds, line 709: Do you still want this to be displayed to the user? Aren't you associated with the manufacturer? :)

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+                        # warn user wrong Duplex
+                        msgs.append(chirp_common.ValidationError(_msg_duplex3))
+
+        return msgs
+
+    def sync_in(self):
+        try:
+            self._mmap = _do_download(self)
+            self.process_mmap()
+
```
kk7ds, line 794: Remove blank line

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+        except Exception as e:
+            raise errors.RadioError("Failed to communicate with radio: %s" % e)
```
kk7ds, line 796: If you would, pass through the `RadioError` exceptions. This makes it a lot easier to debug after the fact, and is why I want you to raise `RadioError` above. So make this (and the upload one) something like:

```
try:
    self_mmap = _do_download(self)
    self.process_mmap()
except RadioError:
    raise  # This causes it to raise straight to the UI instead of being translated
except Exception as e:
    # This is only for unexpected errors
    raise errors.RadioError("Failed to communicate with this radio: %s" % e)
```

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+        dtmf.append(rs)
+
+        return group
+
+    def get_settings(self):
+        try:
+            return self._get_settings()
+        except:
+            import traceback
+            LOG.error("Failed to parse settings: %s", traceback.format_exc())
```
kk7ds, line 1717: I think I mentioned this before, but please do not import and format your own traceback use `LOG.exception` which will include the stack trace automatically.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+            return None
+
+    def set_settings(self, settings):
+
+        def fm_validate(value):
+            if 760 > value or value > 1080:
+                msg = ("FM Channel muse be between 76.0-108.0")
+                raise InvalidValueError(msg)
```
kk7ds, line 1725: *This* is a proper use of `InvalidValueError` :)

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+                            else:
+                                list_val.append(0xFF)
+                                lenth_val += 1
+                        self._memobj.group8.group8 = list_val
+                    elif element.value.get_mutable():
+                        setattr(obj, setting, element.value)
+                except Exception as e:
+                    LOG.debug(element.get_name())
+                    raise
+
```
kk7ds, line 2156: This whole `set_settings()` method is reeeeally long. It might be nice to break it up in the future, like you have already done in some cases, but I won't make you do it now.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+            rf.valid_bands = [normal_bands[0], normal_bands[1]]
+        else:
+            rf.valid_bands = normal_bands
+        rf.memory_bounds = (1, 199)
+        return rf
+
+    def process_mmap(self):
+        self._memobj = bitwise.parse(MEM_FORMAT, self._mmap)
+
+    def validate_memory(self, mem):
```
kk7ds, line 765: This only needs to be on the gmrs variant, so please move this to just be in that subclass so it is only run there. Then you can remove the `if self._gmrs` check on L771. Yes, I know this is probably copied from other drivers, but it's better to put it on the subclass.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+        msgs = chirp_common.CloneModeRadio.validate_memory(self, mem)
```
kk7ds, line 766: This still works, but the python3 way to do this is:
```
msgs = super().validate_memory(mem)
```

So when you move it, please make that change.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2193 @@
...
+
+
+@directory.register
+class TDH8_GMRS(TDH8):
+    VENDOR = "TIDRADIO"
+    MODEL = "TD-H8-GMRS"
+    ident_mode = b'P31184\xff\xff'
+    _gmrs = True
+    _vhf_range = (136000000, 175000000)
+    _uhf_range = (400000000, 521000000)
```
kk7ds, line 2187: Put your `validate_memory()` routine here and then it will only run for this model.

kk7ds (commented):
Thanks @Sandmann34, this is getting really close. Just a few other comments from me (several of which I've mentioned before). Please do go through and clean up the commented-out code. There's still a lot of it. If those are things you might need to use in the future, just say so and we should put them behind a guard so that they can be enabled easily.

There are a few other small things that `flake8` finds with this driver, so please fix these as well:
```
chirp/drivers/tdh8.py:555:5: E722 do not use bare 'except'
chirp/drivers/tdh8.py:631:5: E722 do not use bare 'except'
chirp/drivers/tdh8.py:651:5: F841 local variable 'ident' is assigned to but never used
chirp/drivers/tdh8.py:1239:9: E722 do not use bare 'except'
chirp/drivers/tdh8.py:1715:9: E722 do not use bare 'except'
chirp/drivers/tdh8.py:2153:17: F841 local variable 'e' is assigned to but never used
chirp/drivers/tdh8.py:2167:13: F841 local variable 'e' is assigned to but never used
```
for `E722` use `except Exception` instead of `except`.

If you can get the last few things cleaned up, we can merge this soon.

Thanks for your patience and work on this!

kk7ds:
Oh also, let me know if you need me to rebase this for you, which I can do easily, once the last few things are done. Also you need to add yourself to `tests/py3_driver_testers.txt` for all three radios and then generate the matrix:
```
tox -e makesupported
```

Sandmann34:
I added the three radios to `py3_driver_testers.txt` and ran `tox -e makesupported`，there seems to be a little problem：

> UnicodeEncodeError: 'gbk' codec can't encode character '\u20ac' in position 1618: illegal multibyte sequence

And thank you for being so patient in guiding me. 👍

kk7ds (changes requested):
I was going to merge this, but then I found the `mem.skip` thing, sorry for missing it before. I was going to rebase this for you and fix your missing `LOG.exception`, but I think the skip thing is too much for me to fix without testing. Other than that, I was ready to go.

Can you fix the skip thing?

tests/py3_driver_testers.txt
```diff
@@ -257,6 +257,11 @@ Retevis_RT9000D_400-490,@KC9HI,8-Dec-2022
 Retevis_RT9000D_66-88,@KC9HI,8-Dec-2022
 TDXone_TD-Q8A,@KC9HI,18-Dec-2022
 TIDRADIO_TD-H6,@kk7ds,21-Oct-2022
+TIDRADIO_TD-H8,@Sandmann34,7-Feb-2023
+TIDRADIO_TD-H8-GMRS,@Sandmann34,7-Feb-2023
```
kk7ds, line 261: You should also include a test image for this one because of the special memory behaviors, and probably the ham variant as well. Since it looks like the UV68 is identical to the TD-H8, no need to  have one for that.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2176 @@
...
+            for i in range(0, 8):
+                scan_val = (a._generators["scan" + str(i + 1)])
+                # print(str(scan_val)[3])
+                used_scan_val = str(scan_val)[3]
+                scan_val_list.append(used_scan_val)
+
+        rs = RadioSetting(
+            "scanadd", "Scan Add", RadioSettingValueList(
+                SCAN_VALUES, SCAN_VALUES[int(scan_val_list[number - 1])]))
+        mem.extra.append(rs)
```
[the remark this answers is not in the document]
Sandmann34: Yes, I didn’t understand that `scanadd` is the same as `skip`, now that I know it, I don’t need to put it in `extra`.

tests/py3_driver_testers.txt
Sandmann34: Do you mean that I need to reposition the image of tdh8 and add the test image of ham and gmrs? Because I placed one in images before.

chirp/drivers/tdh8.py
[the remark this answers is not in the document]
kk7ds: Cool, make sure you add the `valid_skips` too, so that the UI knows to show that column. Thanks!

tests/py3_driver_testers.txt
kk7ds: Just add the ham and gmrs images to what you have here. The PR here only adds one image (see "Files changed" tab), perhaps you had others in a previous rev? But yeah, just add those two in addition to what you have here.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2153 @@
...
+                    mem.freq >= 400000000 and mem.freq <= 520000000)):
+                _mem.rxfreq = mem.freq / 10
+
+            else:
+                _mem.rxfreq = mem.freq / 10
+                # msg = (
+                #     "The frequency in channels 31-55 must be between "
+                #     "462.55000-462.72500 in 0.025 increments.")
+                # raise errors.RadioError(msg)
+
```
Sandmann34, line 1069: I have some questions and don’t know how to solve them. In gmrs mode, channel 31-55 is a fixed optional channel, fixed at 462.55000-462.725 with an increment of 0.025, which is:
```
462.550
462.575
432.600
462.650
462.675
462.700
462.725
```
If I use error to prompt the user here, will it pass the tox-driver test.
what should i do here？

Sandmann34:
And can I ask you something? Do you know to use pyinstaller to package into an executable file, does it need to manually integrate all resource files to package? I would be grateful if you could give me some assistance. 🙏

chirp/drivers/tdh8.py
```diff
@@ -1148,46 +1159,41 @@ def set_memory(self, mem):
...
+            _scan.scan4 = mem.skip != "S"
+        elif scanlist[1] == 5:
+            _scan.scan5 = mem.skip != "S"
+        elif scanlist[1] == 6:
+            _scan.scan6 = mem.skip != "S"
+        elif scanlist[1] == 7:
+            _scan.scan7 = mem.skip != "S"
+        elif scanlist[1] == 8:
+            _scan.scan8 = mem.skip != "S"
+
```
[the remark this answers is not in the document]
Sandmann34: I don't know how to use bitmask to deploy scan，use bool but their types are different.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2151 @@
...
+                    mem.mode = "NFM"
+                    mem.duplex = ""
+                else:
+                    gmrs_freqs = int(GMRS_FREQS_B[mem.number - 1] * 1000000)
+                    mem.freq = gmrs_freqs
+                    mem.duplex = ""
+                    _mem.rxfreq = mem.freq / 10
+
+            elif float("%i.%06i" % (mem.freq / 1000000, mem.freq % 1000000)
+                       ) in GMRS_FREQ_A:
```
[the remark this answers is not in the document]
Sandmann34: Thank you so much 👍, now I know how to use `immutable` and `validate_memory()`.

Sandmann34:
Can you help me join Test / Create Support Matrix?

Thanks for your guidance.

chirp/drivers/tdh8.py
```diff
@@ -440,7 +440,8 @@
 GMRS_FREQ_A = GMRS_FREQS1 + GMRS_FREQS3
 
 HAM_GMRS_FREQS = [162.55000, 162.40000, 162.47500, 162.42500, 162.45000,
-                  162.50000, 162.52500, 161.65000, 161.77500, 161.75000]
+                  162.50000, 162.52500, 161.65000, 161.77500, 161.75000,
+                  162.00000]
```
kk7ds, line 444: Can you please use the values already in `bandplan_na`?

chirp/drivers/tdh8.py
```diff
@@ -906,7 +893,47 @@ def get_memory(self, number):
         elif int(scan_val_list[number - 1]) == 1:
             mem.skip = ''
 
-        mem.mode = _mem.wide and "NFM" or "FM"
+        if self._is_txinh(_mem):
+                mem.duplex = ""
```
kk7ds, line 897: CHIRP uses `duplex="off"` to mean "TX inhibit. This `duplex=""` means "same TX and RX frequency.

chirp/drivers/tdh8.py
```diff
@@ -906,7 +893,47 @@ def get_memory(self, number):
...
+            mem.duplex = int(_mem.rxfreq) > int(_mem.txfreq) and "-" or "+"
+            mem.offset = abs(int(_mem.rxfreq) - int(_mem.txfreq)) * 10
+
+        # Offset
+        if self._gmrs:
+            mem.duplex = ""
+            mem.offset = 0
+            mem.immutable = ['duplex']
+            if mem.number >= 1 and mem.number <= 30:
+                mem.immutable = ['freq']
```
kk7ds, line 915: This will overwrite L913. You probably want `mem.immutable.append('freq')` right?

chirp/drivers/tdh8.py
```diff
@@ -906,7 +893,47 @@ def get_memory(self, number):
...
+                if mem.number >= 8 and mem.number <= 14:
+                    mem.mode = 'NFM'
+                    mem.power = TX_POWER[0]
+                    mem.immutable = ['freq', 'mode', 'power', 'duplex',
+                                     'offset']
+            elif mem.number >= 31 and mem.number <= 53:
+                self.validate_memory(mem)
```
kk7ds, line 922: This doesn't make any sense. CHIRP calls `validate_memory()` before `set_memory()` to see if it is suitable. It doesn't make any sense to call it in `get_memory()` because *you* are returning the memory, so you control the rules.

chirp/drivers/tdh8.py
```diff
@@ -987,80 +1014,8 @@ def set_memory(self, mem):
...
-                mem.immutable = ['freq', 'duplex', 'offset', 'power']
-                msg = (
-                    "The frequency in channels 31-55 must be between "
-                    "462.55000-462.72500 in 0.025 increments.")
-                raise errors.RadioError(msg)
-
+        if self._gmrs and (mem.number >= 55 and mem.number <= 199):
+            _mem.rxfreq = mem.freq / 10
         else:
             _mem.rxfreq = _mem.txfreq = mem.freq / 10
```
kk7ds, line 1020: This line is dead code. `_mem.txfreq` will always get set below. It doesn't technically hurt anything, it just doesn't have any purpose.

chirp/drivers/tdh8.py
```diff
@@ -2144,6 +2091,15 @@ class TDH8_GMRS(TDH8):
     _vhf_range = (136000000, 175000000)
     _uhf_range = (400000000, 521000000)
 
+    def validate_memory(self, mem):
+        msgs = super().validate_memory(mem)
+        if mem.freq / 1000000 in GMRS_FREQS3:
```
kk7ds, line 2096: If you use the GMRS definitions in `bandplan_na` you won't have to divide here.

chirp/drivers/tdh8.py
```diff
@@ -2144,6 +2091,15 @@ class TDH8_GMRS(TDH8):
...
+            pass
+        else:
+            msgs = ("The frequency in channels 31-55 must be between "
+                    "462.55000-462.72500 in 0.025 increments.")
```
kk7ds, line 2100: `msgs` needs to be a list of objects, not a single string. This will break the UI. You also need to append, in case the base `validate_memory()` returned any messages.

You're requiring all memories to be in that range now, because you're not checking `mem.number`. You need to do something like this:
```
if 31 <= mem.number <= 55: and mem.freq not in GMRS_FREQS:
        msgs.append(chirp_common.ValidationError("The frequency in channels ...")
```
...right?

kk7ds:
> Can you help me join Test / Create Support Matrix?

I will fix it for you once we're set otherwise. You have a bunch of commits here and some merges, and those all need to be squashed out before merge. So I'll just do that for you and fix your support matrix at the end.

chirp/drivers/tdh8.py
Sandmann34: Here you may need to define it yourself, because I can't find these values ​​in bandplan_na.

chirp/drivers/tdh8.py
Sandmann34: Thanks, these values I can find in `bandplan_na`.

chirp/drivers/tdh8.py
kk7ds: Oh, I didn't even look at the frequencies. These are NOAA weather channels, not `HAM_GMRS_FREQS` so the variable is named incorrectly. But, you should store them in Hz, like in `bandplan_na`. Really we should add these NOAA channels to `bandplan_na` but I won't make you do that. Just store them as `162550000` and not `162.55`. Floating point numbers do not have infinite resolution and are not exact.

Sandmann34:
I have repaired everything you said. I am ready to merge. Thanks.

chirp/drivers/tdh8.py
```diff
@@ -1066,18 +1043,26 @@ def set_memory(self, mem):
...
-            _mem.txfreq = mem.offset / 10
-        elif mem.duplex == "+":
-            _mem.txfreq = (mem.freq + mem.offset) / 10
-        elif mem.duplex == "-":
-            _mem.txfreq = (mem.freq - mem.offset) / 10
+        if self._gmrs:
+            if 55 <= mem.number <= 199:
+                _mem.rxfreq = mem.freq / 10
+            elif 31 <= mem.number <= 55:
+                self.validate_memory(mem)
```
kk7ds, line 1050: Again, `validate_memory()` is called before `set_memory()` for you. You do not need to (nor should you) call it from inside `set_memory()`.

chirp/drivers/tdh8.py
```diff
@@ -1066,18 +1043,26 @@ def set_memory(self, mem):
...
-            _mem.txfreq = mem.freq / 10
-        elif mem.duplex == "split":
-            _mem.txfreq = mem.offset / 10
-        elif mem.duplex == "+":
-            _mem.txfreq = (mem.freq + mem.offset) / 10
-        elif mem.duplex == "-":
-            _mem.txfreq = (mem.freq - mem.offset) / 10
+        if self._gmrs:
+            if 55 <= mem.number <= 199:
+                _mem.rxfreq = mem.freq / 10
```
kk7ds, line 1048: This will always be overridden by L1063. There is no reason to have this here.

chirp/drivers/tdh8.py
```diff
@@ -1066,18 +1043,26 @@ def set_memory(self, mem):
...
+            elif 31 <= mem.number <= 55:
+                self.validate_memory(mem)
         else:
-            _mem.txfreq = mem.freq / 10
+            if mem.duplex == "":
+                _mem.rxfreq = _mem.txfreq = mem.freq / 10
```
kk7ds, line 1053: same for the `_mem.rxfreq` part of this. It will always be overridden by L1063.

chirp/drivers/tdh8.py
```diff
@@ -1066,18 +1043,26 @@ def set_memory(self, mem):
...
-        elif mem.duplex == "+":
-            _mem.txfreq = (mem.freq + mem.offset) / 10
-        elif mem.duplex == "-":
-            _mem.txfreq = (mem.freq - mem.offset) / 10
+        if self._gmrs:
+            if 55 <= mem.number <= 199:
+                _mem.rxfreq = mem.freq / 10
+            elif 31 <= mem.number <= 55:
+                self.validate_memory(mem)
         else:
```
kk7ds, line 1051: When is the `_mem.txfreq` ever set for the gmrs case?

chirp/drivers/tdh8.py
```diff
@@ -2093,11 +2078,12 @@ class TDH8_GMRS(TDH8):
 
     def validate_memory(self, mem):
         msgs = super().validate_memory(mem)
-        if mem.freq / 1000000 in GMRS_FREQS3:
-            pass
+        if mem.freq not in GMRS_FREQS:
+            msgs.append(chirp_common.ValidationError(
+                "The frequency in channels 31-55 must be between"
```
kk7ds, line 2083: This error will always be given for any non-GMRS frequency, even for channels 1-30 and >55 because `validate_memory()` is called before `set_memory()` by the GUI. Are you actually testing this code locally with the GUI?

chirp/drivers/tdh8.py
```diff
@@ -2093,11 +2078,12 @@ class TDH8_GMRS(TDH8):
...
+                "462.55000-462.72500 in 0.025 increments."))
         else:
-            msgs = ("The frequency in channels 31-55 must be between "
-                    "462.55000-462.72500 in 0.025 increments.")
+            pass
```
kk7ds, line 2086: Just remove the entire `else` part. `else..pass` is just dead code.

kk7ds (commented):
This is very close, but several things here still make no sense. I definitely want to get this merged, but it seems like you are missing several things and some obviously wrong things that should be obvious from testing yourself.

kk7ds:
Doing some tests with this, I found some more issues.

Loaded the TD-H8 image and I set memory 8 to 446.0 -5.0MHz. It got a skip=S immediately, and I am unable to unselect skip.

With TD-H8 GMRS, I am unable to enable duplex=+ for the high GMRS channels. If I delete channel 19 I can edit and change it to 442.0, which it accepts, but keeps the frequency and duplex immutable. If I do the same in channel 12, it accepts the non-GMRS channel, but the mode is forced to NFM and power to low. Editing channel 34 doesn't let me set 442.0 (as expected), but if I enter 462.7 I can't set duplex or offset (462.7 is allowed +5MHz duplex on GMRS).

If I edit channel 102 to 442, duplex is immutable, but I assume it shouldn't be, since you allow transmit on 442.

Sandmann34:
I just fixed the problem that you said that you can't choose to cancel skip in TD-H8 image. I didn't find it before. Now I fix it.
Then you said that duplex cannot be used in gmrs. I found an example of driver **UV5R** here to imitate the restriction method of gmrs, so I will make duplex immutable.

https://github.com/kk7ds/chirp/blob/79e500b7f12b9108135e07340d96a7c6a98d9e18/chirp/drivers/uv5r.py#L986-L1012

In tdh8, 1-30 channels are fixed, and the reception frequency and transmission frequency are not allowed to be modified. The transmit power in 8-14 channels is low power by default, and the wide and narrow bands are narrow bands, which cannot be modified.

As for what you said (462.7 is allowed+5MHz duplex on GMRS):
I didn't consider this before. Do I need to change it to this, or do I immutable all duplex as **UV5R**?

kk7ds:
> In tdh8, 1-30 channels are fixed, and the reception frequency and transmission frequency are not allowed to be modified. The transmit power in 8-14 channels is low power by default, and the wide and narrow bands are narrow bands, which cannot be modified.

Well, then you need to make your driver enforce this (use `validate_memory()`) because I was able to store ham frequencies in those channels.

> As for what you said (462.7 is allowed+5MHz duplex on GMRS): I didn't consider this before. Do I need to change it to this, or do immutable all duplex like **UV5R**?

If the radio supports +5MHz duplex on GMRS repeater channels (which it should, if it's going to be useful for GMRS people), you need to allow changing duplex to + on those channels only, and only to 5MHz. That likely means that the GMRS version needs to export different `valid_duplexes`.

@KC9HI Could you maybe suggest a radio that @Sandmann34 should look at that has similar GMRS/immutable restrictions to what he describes here? I've lost track of which ones are very restrictive and not, but I think this description is one of the simpler ones with very specific rules. It's tempting to look at the UV-5R but that driver is very large and supports a lot of models, so it might be better to look at something simpler.

KC9HI:
@kk7ds, I would recommend that @Sandmann34 take a look at the Retevis RB17P radio model (retevis_rb170.py). It is a GMRS radio model with 128 channels. The GMRS channels are locked to memories 1-30. The remaining channels are RX only over the 400-470 MHz frequency range.

Sandmann34:
Now gmrs allows duplex, 31-54 channels can set duplex to +, and offset to 5MHz, just like Retevis RB17P.
And fixed the problem that the skip cannot be modified.
thanks.

Sandmann34:
I refer to the Retevis RB17P radio model, and now it's fixed. I'm ready to merge.

Sandmann34:
This does not seem to be a style issue for this model:
```
style: commands[0]> python ./tools/cpep8.py
./chirp/drivers/rfinder.py:1:1: E902 FileNotFoundError: [Errno 2] No such file or directory: './chirp/drivers/rfinder.py'
```

kk7ds:
> This does not seem to be a style issue for this model:
> 
> ```
> style: commands[0]> python ./tools/cpep8.py
> ./chirp/drivers/rfinder.py:1:1: E902 FileNotFoundError: [Errno 2] No such file or directory: './chirp/drivers/rfinder.py'
> ```

This is because you're doing merge commits and haven't properly resolved the changes to that file. It's still a problem that has to be resolved...

kk7ds:
I squashed your commits, updated the support matrix and fixed your conflict. If you need to make changes from here, you need to pull this locally otherwise you'll overwrite if you push again.

chirp/drivers/tdh8.py
```diff
@@ -0,0 +1,2084 @@
...
+    _vhf_range = (136000000, 175000000)
+    _uhf_range = (400000000, 521000000)
+
+    def validate_memory(self, mem):
+        msgs = super().validate_memory(mem)
+        if 31 <= mem.number <= 54 and mem.freq not in GMRS_FREQS:
+            msgs.append(chirp_common.ValidationError(
+                "The frequency in channels 31-54 must be between"
+                "462.55000-462.72500 in 0.025 increments."))
+        return msgs
```
kk7ds, line 2078: One other thing to check in here is that the duplex (and maybe offset) are right. If I choose duplex="-" for one of the GMRS-only channels, it flips back to + (which is good) but I get no warning about it. Just check here and add a warning (not error) so that the user will see.

Also, I can set offset=3MHz on GMRS channels, perhaps you want to enforce that it's always either simplex or +5MHz on those?

tests/Python3_Driver_Testing.md
```diff
@@ -308,7 +308,11 @@
 | <a name="Signus_XTR-5"></a> Signus_XTR-5 |  |  | Yes |  |
 | <a name="TDXone_TD-Q8A"></a> TDXone_TD-Q8A | [@KC9HI](https://github.com/KC9HI) | 18-Dec-2022 |  | 0.02% |
 | <a name="TIDRADIO_TD-H6"></a> TIDRADIO_TD-H6 | [@kk7ds](https://github.com/kk7ds) | 21-Oct-2022 | Yes | 0.43% |
+| <a name="TIDRADIO_TD-H8"></a> TIDRADIO_TD-H8 | [@Sandmann34](https://github.com/Sandmann34) | 7-Feb-2023 | Yes |  |
+| <a name="TIDRADIO_TD-H8-GMRS"></a> TIDRADIO_TD-H8-GMRS | [@Sandmann34](https://github.com/Sandmann34) | 7-Feb-2023 | Yes |  |
+| <a name="TIDRADIO_TD-H8-HAM"></a> TIDRADIO_TD-H8-HAM | [@Sandmann34](https://github.com/Sandmann34) | 7-Feb-2023 | Yes |  |
 | <a name="TID_TD-M8"></a> TID_TD-M8 | [Implied by Retevis_RT22](#user-content-Retevis_RT22) | 9-Dec-2022 | Yes | 0.04% |
+| <a name="TID_TD-UV68"></a> TID_TD-UV68 | [@Sandmann34](https://github.com/Sandmann34) | 7-Feb-2023 | Yes |  |
```
kk7ds, line 315: Can you explain why some radios are *TIDRADIO* and others are *TID*? Are these actually marketed differently or is one the more correct? I assume that since you're adding one of each that there's a reason, but since changing the models later is hard, I want to make sure this is right.

kk7ds (commented):
Okay, I'm ready to merge this, except for one thing. I want to make *sure* that we need both TID and TIDRADIO (and please explain why). If we do, then that's fine and I'll merge this.

The GMRS comments are not critical, so don't worry about them now, we can follow up after we merge this.

kk7ds:
(last push was just updating with bug reference in the commit and stricter style checks)

kk7ds:
Okay, I don't see much about the UV68 out there, but what I did find does show that they say "TID" on the front unlike the H8 (and others) which say "TIDRADIO". Not sure we really need that distinction for these, but since we already have some in the tree, I guess it's no worse.

Thanks!

PR #393, opened 2023-01-14
Make cell background grey for immutable values

kk7ds:
- Refactor repeaterbook item parsing for errors
- Add RepeaterBook "rest of world" queries
- Add support for RepeaterBook Notes field
- Make cell background grey for immutable values


kk7ds:
@KC9HI What do you think?

KC9HI:
Perfect. I can immediately see that I still have some work to do. ;)

kk7ds:
Heh, okay :)

I'm going to sneak one more fix into here that isn't technically a jimfix, and then I will merge.

PR #397, opened 2023-01-15
Bug 10271: drivers.ftm350: Define and set "visible" bit

sjlongland:
This bit seems to control whether the channel is visible to the user through the radio's UI or not.  Without this set to `1`, the user must remember the channel number and use the DTMF keypad to "dial" in that channel instead of simply recalling it using the knob or up/down buttons.

Fixes: [#10271](https://chirp.danplanet.com/issues/10271)


kk7ds:
Good work, thanks!

PR #398, opened 2023-01-15
Fix str / bytes error on python 3

hb9hnt:
As requested here https://chirp.danplanet.com/issues/10275 the fix for the simple python 3 compatibilty errors in the Yaesu FT4 driver.

The syncing issue [mentioned in bug #10198](https://chirp.danplanet.com/issues/10198) still remains, though.


kk7ds:
Thanks! I just added the bug reference in the proper format to the commit message so it will be tied to the issue on the chirp site.

PR #400, opened 2023-01-15
Allow 'empty' in mem.immutable

kk7ds:


kk7ds:
@KC9HI Hmm, test_delete was the only one I saw fail with your patch applied. Could this fix be the only thing needed?

KC9HI:
The output from tox mostly confuses me, but after taking a quick look at it again, I believe that you are correct.

PR #406, opened 2023-01-17
Generic mem.comment support

kk7ds:
- Move mmap and metadata code to CloneModeRadio
- Refactor get, set, erase memory ops in memedit
- Add interface for storing extra memory fields
- Make wxui support ExternalMemoryProperties
- Limit comment input to 256 characters
- Remove FT-450's abuse of the comment field

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -1133,6 +1133,54 @@ class Radio(Alias):
         return fmt in cls.FORMATS
 
 
+class ExternalMemoryProperties:
+    """A mixin class that provides external memory property support.
+
+    This is for use by drivers that have some way of storing additional
+    memory properties externally (i.e. in metadata or a separate region)
+    that cannot be loaded/updated during get_memory()/set_memory().
+    Implementing this is much less ideal than supporting those properties
+    directly, so this should only be used when absolutely necessary.
+    """
+
+    def get_memory_extra(self, memory):
+        """Update @memory with extra fields.
+
+        This is called after get_memory() and is passed the result
+        for augmentation from external storage.
+        """
+        return memory
+
+    def set_memory_extra(self, memory):
+        """Update external storage of properties from @memory.
+
+        This is called after set_memory() with the same memory object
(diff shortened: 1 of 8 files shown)


kk7ds:
cc @KC9HI

KC9HI:
So far so good. I loaded an image and added comments on a few memory rows. I copied the memory rows to a CSV tab and the comments were copied as well. I saved and loaded the CSV file and all was restored. I saved and loaded the image tab and all was restored.

I can try a variety of drivers. Are there anything types of things that you want me to test?

kk7ds:
I just want you to try to break it :)

Trying multiple different drivers would be good. I found the sketchy 450D behavior in the last commit by grepping not specifically testing, but it makes me worried that there could be other such gotchas. But, if you can't break it with moderate testing, that's probably good enough.

KC9HI:
Challenge accepted. I will get back to you.

KC9HI:
Everything was going very well until...

If a radio has left/right memories or upper/lower memories, when I save to an image and then load the image back into CHIRP the comments are gone. Judging from the size of the image file (it appears to be the same size at the original image without comments), I would say the comments are not being saved.

kk7ds:
Ooooh, good catch! Also, that might be really ugly to fix :(

I friggin hate radios with split memories, why oh why would you ever want that?

kk7ds:
Okay, tested (and now working) with an FT-8800 image. What were you testing with?

KC9HI:
BTECH UV-50X3 (Left/Right) and Luiton LT-725UV (Upper/Lower)

kk7ds:
Okay, cleaned that up a little more and I think it's good now. Excellent catch, thanks! Do you think this is good now or did you have any more things you wanted to try? From my perspective it seems "good enough" at least.

KC9HI:
Yeah, I tried nearly everything and I was just about to give up and say run with it, but then I thought about trying a radio with split memories. I just tested the BTech and Luiton radios and all seems to work OK here. I feel it is "good enough".

PR #407, opened 2023-01-18
Variable power support (for CSV)

kk7ds:
- Fix (again) power conversions
- Add generic power level parsing and a helper
- Add a has_variable_power feature flag
- Make CSVRadio support variable power
- Make memedit support radios with variable power

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -209,6 +209,29 @@ class PowerLevel:
         return "%s (%i dBm)" % (self._label, self._power)
 
 
+class AutoNamedPowerLevel(PowerLevel):
+    """A power level that is simply named by its value in watts"""
+    def __init__(self, watts):
+        fmt = ('%iW' if watts >= 10 else '%.1fW')
+        super().__init__(fmt % watts, watts=watts)
+
+
+def parse_power(powerstr):
+    if powerstr.isdigit():
+        # All digits means watts
+        watts = int(powerstr)
+    else:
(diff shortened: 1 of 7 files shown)


kk7ds:
@KC9HI Here's something else you can try to break, since you did good on comments and also found a bug in the power export stuff. This, as I'm sure you saw from the traffic, makes CSV able to be rather free-form with power levels and still (hopefully) convert back and forth from regular radios.

This isn't going to merge tonight, so no rush.

KC9HI:
OK. It is almost 10:30 pm here so I will get it on my computer tonight and then give it a test tomorrow.

kk7ds:
Well, you did better than you thought, actually, because you hit three bugs:

1. The unformatted error message, obviously.
2. The 1585W limit that was supposed to be 1500. This is because of rounding error in the comparison between the entered value and the limit. 1585W is 62.0 dBm, 1584 rounds down to 61dBm as does 1500W :)
3. The failure to reject the bad value is actually a bug that has been in the CSV driver, probably since the beginning. Because it's so simple, it actually wasn't making a copy of the memory that the UI gives to it to store, which means when the UI went later to try to modify it before validation, it was modifying the same copy the CSV driver had stored. So even though we rejected the edit, the damage was done. I expect the way the old UI worked this never manifested, but alas, it's basically a latent bug since 2008!

So, good job :) I'll push up a couple small fixes for those and then call this good. Thanks!

PR #417, opened 2023-01-20
Fix pasting memories over top of immutable fields

kk7ds:
If we're pasting a memory, we need to respect the immutable list of
the memory we're pasting over. So check that any of the immutable
fields are unchanged in the new one we're trying to paste.


kk7ds:
CC @KC9HI

kk7ds:
This should pass tests and restores the immutable behavior for unsanctioned channels that was there before. Hopefully this is pretty much right?

PR #418, opened 2023-01-20
GridSelectNone not available before 4.2.0

creeble:
requirements.txt for Linux shows >=4.0,<4.2.0 but GridSelectNone not available before 4.2.0.

This causes Yaesu FT-65R and ICOM IC-T70 (and probably other bank-based radios) to fail.

### DIFF
--- a/chirp/wxui/bankedit.py
+++ b/chirp/wxui/bankedit.py
@@ -80,8 +80,9 @@ class ChirpBankEdit(common.ChirpEditor):
         self._grid.CreateGrid(
             self._features.memory_bounds[1] - self._features.memory_bounds[0] +
             1, len(self._col_defs))
-        # GridSelectNone only available in >=4.2.0, which is not in requirements.txt for Linux
-        #self._grid.SetSelectionMode(wx.grid.Grid.GridSelectNone)
+        # GridSelectNone only available in >=4.2.0
+        if hasattr(wx.grid.Grid.GridSelectNone):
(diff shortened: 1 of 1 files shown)


chirp/wxui/bankedit.py
```diff
@@ -80,7 +80,8 @@ def __init__(self, radio, *a, **k):
         self._grid.CreateGrid(
             self._features.memory_bounds[1] - self._features.memory_bounds[0] +
             1, len(self._col_defs))
-        self._grid.SetSelectionMode(wx.grid.Grid.GridSelectNone)
+        # GridSelectNone only available in >=4.2.0, which is not in requirements.txt for Linux
+        #self._grid.SetSelectionMode(wx.grid.Grid.GridSelectNone)
```
kk7ds, line 84: This breaks it for everyone else. Can you please just guard it this?
```
if hasattr(wx.grid.Grid.GridSelectNone):
    self._grid.SetSelectionMode(wx.grid.Grid.GridSelectNone)
```

creeble: So done.

I'm sure there are multiple reasons why 4.2.0 isn't used for Linux, but not clear what they would be - just distro standard?

kk7ds: Yeah, all the current debian-based distros have 4.0.7 and installing 4.2.0 via binary wheel is broken on linux. It's...annoying, but once bookworm comes out, we'll hopefully be able to lock to 4.2.0 everywhere shortly thereafter.

Thanks for the change, I'll hit this later today.

PR #419, opened 2023-01-20
Add support for Wouxun KG-UV9PX

mrtjr1:
# CHIRP PR Checklist

* All tests must be passing.


mrtjr1:
Commit text for 8981215 should have read
"Add Support for Wouxun KG-UV9PX"

chirp/drivers/kguv9dplus.py
```diff
@@ -1263,73 +1577,75 @@ def _core_tab(self):
         Radio settings common to all modes and areas go here.
         """
         s = self._memobj.settings
+        if self.MODEL == "KG-UV9PX":
+            sm = self._memobj.screen
```
kk7ds, line 1581: Not critical, but it seems like it would be better to put this down near where it's used, no?

chirp/drivers/kguv9dplus.py
```diff
@@ -1808,20 +2161,25 @@ def _key_tab(self):
...
                                    PF1KEY_LIST,
                                    PF1KEY_LIST[s.pf1])))
         kf.append(RadioSetting("settings.pf2",
-                               "PF2 Key function(56)",
+                               "PF2 Key function (Menu 56)",
                                RadioSettingValueList(
                                    PF2KEY_LIST,
                                    PF2KEY_LIST[s.pf2])))
+        if self.MODEL == "KG-UV9PX":
+            pfkey3 = PF3KEY_LIST2
```
kk7ds, line 2174: I think it would be better to rename this to `PF3KEY_UV9PX` or something. If someone comes later the change the names up above where it's defined, they might accidentally edit the one they don't intend because `LIST` and `LIST2` are not very different.

chirp/drivers/kguv9dplus.py
```diff
@@ -2005,9 +2367,469 @@ def set_settings(self, settings):
...
+                raise InvalidValueError(
+                    "Caller ID code has illegal byte 0x%x" % b)
+            cidstr += bin2ascii[b]
+        return cidstr
+
+    def _get_settings(self):
+        """Build the radio configuration settings menus
+        """
+
+        core_grp = self._core_tab()
```
kk7ds, line 2479: Cool, so this uses the definition from the base class to avoid reproducing it here. Thanks for doing that, I think it's much better.

kk7ds:
Thanks very much for subclassing and separating the cleanup change. I'm going to rebase this on master and then I will merge.

mrtjr1:
When should I rebase?  Always before a comit?

kk7ds:
You were good, I merged something before yours today which is why I rebased yours. It's just a button in the web UI for me if there are no conflicts, so not a huge deal. But, in general it's good to rebase before you submit a PR, as it's possible there could be test changes that you're no longer passing or something.

chirp/drivers/kguv9dplus.py
mrtjr1: agree... will update it in a future clean-up

chirp/drivers/kguv9dplus.py
mrtjr1: agree... will update it in a future clean-up

PR #422, opened 2023-01-21
Verify VX-2

kk7ds:
Tested read/modify/write/verify for editing: memory frequencies/labels, settings (opening message text).
System: macOS 10.14.6 and SparkFun DEV-14050 (CH340G) breakout


kk7ds:
@thomasrussellmurphy here

PR #424, opened 2023-01-22
Upgrade gmrsv2.py to use mem.immutable (revisited) - #10288

KC9HI:
This model supports "fully customizable channels" which created some additional challenges.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -1067,6 +1067,12 @@ class ValidationError(ValidationMessage):
     pass
 
 
+def split_validation_msgs(msgs):
+    """Split a list of msgs into warnings,errors"""
+    return ([x for x in msgs if isinstance(x, ValidationWarning)],
+            [x for x in msgs if isinstance(x, ValidationError)])
+
+
 class Alias(object):
     VENDOR = "Unknown"
     MODEL = "Unknown"
--- a/chirp/drivers/gmrsv2.py
+++ b/chirp/drivers/gmrsv2.py
@@ -392,6 +392,22 @@ class GMRSV2(baofeng_common.BaofengCommonHT):
         """Process the mem map into the mem object"""
         self._memobj = bitwise.parse(self.MEM_FORMAT, self._mmap)
 
+    def validate_memory(self, mem):
+        msgs = baofeng_common.BaofengCommonHT.validate_memory(self, mem)
+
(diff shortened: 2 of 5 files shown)


KC9HI:
I really like how this works not. It almost passed with flying colors and I was about to tell you to go ahead and merge this. Unfortunately while copying channels from my radio into a GMRS-V2 image, 3 channels encountered errors. The errors for all 3 were the same so I just copied one of them. But I will attach my image file so you can test. The problem channels are 10, 11 and 12. One uses "split" and the other 2 use "-" for the duplex.
[GMRS-V2_errors.zip](https://github.com/kk7ds/chirp/files/10530538/GMRS-V2_errors.zip)

KC9HI:
Oops. I like how it works 'now'.

kk7ds:
Ah yep, just a typo basically. Fixed now, I think, thanks!

KC9HI:
OK. Great. I am in the process of adding RB17P to this. Will push it to you soon.

kk7ds:
Roger, I'll wait for that before merge, thanks!

KC9HI:
With what you have done, there shouldn't be a need for validate_memory() like being added to GMRS-V2, right?

kk7ds:
No, `validate_memory()` *is* required. That's how we get the warning that "you pasted duplex + but I'm going to change it to off because I have to". Without that, the user gets no indication why duplex was changed to off, and I definitely prefer that we explain to them so they're not thinking it's a bug.

KC9HI:
OK. Once I add validate_memory(), it should be ready. I need to a a "-f" to force it, right?

kk7ds:
If you're up-to-date and just adding another commit on top, then you don't. However, if you checked out the branch before I made the typo fix then yeah, just push -f and I'll re-fix it after you do.

kk7ds:
Ah, looks like you did commit `--amend` so the RB17 is in the same commit as the GMRSv2. That's okay I'll just update the commit message when I re-fix the typo.

KC9HI:
Yeah. I didn't realize could leave the --amend of and start another commit on the same branch. I'll have to remember that for the future.

kk7ds:
Yep, I regularly put several related commits in one PR. Anyway, no big deal, should be good now. Thanks!

KC9HI:
This is great! I can now go back to working on my list. Thank you.

PR #429, opened 2023-01-23
Update mursv1.py driver to use mem.immutable - #10288

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/mursv1.py
+++ b/chirp/drivers/mursv1.py
@@ -22,6 +22,7 @@ import re
 from chirp.drivers import baofeng_common
 from chirp import chirp_common, directory, memmap
 from chirp import bitwise, errors, util
+from chirp import bandplan_na
 from chirp.settings import RadioSettingGroup, RadioSetting, \
     RadioSettingValueBoolean, RadioSettingValueList, \
     RadioSettingValueString, RadioSettingValueInteger, \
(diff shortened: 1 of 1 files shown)


chirp/drivers/mursv1.py
```diff
@@ -67,8 +68,7 @@
 LIST_VOICE = ["Off", "English", "Chinese"]
 LIST_WORKMODE = ["Frequency", "Channel"]
 
-MURS_FREQS = [151820000, 151880000, 151940000, 154570000, 154600000] * 3
-FM_MODE = [3, 4, 8, 9, 13, 14]
+MURS_FREQS = bandplan_na.ALL_MURS_FREQS * 3
```
kk7ds, line 71: I saw this in the other driver, what's this `* 3` about?

KC9HI: The radio has 15 channels. 
ALL_MURS_FREQS + ALL_MURS_FREQS +ALL_MURS_FREQS or ALL_MURS_FREQS * 3

kk7ds: Okay, but.. you don't use this list anywhere to initialize the radio do you? If you just use it for checking whether something is a MURS frequency or not, then this:

```
if mem.freq in ALL_MURS_FREQS
```

is the same whether `ALL_MURS_FREQS` has just the unique frequencies, or three copies of the unique frequencies. It's not wrong, I just don't think it's necessary to triplicate (is that a word?) them for that purpose, so I was wondering if there was some other reason.

KC9HI: I do use it. But now that you mention it, I might be using it wrong in one or more places. Let me review it.

PR #434, opened 2023-01-24
Remove some debug and tweak tests

kk7ds:
- Remove accidental debug
- Decorate driver tests with the module name
- Add missing Icom IC-E90 test image


kk7ds:
@mrtjr1 The third commit in this series will let you use `'-k module'` where module is the name of a file without the `.py` and it will run tests for all the drivers in that file. So that's *close* to what you're asking for.

PR #436, opened 2023-01-25
Install .desktop and icon files

masenf:
For linux platforms, install a .desktop file and icon for launching CHiRP.

This works for both user-installs (~/.local) and system-installs. Uninstalling via `pip uninstall` properly removes the files.

--------

Allow `python3 -m chirp.wxui`, to start chirp without invoking the console script `chirp`. If the user hasn't added `~/.local/bin` to their PATH, then the desktop launcher can't easily launch the script. This avoids having to set a different Exec= line in the desktop file for user vs. system install (.local/bin/chirp vs chirp)

# CHIRP PR Checklist

* All tests must be passing.


kk7ds:
CC @vishwin

chirp/wxui/__main__.py
```diff
@@ -0,0 +1,4 @@
+"""Entry point for starting chirp via `python -m chirp.wxui`"""
+from . import chirpmain
+
+chirpmain()
```
kk7ds, line 4: I don't understand why we need to do this (meaning you running it with `python3 -mchirp.wxui` in the desktop file) when we have `chirp` installed via `console_scripts`. Can't you just run `chirp`?

kk7ds: Okay makes sense. Annoying, but fair enough :)

I'mma wait for @vishwin to look over this as he was pushing the `setup.py` to the current state.

vishwin:
Apart from `chirp.wxui.__main__`, all of this is not necessary and actually wrong, as [`data_files` is deprecated](https://github.com/pypa/setuptools/discussions/2648).

Within the Python packaging ecosystem, there is no reliable way to install and manage platform-specific files and locations like `.desktop`, hence why they are located within the package and left as further exercises (for system package managers, etc). `entry_points` is the lone exception as it can be reliably managed across platforms. Certain users not having `~/.local/bin` in `$PATH` is not our problem, but having `chirp.wxui` as an executable module helps with providing entry points, especially when not installed as a (Python) package.

As for current state of `setup.py`, it is officially temporary and a stop-gap until we implement PEP-517 support.

masenf:
Thanks for looking. Will leave the .desktop as documentation then and submit separate PR for the main entry point.

kk7ds:
@masenf How about a command-line option to install a desktop file, which we only expose on linux? Then once you've got it installed, you can:
```
$ python3 -mchirp.wxui --install-desktop
```
once and it'll just barf the desktop file into the user or system directory depending on if the system location is writeble or not?

PR #438, opened 2023-01-25
Quick zipvalidator fix.

jebbgrenham:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/wxui/query_sources.py
+++ b/chirp/wxui/query_sources.py
@@ -526,6 +526,15 @@ class RRQueryDialog(QuerySourceDialog):
 
         return vbox
 
+    def _call_validations(self, parent):
+        if parent == self:
+            # If we're calling valiations at the top-level, redirect that to
+            # the start of our country-specific widgets, based on whatever tab
+            # is selected. This avoids trying to validate US things when
+            # Canada is selected and so on.
+            parent = self.tabs.GetPage(self.tabs.GetSelection())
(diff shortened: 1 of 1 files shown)


kk7ds:
Thanks for working on this. What you've got here is definitely a quick fix, but it means that US people get a default zipcode of `00000` in their box, which is kinda weird looking. It also means that if I type garbage in the zipcode box and then switch to canada, it still fails to execute. What about this:
```
diff --git a/chirp/wxui/query_sources.py b/chirp/wxui/query_sources.py
index 703b7bdc..9d475956 100644
--- a/chirp/wxui/query_sources.py
+++ b/chirp/wxui/query_sources.py
@@ -526,6 +526,15 @@ class RRQueryDialog(QuerySourceDialog):

         return vbox

+    def _call_validations(self, parent):
+        if parent == self:
+            # If we're calling valiations at the top-level, redirect that to
+            # the start of our country-specific widgets, based on whatever tab
+            # is selected. This avoids trying to validate US things when
+            # Canada is selected and so on.
+            parent = self.tabs.GetPage(self.tabs.GetSelection())
+        return super()._call_validations(parent)
+
```
That makes us only validate the fields that are south of the selected country tab. This way anything can be in the US box and it'll still only validate canada if that's what is selected.

What do you think?

jebbgrenham:
Much better and significantly more robust.

PR #439, opened 2023-01-25
Various fixes and tests

kk7ds:
- Fix Icom IC-E90 special memory editing
- Fix editing specials with negative indexes
- Add special memories test and fix issues
- Add a test to check for memory name whitespace
- Add a test and fixes for trailing name whitespace
- Make FloatProperty honor precision


kk7ds:
FYI @mrtjr1 The last commit here has the fix for the float property rendering.

mrtjr1:
It works perfectly - Thanks!!!
[image: image.png]

PR #442, opened 2023-01-26
Update radioddity_r2.py to use mem.immutable - #10288

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/radioddity_r2.py
+++ b/chirp/drivers/radioddity_r2.py
@@ -513,6 +513,31 @@ class RadioddityR2(chirp_common.CloneModeRadio):
         compand.set_doc("Compress Audio for TX")
         mem.extra.append(compand)
 
+        immutable = []
+
+        if self._frs16:
+            if mem.freq in FRS16_FREQS:
+                if mem.number >= 1 and mem.number <= 16:
+                    FRS_FREQ = FRS16_FREQS[mem.number - 1]
+                    mem.freq = FRS_FREQ
+                mem.duplex == ''
+                mem.offset = 0
+                mem.mode = "NFM"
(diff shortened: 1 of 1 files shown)


chirp/drivers/radioddity_r2.py
```diff
@@ -513,6 +513,31 @@ def get_memory(self, number):
...
+
+        if self._frs16:
+            if mem.freq in FRS16_FREQS:
+                if mem.number >= 1 and mem.number <= 16:
+                    FRS_FREQ = FRS16_FREQS[mem.number - 1]
+                    mem.freq = FRS_FREQ
+                mem.duplex == ''
+                mem.offset = 0
+                mem.mode = "NFM"
+                immutable = ["empty", "freq", "duplex", "offset"]
```
kk7ds, line 526: Hey @KC9HI I was just fixing a test that wasn't running by accident and I found a failure that I think is related. Shouldn't `mode` be in this list to restrict it to `NFM`? The test fails because it thinks it can change mode, but it always comes back `NFM`...

KC9HI: Yes. FRS is always NFM. I either forgot add "mode" or I had brain fade and was thinking it was like GMRS where it can be either on the 462.xxx frequencies. So it should be 

immutable = ["empty", "freq", "duplex", "offset", "mode"]

just like it is in the PMR section.

KC9HI: Well not exactly like PMR. PMR includes "power", which FRS16 should not.

kk7ds: Cool, I'll fix in my set to fix the test, thanks.

KC9HI:
Thank you

PR #444, opened 2023-01-27
Record Yaesu FT-65E as working

kwirk:
Successfully tested download, modify memory and setting, upload, and verified changes on radio.


kwirk:
> Somewhat related, any chance you could look at: https://chirp.danplanet.com/issues/10286 ?
> 
> You might have some data points to offer there, and on the related FT-65 bug.

Sure. I'll take a look and comment on the issue, and the relate FT-65 one.

PR #451, opened 2023-01-29
Fix brute force duplex test

kk7ds:
- Fix setting duplex=split on TK-8160
- Fix can_odd_split flag in FRS-A1 and RB27
- Fix brute force duplex test

### DIFF
--- a/tests/test_brute_force.py
+++ b/tests/test_brute_force.py
@@ -87,8 +87,10 @@ class TestCaseBruteForce(base.DriverTest):
             self.set_and_compare(m)
 
     @base.requires_feature('valid_duplexes')
-    def do_duplex(self):
+    def test_duplex(self):
         m = self.get_mem()
+        if 'duplex' in m.immutable:
+            self.skipTest('Test memory has immutable duplex')
         for duplex in self.rf.valid_duplexes:
             if duplex not in ["", "-", "+", "split"]:
                 continue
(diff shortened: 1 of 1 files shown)


kk7ds:
@KC9HI This is the other test I found that wasn't running properly. Several of the drivers that needed fixing here were obvious changes. The last one is the RB17P, which fails because of GMRS channels with "opinionated" duplex requirements. I assume this is on your list to immutable-ify right? I'd like to get this test re-enabled (it would have caught a bug in a driver I just wrote a few days ago), but can wait if the RB17P is coming soon. If not, maybe I should just add an exception for it.

KC9HI:
I can start on this. Do I do it for now under this PR or do I create my own local branch and then push it when I am done?

kk7ds:
Okay, I don't need to derail anything else you've got going on, I just wanted to know if it was on the horizon. But yeah, I'd say just do it in a separate branch and once that's merged, I'll rebase this. I can pull the fixes out of this and merge them separately in the meantime too. Thanks!

KC9HI:
It was one of several on my list that I was on a holding pattern because it was fully customizable like the GMRS-V2 and wouldn't pass tests. You had something for me to test once I got to a good place, which today probably is one. I don't know if can remember which PR it was.

KC9HI:
I may have forgotten. Let me see how it goes with RB17P. That will certainly refresh my memory.

kk7ds:
Okay, you might want to check out #424, and if that looks better, do the RB17P in that branch so you have the gui changes available. Then you can commit it on top and push the result up to that PR so they're both together. Totally your call though.

PR #457, opened 2023-01-30
Update btech.py to use mem.immutable - #10288

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/btech.py
+++ b/chirp/drivers/btech.py
@@ -758,9 +758,26 @@ class BTechMobileCommon(chirp_common.CloneModeRadio,
     def validate_memory(self, mem):
         msgs = chirp_common.CloneModeRadio.validate_memory(self, mem)
 
-        _msg_duplex1 = 'Memory location only supports "Low"'
-        _msg_duplex2 = 'Memory location only supports "off"'
-        _msg_duplex3 = 'Memory location only supports "(None)", "+" or "off"'
+        _mem = self._memobj.memory[mem.number]
+        _msg_duplex = 'Duplex must be "off" for this frequency'
+        _msg_offset = 'Only simplex or +5MHz offset allowed on GMRS'
+
+        if self.MODEL == "GMRS-20V2":
+            if mem.freq not in GMRS_FREQS:
+                if mem.duplex != "off":
+                    msgs.append(chirp_common.ValidationWarning(_msg_duplex))
+            elif mem.duplex and mem.offset != 5000000:
+                msgs.append(chirp_common.ValidationWarning(_msg_offset))
+        if self.MODEL == "GMRS-50X1":
+            if not (mem.number >= 1 and mem.number <= 30):
+                if mem.duplex != "off":
+                    msgs.append(chirp_common.ValidationWarning(_msg_duplex))
+        if self.MODEL == "DB25-G":
+            if mem.freq in GMRS_FREQS3:
+                if mem.duplex and mem.offset != 5000000:
+                    msgs.append(chirp_common.ValidationWarning(_msg_offset))
+                if mem.duplex and mem.duplex != "+":
+                    msgs.append(chirp_common.ValidationWarning(_msg_offset))
 
         return msgs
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_chirp_common.py
```diff
@@ -715,7 +715,9 @@ class TestOverrideRules(base.BaseTest):
     # please ask permission first.
     IMMUTABLE_WHITELIST = [
         # Uncomment me when the time comes
+        'BTECH_GMRS-20V2',
         'BTECH_GMRS-V2',
+        'Radioddity_DB25-G',
```
kk7ds, line 720: Is this the error?
```

    def test_radio_overrides(self):
        for rclass in directory.DRV_TO_RADIO.values():
            if directory.radio_class_id(rclass) in self.IMMUTABLE_WHITELIST:
                self._test_radio_override_calls_super(rclass)
            else:
>               self._test_radio_override_immutable_policy(rclass)
```

I thought it was different. This is because one of the radios that has overridden `check_set_memory_immutable_policy` is not in this whitelist. Perhaps a subclass of one of these? I'll pull it down and see if I can figure it out.

kk7ds: It's this one: `BTECH_UV-2501`

I'll modify the error message so it shows the radio in the future.

kk7ds: Okay, it's a bunch I guess:
```
        'BTECH_UV-2501',
        'BTECH_UV-2501+220',
        'BTECH_UV-5001',
        'WACCOM_MINI-8900',
        'QYT_KT-UV980',
        'QYT_KT8900',
        'QYT_KT8900R',
        'LUITON_LT-588UV',
        'BTECH_UV-25X2',
        'BTECH_UV-25X4',
        'BTECH_UV-50X2',
```

So we probably need to do something to separate those out, because I'm assuming they all don't need the same override rules and having them in the whitelist like that defeats the point.

kk7ds: Okay yeah, so I think what you want is to add the `BTECH_GMRS-50X1` to this whitelist and then only put the `check_memory_set_immutable_policy()` override on the GMRS-20V2, GMRS-50X1, and DB25-G classes, as (from the look of it) only those in `btech.py` are the ones that should have the special behavior, is that right?

KC9HI:
Correct. Except I wasn't sure if the GMRS-50V1 should be added because it isn't a "fully configurable memory". Its 30 GMRS channels are immutable and shouldn't be changed or removed.

Then later when the  GMRS-50V2 is released, it would be added.

kk7ds:
Okay, I added it to make the test pass because of immutable failures, but perhaps that means something was left out of the immutable set that is being returned and it shouldn't have the special handler. In that case, don't add the policy override and don't add it to the whitelist.

KC9HI:
OK. How do I get a copy of what you have done so I can continue after supper?

chirp/drivers/btech.py
```diff
@@ -923,8 +930,92 @@ def get_memory(self, number):
...
+                        immutable += ["duplex", "offset"]
+                    elif mem.number >= 15 and mem.number <= 22:
+                        mem.duplex = ''
+                        mem.offset = 0
+                        immutable += ["duplex", "offset"]
+                    elif mem.number >= 23 and mem.number <= 30:
+                        mem.duplex = '+'
+                        mem.offset = 5000000
+                        immutable += ["duplex", "offset"]
+                else:
```
kk7ds, line 959: Er, this makes it look like the channels aren't all immutable.. can you store non-sanctioned channels in it? If so, I think it needs the policy override so that we don't apply the immutable rules of the channel you're writing to to the thing you're trying to store. The policy check lets you basically say "just because something is immutable in channel X right now doesn't mean it will be immutable if you put something different into it". By my read of this section from L936 to L964, that needed override applies to the 50X1.

kk7ds:
I can push it up on top of here and you can have a look.

kk7ds:
Okay, what I just pushed is what I think you want based on my reading of the behavior of the 50X1 accepting other-than-GMRS channels in any spot. If my read on that was wrong, then this isn't right.

chirp/drivers/btech.py
KC9HI: Since immutable = ["empty", "freq"] at L941, immutable becomes ["empty", "freq", "duplex", "offset"] at line 958. That should prevent it from being deleted and from changing the frequency. Right?

KC9HI:
Do I get this as #454? Do I need to remove my immutable-btech branch first?

chirp/drivers/btech.py
kk7ds: Ah, because there's no `elif` at L942... okay. So only memories 1-30 are fixed, but 0, and 30-255 are allowed to be anything, but without transmit ability regardless of what they are?

kk7ds: Ah, because there's no `elif` at L942... okay. So only memories 1-30 are fixed, but 0, and 30-255 are allowed to be anything, but without transmit ability regardless of what they are?

kk7ds:
You can `gh pr checkout 454 -f` and it will just reset your local `immutable-btech` branch to what this is currently.

chirp/drivers/btech.py
KC9HI: Correct.

KC9HI:
Great. That "-f" can come in handy for several things.

chirp/drivers/btech.py
kk7ds: Okay, so the immutable-ness of 1-30 is always sticky, but the immutable-ness of 0,30-255 is not (although it will be overridden anyway). So I think the policy check should only set `immutable=[]` if the number is not 1-30, and then make validate return a warning for the "duplex gonna be off" thing. This might be too obscure for the tests, so if it seems like it's all right and still not passing after that, push it up again and I'll have a look at it.

KC9HI: I'm not 100% sure what you mean by "the policy check should only set immutable=[] if the number is not 1-30." Do you mean something like this?

    def check_set_memory_immutable_policy(self, existing, new):
        if not (mem.number >= 1 and mem.number <= 30):
            existing.immutable = []
            super().check_set_memory_immutable_policy(existing, new)

Other than having done this part, everything else seems to work OK and passes the tests.

KC9HI: Gotcha...

KC9HI: It tells me "NameError: name 'mem' is not defined. How do I pass the memory number to this?

kk7ds: oh, I didn't read close enough. You want to check `new.number`. Both `new` and `existing` are memories.

KC9HI: I think that got it. I forgot to fetch and rebase before I sent it again. I can to that in a little bit if you need me to.

kk7ds: I rebased it, should be okay, thanks!

PR #458, opened 2023-01-31
Wouxun KG-935G Call Code fix

mrtjr1:
Fixes: 10333

Correct issue where ANI-EDIT and Call Codes would not get updated on the radio.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/kg935g.py
+++ b/chirp/drivers/kg935g.py
@@ -31,7 +31,8 @@ from chirp import util, chirp_common, bitwise, memmap, errors, directory
 from chirp.settings import RadioSetting, RadioSettingGroup, \
     RadioSettingValueBoolean, RadioSettingValueList, \
     RadioSettingValueInteger, RadioSettingValueString, \
-    RadioSettingValueFloat, RadioSettingValueMap, RadioSettings
+    RadioSettingValueFloat, RadioSettingValueMap, RadioSettings, \
+    InvalidValueError
 
 
 LOG = logging.getLogger(__name__)
@@ -174,27 +175,8 @@ _MEM_FORMAT = """
 
     #seekto 0x0500;
     struct {
-        u8    call_code_1[6];
-        u8    call_code_2[6];
-        u8    call_code_3[6];
-        u8    call_code_4[6];
-        u8    call_code_5[6];
-        u8    call_code_6[6];
-        u8    call_code_7[6];
-        u8    call_code_8[6];
-        u8    call_code_9[6];
-        u8    call_code_10[6];
-        u8    call_code_11[6];
-        u8    call_code_12[6];
-        u8    call_code_13[6];
-        u8    call_code_14[6];
-        u8    call_code_15[6];
-        u8    call_code_16[6];
-        u8    call_code_17[6];
-        u8    call_code_18[6];
-        u8    call_code_19[6];
-        u8    call_code_20[6];
-    } call_groups;
+        u8 cid[6];
+    } call_ids[20];
 
     #seekto 0x0580;
     struct {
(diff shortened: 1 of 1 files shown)


chirp/drivers/kg935g.py
```diff
@@ -430,6 +412,35 @@
...
+            "Caller ID must be at least 3 and no more than 6 digits")
+    if s[0] == '0':
+        raise InvalidValueError(
+            "First digit of a Caller ID cannot be a zero '0'")
+    blk = bytearray()
+    for c in s:
+        if c not in ascii2bin:
+            raise InvalidValueError(
+                "Caller ID must be all digits 0x%x" % c)
+        b = (0x0, 0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8, 0x9)[int(c)]
```
kk7ds, line 433: This is what you want here:
```
b = asc2bin.index(c)
```

chirp/drivers/kg935g.py
```diff
@@ -430,6 +412,35 @@
...
+        blk.append(b)
+    if len(blk) < 6:
+        blk.append(0xf)  # EOL a short ID
+    if len(blk) < 6:
+        for i in range(0, (6 - len(blk))):
+            blk.append(0xf0)
```
kk7ds, line 439: Surely you can combine L435-L439 right? However, if you use a `bytes` instead, you can do:
```
blk.ljust(6, 0xF0)
```
Which makes sure to fill up to 6 positions with a byte.

chirp/drivers/kg935g.py
```diff
@@ -941,23 +955,14 @@ def _get_settings(self):
...
-                value = []
-                for j in range(0, 6):
-                    try:
-                        value.append(callchars.index(str(setting.value)[j]))
-                    except IndexError:
-                        value.append(0xFF)
-                obj.call_code = value
-            rs.set_apply_callback(apply_call_code,
-                                  eval("_call.call_code_"+callnum))
+            x = i-1
```
kk7ds, line 958: Spaces around the `-` please. Our style checker doesn't get these, but most do, so when I see them, I try to point them out for the future.

kk7ds (commented):
A couple small comments. Can you fix these tonight? If so, I'll hold off on merging this.

chirp/drivers/kg935g.py
mrtjr1: will fix right now

mrtjr1: where does asc2bin come from.. not seeing it code
do I need to import it?

chirp/drivers/kg935g.py
kk7ds: So on this one, instead of bytearray above, try:
```
blk = bytes()
for c in s:
  blk.append(asc2bin.index(c))
return blk.ljust(6, b'\xF0')
```

chirp/drivers/kg935g.py
kk7ds: Oh sorry I meant `ascii2bin`, L418

chirp/drivers/kg935g.py
mrtjr1: ok - this passed the tox style check --- must be a different set of rule checks.  will fix.

kk7ds: Yeah, that's why I'm saying ours does, but it's too loose and I want to tighten it up, so I just try to spot these when I can to reduce the number of things we'll have to fix later. Not a huge deal, but if you're revising, might as well.

chirp/drivers/kg935g.py
mrtjr1: so this one has 2 different fills.  
end of data byte is 0xF while the rest will be filled with 0xF0

mrtjr1: probably should be cleaner and use 0x0F and 0xF0 to not add confusion

kk7ds: Ah, okay I missed that. Just leave it then, that's fine.

kk7ds:
Mel saves the day! Didn't have any meat to commit today to generate a build tomorrow. Almost broke the streak... Thanks!

mrtjr1:
I am sure you could use some time off.   Sorry to create more work for you!

PR #459, opened 2023-02-01
Update ga510.py driver to use mem.immutable - #10288

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/ga510.py
```diff
@@ -285,6 +278,30 @@ def get_features(self):
...
+                if mem.duplex and mem.duplex != "+":
+                    msgs.append(chirp_common.ValidationWarning(_msg_offset))
+        if self.MODEL == "RA685":
+            if not ((mem.freq >= self.vhftx[0] and mem.freq < self.vhftx[1]) or
+                    (mem.freq >= self.uhftx[0] and mem.freq < self.uhftx[1])):
+                if mem.duplex != "off":
+                    msgs.append(chirp_common.ValidationWarning(_msg_duplex))
+
+        return msgs
+
```
kk7ds, line 304: I guess I wasn't really paying close enough attention to the others, but we really should put these two chunks (without the if) on the subclasses. Meaning only override validate_memory for those special subclasses instead of doing the `if self.MODEL`. The latter works, but it's much less conventional in python, even though we have a lot (too much) of it in chirp. Not a huge deal, and I already let the others in so I won't hold this up, but ... my OCD...

chirp/drivers/ga510.py
```diff
@@ -285,6 +278,30 @@ def get_features(self):
                           (400000000, 480000000)]
         return rf
 
+    def validate_memory(self, mem):
+        msgs = chirp_common.CloneModeRadio.validate_memory(self, mem)
```
kk7ds, line 282: This is really the python2 syntax for calling the parent, and I'm so used to the old way I forget that it's technically wrong now. But it really should be `super().validate_memory(mem)` in python3. It works, but the python2 way could stop working at some point in the python3.x timeline, so the fewer things we add, the better.

KC9HI: OK. I will make another list. ;-)

kk7ds: I know, it's annoying. Being through 15 years of python development it's hard to keep track of seemingly senseless changes like this. They all have reasons, but sometimes it's hard to be on board with it all :)

KC9HI: Not annoyed. I'm happy to find out now. List made. There are 8 drivers that need attention. Nothing urgent. Should be easy to do. My kind of task. ;)

PR #465, opened 2023-02-02
TK-3140 and several fixes

kk7ds:
- Allow registering multiple formats with same ext
- Expand repeaterbook tests to another state
- Fix truncated multiword names in properties box
- Add Kenwood TK-2140/3140 support


kk7ds:
PR checks set off by the import of `builtins` to provide fake translation support. False positive for legacy stuff.

PR #467, opened 2023-02-03
Update es_ES.po

MELERIX:
Update Spanish Translation

### DIFF
--- a/chirp/locale/es_ES.po
+++ b/chirp/locale/es_ES.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


kk7ds:
I've been looking for someone to confirm that the translation stuff is working. In the bug you opened about the TYT, it looked from the logs as though it could not detect the locale from the system. Are you getting translations in the UI?

kk7ds:
Okay, some of the things that you see as translated may be the "stock items" -- things like the file menu and not chirp-specific ones. So, I merged this, please let me know tomorrow if your new translations show up. If not, maybe you can help me test some fixes.

kk7ds:
Okay, the menu items are actually the stock things that should be translated by the system. I'm not sure why those are not working but the others are, I'll have to look into that.

The radio cloning message(s) are in the drivers and haven't historically been translated.

The welcome screen is an image and can't really be translated.

PR #479, opened 2023-02-07
New Model Support- Wouxun KG-935G Plus

mrtjr1:
Fixes: #10355

Add support for Wouxun KG-935G Plus

# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/kg935g.py
```diff
@@ -597,32 +893,42 @@ def _upload(self):
...
-                raise Exception("Radio did not ack block %i" % ptr)
-            ptr += blocksize
-            if self.status_fn:
-                status = chirp_common.Status()
-                status.cur = i
-                status.max = end
-                status.msg = "Cloning to radio"
-                self.status_fn(status)
+    def _do_upload(self):
+        cfgmap = config_map_935G
```
kk7ds, line 904: I'm missing something. It seems like you're changing `do_upload()` so it can handle multiple regions in the map to upload, but you've only got one map for the 935G. Did the 935G not really care but the 935G plus *does*, yet the same map works for both?

mrtjr1: The 935G driver originally just fully uploaded all 0x8000 bytes.   This change makes the driver only update the values in known memory areas.  The 2 radios have slightly different maps but the memory regions that contain the known values is the same.  Since both radios identify the same way, I don't know what is in the unknown areas and think it is good to only upload what is known.

mrtjr1: If you mean why am I setting cfgmap to the 935G map....when I first got the plus, I was prepared to need 2 maps and implemented an IF --Else to assign cfg map the right map.   Since it isn't needed, I just left the local assignment and removed the IF....

kk7ds: Yeah, this is all what I meant, just checking.

kk7ds:
This had a looot of cleanup mixed in with the actual fixes. I'd really prefer to have as much of that be separate, but I know it's a pain. Imma let you slide this time :)

mrtjr1:
Thanks Dan.    Appreciate it.   So as long as the clean ups are done in a 2nd commit, and included in a single PR... all is good?

kk7ds:
Yeah, for sure, stack the cleanups under or over the actual thing you want to change in a single PR, that's fine. It's just good to separate things that change the *operation* from things that make things *look* better. If we find a bug later and are trying to track down why it's broken, it's much easier to look at smaller commit with smaller scope.

PR #480, opened 2023-02-07
Fix partial translation issue

kk7ds:
- Use wx.Locale instead of gettext
- Fix untranslated menu names
- Refresh locale files


kk7ds:
PR checks failed just because this imports builtins to install the marking function.

PR #487, opened 2023-02-10
Fix win32 driver import

kk7ds:
- Fix frozen check for win32 driver import

### DIFF
--- a/chirp/directory.py
+++ b/chirp/directory.py
@@ -180,8 +180,9 @@ def get_radio_by_image(image_file):
 
 
 def import_drivers(limit=None):
-    if sys.platform == 'win32':
-        # Assume we are in a frozen win32 build, so we can not glob
+    frozen = getattr(sys, 'frozen', False)
+    if sys.platform == 'win32' and frozen:
+        # We are in a frozen win32 build, so we can not glob
         # the driver files, but we do not need to anyway
         import chirp.drivers
         for module in chirp.drivers.__all__:


kk7ds:
@KC9HI This should cause the tests to fail and actually show the right error. Can you try and see? I still can't reproduce the error you have (even on windows) but I have a good idea of the cause.

KC9HI:
[errors2.txt](https://github.com/kk7ds/chirp/files/10702829/errors2.txt)

KC9HI:
No sorry. I was just thinking I needed to fetch/rebase. Hang on.

kk7ds:
Okay, the thing I was thinking might be obvious isn't the thing, so I'm not quite sure what's going on, but the hopefully better error from this patch might help.

KC9HI:
succeeded. want me to copy my new driver into it?

kk7ds:
Succeeded meaning tests didn't fail?

KC9HI:
[errors3.txt](https://github.com/kk7ds/chirp/files/10702887/errors3.txt)

KC9HI:
Correct. It must be something that is triggered by my new driver.

kk7ds:
Okay I'm confused, is `errors3.txt` with or without you driver in the tree?

KC9HI:
errors3.txt includes my driver. without it the tests passed

kk7ds:
Ah, okay. So your driver must be deleting a tone from the master list in `chirp_common.TONES`. Those other drivers do the right thing, which is make a copy and then delete from the copy:
```
MAH_TONES = list(chirp_common.TONES)
MAH_TONES.remove(88.5)
```
Perhaps you're missing the first step in your new driver?

KC9HI:
This is what I am doing

PMR_TONES = chirp_common.TONES
[PMR_TONES.remove(x) for x in [ 69.3, 159.8, 165.5, 171.3, 177.3, 183.5,
                               189.9, 196.6, 199.5, 206.5, 229.1, 254.1]]

PMR_DTCS_CODES = chirp_common.DTCS_CODES
[PMR_DTCS_CODES.remove(x) for x in [ 36,  53, 122, 145, 212, 225, 246,
                                    252, 255, 266, 274, 325, 332, 356,
                                    446, 452, 454, 455, 462, 523, 526]]

kk7ds:
Yep, you're breaking the master list by removing the tone from it without making a copy. Then, when the other drivers come along after yours, they make a copy, but of the mutilated master list and then fail to remove the tone from their copy, because you already removed it from the master one they made a copy of. Make sense?

We actually need to convert the master lists to tuples so that you'll get an error when you try to do what you're doing, but several drivers rely on them being lists. So we should fix that after this. But, make your first line do `list(chirp_common.TONES)` and the problem should go away. Confirm that and I'll merge this to fix the obscure test error problem.

KC9HI:
Success. That worked. Thank you.

kk7ds:
Well, equal is not the right word, but yeah. It's because:
```
PMR_TONES = chirp_common.TONES
```
Does not create a copy, it just points to the same list that `chirp_common.TONES` points to. So you can use either "name" to modify the list, but its the same list in both cases. But if you do:
```
PMR_TONES = list(chirp_common.TONES)
```
the `list()` call makes a copy of the thing you pass to it and returns that. So once you do that, `PMR_TONES` points to a list that is exactly like the original (which I would say is "equal") but it's not the same list, it's a copy. Then you can modify it and the original won't be changed.

Okay, so I'll merge this so that we'll see better errors in the future. Then I'll put up a patch to change the master list to a tuple, but a lot of drivers will fail with that. Then maybe we can iterate through fixing those. Once we do that, you'll get an electric shock when you try to modify those, instead of this obscure "other driver fails somewhere else" behavior.

KC9HI:
OK. Got it. Thanks again.

PR #488, opened 2023-02-10
Add Retevis RB28 - fixes #10324

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/retevis_rb28.py
```diff
@@ -0,0 +1,717 @@
...
+
+    try:
+        serial.write(b"\x02")
+        ident = serial.read(8)
+    except:
+        raise errors.RadioError("Error communicating with radio")
+
+    # check if ident is OK
+    itis = False
+    for fp in radio._fingerprint:
```
kk7ds, line 191: Why not just do:
```
itis = ident in radio._fingerprint
```
? Is `ident` longer than all the fingerprints? Doesn't matter much of course. Another trick to do what you're doing but shorter is to use `else` on the for, which is only called if no break happens:
```
for fp in radio._fingerprint:
    if fp in ident:
        break
else:
    # No break ever got called, which means we didn't find it
    raise errors.RadioError('Radio identification failed')

# If we made it this far, we're good.
```

KC9HI: It is what Pavel developed when he worked with me on the original btech.py driver. I like this, though. I will keep it in mind. Thanks for the "trick".

PR #489, opened 2023-02-10
Make common master lists immutable

kk7ds:
This makes several of our master lists of things immutable by making
them tuplex instead of lists. This will help to avoid accidental (or
intentional) modification of those lists. As it turns out, the
ft450 and ts580 drivers were doing this, so this removes that
mutilation from those drivers. That means they are a little broken
now for memories with those modes, but those drivers are broken for
python3 anyway.

The rest of the drivers changed here just needed some minimal tweakage
to use tuples when adding things to the base lists.

This also (necessarily) simplifies the RadioFeatures type checking,
which relied on special behaviors for tuples.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -30,25 +30,27 @@ LOG = logging.getLogger(__name__)
 SEPCHAR = ","
 
 # 50 Tones
-TONES = [67.0, 69.3, 71.9, 74.4, 77.0, 79.7, 82.5,
-         85.4, 88.5, 91.5, 94.8, 97.4, 100.0, 103.5,
-         107.2, 110.9, 114.8, 118.8, 123.0, 127.3,
-         131.8, 136.5, 141.3, 146.2, 151.4, 156.7,
-         159.8, 162.2, 165.5, 167.9, 171.3, 173.8,
-         177.3, 179.9, 183.5, 186.2, 189.9, 192.8,
-         196.6, 199.5, 203.5, 206.5, 210.7, 218.1,
-         225.7, 229.1, 233.6, 241.8, 250.3, 254.1,
-         ]
+TONES = (
+    67.0, 69.3, 71.9, 74.4, 77.0, 79.7, 82.5,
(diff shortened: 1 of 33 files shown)


kk7ds:
CC @KC9HI This touches a lot of your drivers, so if you could skim over these changes that might be good. No need to merge this tonight. It pointed out that two drivers were intentionally, maliciously, mutilating the master lists, which is why these should be tuples :)

KC9HI:
OK. I just now started taking a good look that this. There is a problem with CHIRP-next. For example, I loaded a couple of images that are affected by this (Baofeng_UV-9G.img and Baofeng_BF-T1.img) and tried to program a DTCS code of 645. This value is not listed at all for CHIRP next. But if I load the same image into classic CHIRP and try to choose 645, it is available. I will check a few more but I expect pretty much the same results

I don't know if this is something that can be fixed in CHIRP-next or changes have to be made in all of the drivers.

But looking through the diffs, it all looks good to me. I don't think would break anything. But it is hard to tell for sure because everything that it would break, if it was to break it, is already broken. :(

kk7ds:
You mean the extra tones just aren't selectable in the GUI right? That's just because the UI populates the drop-down lists from the master list and not radio features. AFAIK the old UI never did that for tones because we didn't have a list of tones in RadioFeatures at the time. I assume it did for dtcs codes though, so I just need to make the new one do that.

kk7ds:
@KC9HI Wanna give this another shot? Also includes the fix to make `anytone_778uv` STFU during `makesupported` (et al).

KC9HI:
I will begin testing soon. I will let you know.

KC9HI:
I have checked at least 6. All added codes are displayed, all removed codes are not showing up. I'd say it is good, but I can test more after supper if you want me to.

kk7ds:
Nope I think it's good, just wanted to make sure it addressed the problem. When making those lists immutable I was focused on making sure that the import logic honored the modified lists (which it does) but didn't even think about the UI, I guess because I always think about the UI just pulling from the master lists.

PR #492, opened 2023-02-11
Add greek translation 

alichani:
Feature #5291


kk7ds:
I just rebased to get rid of the unnecessary merge commit and added the bug reference to the commit message.

I'll merge this when the tests finish, thanks very much!

PR #496, opened 2023-02-14
Q-Mac HF-90 driver

fthain:
Please review at your convenience.
Some manuals for this radio can be found at https://fccid.io/RK2HF90/


kk7ds (commented):
Other than the selcall/tone mode issue, I just have some minor comments, but otherwise it seems sane.

fthain:
Hi Dan,

Thanks for taking the time to review these patches.

On Tue, 14 Feb 2023, Dan Smith wrote:

> 
> > @@ -94,6 +94,7 @@
>      "DTCS-R",
>      "TSQL-R",
>      "Cross",
> +    "Selcall",
> 
> So, I'd like to avoid doing this. Each of these tone modes are important 
> points of compatibility, in as much as being able to copy a memory from 
> one radio and paste it into another should mean "these two will be able 
> to talk on this channel". I don't think this plus your code is enough to 
> say that. Further, other radios have things they call "selcall" that are 
> not the same (AFAIK) as what it means for this radio. For the same 
> reason that we don't consider "digital code" a tone/squelch mode on 
> D-STAR and YSF radios, I think we should avoid doing this here as well. 
> We also don't expose ANI as a tone mode for radios that support it.
> 

Selcall is another kind of digital coded squelch, so conceptually I think 
it belongs with the other kinds. Note that the various ways of 
automatically opening the squelch gate are all mutually exclusive, and 
that is why it's called a mode.

I think a "tone made" (really a "sub-audible tone mode") is really just 
one of a number of squelch modes, all of which are mutually exclusive, 
that is, choose one or none.

Such a choice needs just one menu -- I can't think of a better UI. We 
could add a "squelch mode" column but would it be an improvement?

An incoming selcall signal also generates an alarm but that's neither here 
nor there (just as enabling roger beep does not imply a different "mode").

Perhaps the UI difficulty is compounded by confusing terminology because a 
mode in this context is not an RF emission mode (like SSB).

I agree that the term "selcall" is ambiguous and that's a problem. And I 
agree that it would be good to copy and paste memories between 
interoperable selcall implementations such as Q-MAC and Codan. Perhaps 
chirp can support this if we add a little detail. What about "CCIR 493-4 
Selcall" for such radios? See also, 
https://en.wikipedia.org/wiki/CCIR_(selcall)

> I would propose making this a `mem.extra` thing for the memories of this 
> radio. It's still editable in bulk via the properties dialog, but it 
> isn't confused with what are actually tone modes.
> 

I don't think that is going to work well because the Skip function is tied 
to the Selcall function. "Skip" means "don't scan" and "scan" means 
"listen for selcall tones on multiple channels in turn". That's why my 
validation routine checks the combination of skip and selcall flags.

> > +                status.msg = "Downloading from radio"
> +                self.status_fn(status)
> +
> +        resp = self._read(3)
> +        if resp != b'\r\n#':
> +            raise errors.RadioError("Incorrect response from radio")
> +
> +        return memmap.MemoryMapBytes(data)
> +
> +    def _upload_chunk(self, addr):
> +        if addr % 16:
> +            raise Exception("Addr 0x%04x not on 16-byte boundary" % addr)
> +
> +        data = self._mmap.get_byte_compatible()[addr : addr + 16]
> +        ihex_rec = to_ihex(addr, 0, list(data[0 : 16]))
> +        cmd = b':' + bytes(bytes(ihex_rec).hex(), 'ascii').upper()
> 
> Why not use b'%04X' here like you do in download?
> 

The _download() case is dealing with 16-bit quantities whereas an IHEX 
record can have any number of bytes. I did consider using a loop and a 
%02X format on each byte but I assumed it would be both slower and more 
verbose. However, I am new to python and I'm probably missing something. 
Can you be more specific please?

> > +        self._memobj = bitwise.parse(HF90_MEM_FORMAT, self._mmap)
> +
> +    def sync_in(self):
> +        try:
> +            self._mmap = self._download(self._memsize)
> +        except errors.RadioError:
> +            raise
> +        except Exception as e:
> +            raise errors.RadioError("Failed to communicate with radio: %s" % e)
> +        self.process_mmap()
> +
> +    def sync_out(self):
> +        highest_nonempty = None
> +        lowest_empty = None
> +        for i in range(self._lower, self._upper + 1):
> +            m = self.get_memory(i)
> 
> You shouldn't assume that `process_mmap()` has been called before a 
> `sync_out()`. If you need to do this, then you should run it here, or 
> guard with `hasattr(self, '_memobj')` but I think it would be better if 
> you can just poke for empty blocks, perhaps while doing the actual 
> upload.
> 
> *...later...*
> 
> Ah yes, see this is why you're failing the clone tests...
> 

I see. I will look into it. BTW, I read all the developer documentation on 
the wiki and in the repo but didn't find any instructions for running the 
tests. Is tox used for that?

> > +        except errors.RadioError:
> +            raise
> +        except Exception as e:
> +            raise errors.RadioError("Failed to communicate with radio: %s" % e)
> +
> +    def get_raw_memory(self, number):
> +        return repr(self._memobj.memory[number])
> +
> +    @classmethod
> +    def get_prompts(cls):
> +        rp = chirp_common.RadioPrompts()
> +        rp.experimental = ("This radio driver is currently under development.\n"
> +                           "There are no known issues with it, but you should "
> +                           "proceed with caution.\n"
> +                           "WARNING -- firmware bugs or chirp bugs could "
> +                           "potentially brick your radio!")
> 
> This is a sensitive topic here. We've never really bricked a radio with 
> CHIRP (that we know of, despite concerns) and I don't want to start now. 
> Are you just covering yourself or is there a real concern with this? If 
> there is, perhaps it's not a good fit for actual inclusion in chirp. If 
> it's not, I think toning down this language would be good.

You're right, the language can be improved. "Chirp bugs" should be taken 
to mean driver bugs. As long as the driver is relatively untested there is 
risk. And I agree that risky code is not ready to be merged. I'm happy to 
delay merging until there has been more testing.

> IMHO, no radio worth its name should allow programming software 
> (especially without serious checksumming) to corrupt it so much that it 
> can't recover.
> 

I agree, and that's why we have firmware updates. The problem with the 
HF-90 is that you have to have an EEPROM programmer to deploy them.

> Perhaps "void your warranty" or "may cause problems" would be more 
> reasonable?

OK, I'll revisit this.

> > +        rp = chirp_common.RadioPrompts()
> +        rp.experimental = ("This radio driver is currently under development.\n"
> +                           "There are no known issues with it, but you should "
> +                           "proceed with caution.\n"
> +                           "WARNING -- firmware bugs or chirp bugs could "
> +                           "potentially brick your radio!")
> +        rp.pre_download = ("Follow these instructions to down/upload:\n"
> +                           "1. Turn off your radio.\n"
> +                           "2. Connect the serial interface cable.\n"
> +                           "3. Turn on your radio.\n"
> +                           "4. Hold down 'CHAN UP' and 'CHAN DOWN' buttons "
> +                           "until 'RS-232' is displayed.\n"
> +                           "5. Click OK to proceed.\n"
> +                           "6. After the process completes, cycle power on the "
> +                           "radio to exit RS-232 mode.\n"
> +                           "NOTE: the HF-90 is compatible with standard PC "
> 
> I also don't think this is the right place for this. This prompt is to 
> tell you how to get the radio into clone mode, etc. The user has already 
> committed to doing it at this point. I also think that mentioning the 
> models of USB adapters that may work is probably best not to codify in 
> the program here, especially since the PL2303 is no longer supported by 
> Prolific, *and* there are arguably more counterfeit clones out there 
> than real ones.
> 
> Can you just put a comment in the experimental or info prompt about the 
> radio being sensitive to serial signaling and save the details for 
> somewhere else?
> 

RS-232 compatibility is important and I think we need to mention it in the 
RadioPrompt text. But I probably wrote too much here as I do see similar 
information in the wiki. And more can be written on HF-90 compatibility. 
Would it be acceptable to link to an HF-90 wiki page from the RadioPrompt 
text?

> > +        except ValueError:
> +            i = 9999
> +        return i
> +
> +    def get_settings(self):
> +        _settings = self._memobj.settings
> +
> +        LOG.debug("get_settings: %s %s" % (type(self), self.MODEL))
> +
> +        grp = RadioSettingGroup("settings", "Settings")
> +
> +        if isinstance(self, EarlyHF90Radio):
> +            rs = RadioSetting("selcall_id", "Selcall ID (use 0 for standard type)",
> +                              RadioSettingValueInteger(0, 9999,
> +                                  self._get_selcall_id(_settings)))
> +            rs.set_apply_callback(self.apply_selcall_id, _settings)
> 
> This looks to be one selcall id per radio, is that right? I was 
> expecting a for loop here that iterated over all the memories based on 
> our discussion before, but I don't see it here. What I'm seeing is that 
> you would program your selcall id into the radio, and then have some 
> channels with selcall enabled and others without, is that right?
> 

Yes. Also, the EarlyHF90Radio enforces the rule that a "standard" radio 
does not have any channels with selcall or scan flags set. This is the way 
that the vendor software works.

fthain:
On Tue, 14 Feb 2023, Dan Smith wrote:

> If you could reply to the individual comments in the future, that'd 
> surely be helpful to keep the discussion in the context of the code 
> where it's easier to browse around. I'm replying via email because I 
> can't really reply to yours in the web UI in the format it's in.
> 

Will do.

> ... Further, you can't copy a memory between two separate radios and 
> have the channel work because the selcall id is not a property of the 
> memory.

I agree that selcall squelch is not quite like tone squelch but there is 
still value in copy and paste despite that.

With regards to selcall scanning, there is value in copy and paste because 
a group of radios needs to agree on the set of frequencies which is to be 
scanned for incoming selcall signalling (and the scan bit requires the 
selcall bit for EarlyHF90Radio).

> Unlike TSQL and 107.2 (for example) which will work between two radios 
> with a single copy and paste. The selcall squelch mode depends on more 
> (and stuff outside the memory) in order to work, which further supports 
> (IMHO) that it's not a match.
> 

I would not be surprised if regulations around emergency frequencies have 
something to say about coded squelch. If so, there would be a need for 
copy and paste to retain a selcall inhibit setting.

> Kenwood radios implement selcall as a per-zone thing (i.e. affecting 
> many memories) and those memories can also have tone squelch modes 
> separate from the per-zone selcall, which are used for various other 
> things, like trigger an automatic ID, a message, or data transmission.
> 

It would appear that Kenwood's selcall implementation is not mutually 
exclusive with squelch modes. But I still think that CCIR 493-4 Selcall 
actually is.

The zones concept you describe is an interesting angle. It sounds a bit 
like the memory banks on my Yaesu FT-60. Perhaps we could have a selcall 
bank, such that memories in that bank would get the selcall bit set...

> > I agree that the term "selcall" is ambiguous and that's a problem. And 
> > I agree that it would be good to copy and paste memories between 
> > interoperable selcall implementations such as Q-MAC and Codan. Perhaps 
> > chirp can support this if we add a little detail. What about "CCIR 
> > 493-4 Selcall" for such radios?
> 
> No, I don't think that would be reasonable, but the other selcall 
> implementations are why I definitely think that just adding "selcall" as 
> a tone mode.
> 
> > I don't think that is going to work well because the Skip function is 
> > tied to the Selcall function. "Skip" means "don't scan" and "scan" 
> > means "listen for selcall tones on multiple channels in turn". That's 
> > why my validation routine checks the combination of skip and selcall 
> > flags.
> 
> I don't think this matters. Either don't set the skip flag in the radio 
> if selcall is enabled, or use a different skip flag (other than the 
> default "S") to imply "don't selcall scan" or something.
> 

I don't understand why the choice of "S" matters... Nonetheless I will try 
to implement your suggestion i.e. use mem.extra for the selcall flag. 

> > 
> > > + cmd = b':' + bytes(bytes(ihex_rec).hex(), 'ascii').upper()
> > > 
> > > Why not use b'%04X' here like you do in download?
> > > 
> > 
> > The _download() case is dealing with 16-bit quantities whereas an IHEX 
> > record can have any number of bytes. I did consider using a loop and a 
> > %02X format on each byte but I assumed it would be both slower and 
> > more verbose. However, I am new to python and I'm probably missing 
> > something. Can you be more specific please?
> 
> IIUC you're returning a tuple of ints from `to_ihex()`, then converting 
> it to a byte string, converting it to a hex string and then parsing 
> *that* back as a byte string with ascii encoding. It's just not very 
> conventional. Something like this would be easier to grok I think:
> 
> ```
> >>> i = (1, 2, 3)
> >>> b''.join(b'%02x' % b for b in i)
> b'010203'
> ```
> 

Thanks for that. I can see that your code is more idiomatic. I still have 
a lot to learn about python.

> > I see. I will look into it. BTW, I read all the developer 
> > documentation on the wiki and in the repo but didn't find any 
> > instructions for running the tests. Is tox used for that?
> 
> Yes, I just (incidentally) added some examples here:
> 
> https://chirp.danplanet.com/projects/chirp/wiki/DevelopersPython3Environment#Running-tests
> 

Thanks.

> > You're right, the language can be improved. "Chirp bugs" should be 
> > taken to mean driver bugs. As long as the driver is relatively 
> > untested there is risk. And I agree that risky code is not ready to be 
> > merged. I'm happy to delay merging until there has been more testing.
> 
> ...and by extension, not say "this may brick your radio" right?
> 

Right. When it's ready to merge, it won't need/have such a warning.

> > > IMHO, no radio worth its name should allow programming software 
> > > (especially without serious checksumming) to corrupt it so much that 
> > > it can't recover.
> > > 
> > 
> > I agree, and that's why we have firmware updates. The problem with the 
> > HF-90 is that you have to have an EEPROM programmer to deploy them.
> 
> Right, but surely programming memories via CHIRP will not actually 
> *brick* the radio, correct?
> 

My driver mimics the latest Q-MAC software. If their software is liable to 
brick radios then so is mine. Here's the actual warning message from 
Qmac.exe v3.0.4:

    This programming software is for use on radios with Export Version 
    firmware (HF-90E or HF-90H) and firmware version 301 or greater.  
    Earlier firmware versions will be corrupted if programmed using this 
    software and will render the HF-90 inoperable.

But I'm happy to report that no radio has been bricked by my chirp driver 
thus far.

The HF-90 Technical Manual seems to say that known bugs in early firmware 
revisions can cause certain radios to be bricked by programming but I 
suspect that those radios went in for repair years ago. For example, 
section 13.3 "Software releases" says that v205 fixed the "no channel 
lockup problem".

I don't have detailed information about firmware changes and I suspect 
many chirp drivers were written under the same circumstances. Before I 
sent this PR, I'd already asked Barrett Communications for detailed 
information about firmware changes but they have not provided it as yet.

By my calculation, chirp wins if it correctly mimics the vendor software. 
That's just a consequence of all of the other benefits it brings: the 
freedom to fix bugs (GPL), cross-platform portability, an active 
community, etc.

> > RS-232 compatibility is important and I think we need to mention it in 
> > the RadioPrompt text. But I probably wrote too much here as I do see 
> > similar information in the wiki. And more can be written on HF-90 
> > compatibility. Would it be acceptable to link to an HF-90 wiki page 
> > from the RadioPrompt text?
> 
> I'd rather not have actual links in these, because they can get used 
> elsewhere and not necessarily in the GUI app (i.e. the CLI). I think we 
> have a hundred other places where people might have the wrong cable, 
> driver, configuration, etc and the users have to find that 
> documentation. But yes, as you note, I think a wiki page is fine and 
> that's what I expect people will find when they google.
> 

Should I refer users to the documentation in the wiki? e.g. "Please refer 
to the CHIRP wiki for information about HF-90 serial port compatibility 
issues."

kk7ds:
>     1. Tx band limits are not the same as the rx band limits. Proposed solution is to implement a Radio.filter_freq() method similar to Radio.filter_name(). This could be overridden by derived radio classes where necessary.

Please rebase on top of this:

https://github.com/kk7ds/chirp/pull/506/commits/55bfe706ded8caba402a011571b53a1c37d875cf

And then add the tx and rx as separate adjacent `valid_bands`.

>     2. test_clone_serial_* unit tests all provide a zeroed mmap, i.e. no channel memories defined. Proposed solution is to define one.

The reason you're failing the test is because you `raise Exception` instead of `errors.RadioError`. The test is asserting that random serial troubles don't raise random errors to the GUI, but instead all get funneled through the expected exception class. The assumption is that if you raise something else, you're not handling those errors gracefully. That said, if you do, you'll just be passing the test without actually hitting some of the other things. So I'll add this, which should resolve it for you:

https://github.com/kk7ds/chirp/pull/506/commits/4fdaba22825c081efe9bd45882ccc79cb6e2af21

chirp/drivers/hf90.py
```diff
@@ -0,0 +1,567 @@
...
+                raise
+
+    def get_memory(self, number):
+        _mem = self._memobj.memory[number - 1]
+
+        mem = chirp_common.Memory()
+        mem.number = number
+
+        _rx = _mem.rx0_7 | (_mem.rx8_15 << 8) | (_mem.rx16_19 << 16)
+        _tx = _mem.tx0_3 | (_mem.tx4_7 << 4) | (_mem.tx8_11 << 8)
```
kk7ds, line 421: Why are you doing this? Aren't these just an `le24` type?

chirp/locale/check_parameters.py
```diff
@@ -10,7 +10,7 @@
 formatter = Formatter()
 
 for name, po in pos.iteritems():
-    print "Testing", name
+    print("Testing", name)
```
kk7ds, line 13: I don't even know what this file was for back when it was relevant, but .. it's not anymore. What made you feel like you had to do this?

Let me just nuke this separately.

chirp/wxui/developer.py
```diff
@@ -347,7 +347,7 @@ def selected(self):
             label = wx.StaticText(self)
             pos = wx.StaticText(self, label='%i bits (%i bytes) at 0x%06x' % (
                     self._memobj.size(),
-                    self._memobj.size() // 8,
+                    int(self._memobj.size() // 8),
```
kk7ds, line 350: Thanks. I've been meaning to track this down. However, this just papers over the issue, because `//` should return an integer of course, and not need to be int-ified like this. I'll push a fix separately and you can drop this.

chirp/locale/check_parameters.py
kk7ds: Okay, I didn't write this so I didn't recognize it. I'll cherry pick this so it's not part of your driver PR.

chirp/drivers/hf90.py
kk7ds: Okay, but `_rx` is really a `ul24` right? I read the first and didn't pay attention to the fact that the tx one was (weirdly) not the same.

fthain: Sorry, I misspoke. Tx and rx are 20 bit quantities (10 nybbles altogether).

fthain: You're right, there is a simpler representation.
`  ul16 rx0_15;
  u8 tx0_3:4, rx16_19:4;
  ul16 tx4_19;
`

kk7ds: Sorry, I see now that the tx and rx bytes are interleaved with each other in memory. I just wasn't seeing it before. I dunno why someone would design something like that, but there are similarly crazy things in other radios, so it must be in the water at conferences or something.

chirp/drivers/hf90.py
```diff
@@ -0,0 +1,567 @@
...
+                if mem.extra["scan"].value or mem.extra["selcall"].value:
+                    msgs.append(chirp_common.ValidationError(
+                        "standard type but scan or selcall enabled"))
+            else:
+                if mem.extra["scan"].value and not mem.extra["selcall"].value:
+                    msgs.append(chirp_common.ValidationError(
+                        "scan enabled but selcall disabled"))
+
+        return msgs
+
```
kk7ds, line 532: The unit failure is related to detection of another radio's image as your driver, because it is the same as your declared `_memsize`. Long ago, we detected models by their memory size, and so that's what the default implementation still does for compatibility. We now append a metadata blob at the end of the file, so radios added after should really opt out of file size detection. So you should add this here:
```
@classmethod
    return False
```
So that your driver only gets detected by the metadata and never by the size or content of the (memory part of the) file. That should clear up that unit fail.

fthain:
Unit tests pass now and all review criticisms have been addressed but I'd like to see some acceptance testing before we proceed so I will try to recruit some more testers.

kk7ds:
Okay, I was going to ask if you were ready because I know you wanted to do some more testing. Can we not leave this open? I won't apply until you're ready, but I'd like to keep all the discussion here with it when it's time to merge.

Also, if you could create a new model request on the chirp site and reference that here when it's ready, that would be helpful.

chirp/drivers/hf90.py
```diff
@@ -0,0 +1,572 @@
...
+# EarlyHF90Radio imitates version 2 of the dealer software whereas
+# LateHF90Radio imitates version 3. It seems both are needed because
+# version 3 warns that it is incompatible with firmware revisions below V301.
+# Unfortunately, the author does not have access to the technical information
+# required to better accomodate the quirks of the various firmware revisions.
+
+@directory.register
+class EarlyHF90Radio(HF90StyleRadio):
+    """Base class for early HF-90 radios"""
+    MODEL = "HF-90 v300 or earlier"
```
kk7ds, line 545: I haven't looked to see if this changed since I looked last, but.. could I convince you to shorten this? We have a couple of bad examples of super long model strings which have caused various issues in the past with rendering, file naming, etc. Could we make this "HF-90 Early" and "HF-90 Late" and put the version numbers in the info prompt? The info prompt is shown as soon as you select the model, so that should be a reasonable time to explain to the user which is which.

Are there more versions of this radio coming out? I was thinking you said this was effectively dead now... perhaps even "HF-90" and "HF-90 Old" ? TBH I really wish we could eliminate spaces from the model strings entirely just so we know we can consider it a non-breaking short string, but the cat is pretty far out of the bag there.

fthain: There are no more versions of this radio coming out. What may change in the future is that technical information may come to light which would provide better criteria to distinguish variants and handle their quirks.

It is possible that we will need to add entries to MODEL_COMPAT at some point. But if that's the price of solid technical information (so far unobtainable) it's cheap, IMHO.

I had trouble finding better model names but they can probably be shortened somehow. What's the character count limit that you'd like to impose?

I would not choose "Late" and "Early" for cosmetic reasons unless those were terms already understood by HF-90 owners. Given that late/early is terminology that we invented, it seems inconsiderate to force users to go and look up what we mean by the new terminology.

Whereas, to distinguish models by firmware revision is user-friendly because the firmware revision level is displayed prominently every time the radio powers up (please see my qrz.com page for photos).

kk7ds: I just want it to be "shortish", I don't have an actual character limit, but when developing the new UI, it was definitely annoying to get all the models to look right on all three platforms. I also kinda want it to be non-breaking (i.e. no spaces) although as I mentioned, that's somewhat of a moot point.

So, we have another attribute called `VARIANT` that was supposed to be for things like "the american version" or "the european version" of a given radio, where it was basically the same implementation, but with different band limits (or something). That lets us make a distinction between the actual "model" of a thing and some qualifying string/text that distinguishes, well, a variant, which means we could make an informed decision when rendering if needed. In the clone box, it just gets appended to the model string, so it would look the same to the users, but we'd have the opportunity to (potentially) only display the actual "HF-90" in the bottom right status bar, on the website listing, etc.

We don't currently distinguish between those when loading files based on the metadata blob, but I think that should change. Part of the reason I didn't do that originally is because it hasn't always been uniformly applied, but it seems like this is a good example of how it would be used. Maybe I could cook up the change in the directory to make sure it works and then we could consider putting the firmware part in the variant field?

fthain: As it happens, the first cut of this driver had one model with two variants. I had to abandon that approach as it didn't work correctly. I don't recall exactly why. But I can try again if there's some benfit. If the only benefit is a shorter model name then I guess I'd like to know how short is short enough, as it may still be more easier to have two models with shorter names and no variants.
I agree we'd have to use metadata to determine which of the two variants was applicable to a given image file (the format is ambiguous meaning that it's easy to create an image file that's valid for both variants).
Regarding model names, strictly speaking there's no HF-90. There's HF-90A, HF-90E, HF-90H and HF-90M but it seems best to lump them all together for now (mainly because the vendor PC software is compatible with several of these). Have you given any thought to a compat map for variant strings (similar to MODEL_COMPAT)?

kk7ds: > As it happens, the first cut of this driver had one model with two variants. I had to abandon that approach as it didn't work correctly. I don't recall exactly why. But I can try again if there's some benfit. If the only benefit is a shorter model name then I guess I'd like to know how short is short enough, as it may still be more easier to have two models with shorter names and no variants.

To be clear, it's the same number of classes either way. It doesn't really look any different for you from what you have here, except that the firmware part of the string is in `VARIANT` instead of `MODEL`. The only thing I can think of that would have been a problem would have been that loading the image file wouldn't select the proper class.

> I agree we'd have to use metadata to determine which of the two variants was applicable to a given image file (the format is ambiguous meaning that it's easy to create an image file that's valid for both variants). Regarding model names, strictly speaking there's no HF-90. There's HF-90A, HF-90E, HF-90H and HF-90M but it seems best to lump them all together for now (mainly because the vendor PC software is compatible with several of these).

Yeah, the reason it matters is so that when someone selects a model/variant (i.e. a class) that we keep that consistent through the download, save, open, upload chain. Not knowing the differences in the A and E models (for example) I can't really comment, but the only reason to distinguish would be when things need to be handled differently (i.e. the cloning or specific rules). If you can determine what firmware version the radio is from the cloning process and/or the image file, I'd say we need only one class and you can just handle the behavioral differences internally. I assume that's the case for the letter differences but for the firmware differences you need the distinct implementations, right? If so, model/variant (and thus separate classes) are the right way.

> Have you given any thought to a compat map for variant strings (similar to MODEL_COMPAT)?

No because it hasn't mattered until now. `MODEL_COMPAT` is just there because we had to change the model name of a few radios and we needed to honor the old format when loading so we didn't make old image files un-openable after the change. We had successfully avoided that for like 14 years until that point, and we definitely should avoid it if at all possible (which is why I want to get this model thing sorted before merge). But, if we *had* to make a variant change, then yes, we'd need to extend that mapping to be a 3-tuple of vendor,model,variant.

fthain: I just noticed your question about detecting a radio's firmware revision level. I don't think it is feasible. It is possible to get the radio to dump any 16 byte region in the address range 0x0000 through 0x1fff but I believe that this doesn't cover all the relevant memory, and figuring out which bytes can be used to fingerprint the radio (model and firmware rev) is not feasible without access to either technical  documentation or a large variety  radios to examine. The bottom line is that the CHIRP user must choose.
Regarding your `directory-variant` branch, I rebased this branch onto that one and the `VARIANT` conversion was straight-forward.
My only concern would be old img files out there, created by old versions of the existing drivers. The trick would be to ensure that old metadata works with the current drivers, in cases where the VARIANT strings for a given driver have changed over time...

chirp/wxui/main.py
```diff
@@ -815,7 +815,7 @@ def _editor_page_changed(self, event):
         self._editors.GetPage(event.GetSelection())
         self._update_window_for_editor()
         radio = self.current_editorset.radio
-        radio_name = '%s %s%s' % (radio.VENDOR, radio.MODEL, radio.VARIANT)
+        radio_name = '%s %s %s' % (radio.VENDOR, radio.MODEL, radio.VARIANT)
```
kk7ds, line 818: So, the problem with this is that some drivers have things like `R` or `E` as their variant, for European or American versions of the radio, which will make this look like `Yaesu FT-2D R` which is.. odd.

I assume you don't need to see the variant information in your case, you just don't want it to look squashed? I would say the best thing for us to do is just drop the variant out of this string entirely as `Yaesu FT-2D` is better than the alternative, especially since most of the drivers don't even put the `R` or `E` in the model string at all. You okay with that?

fthain:
On Thu, 2 Mar 2023, Dan Smith wrote:

> @kk7ds commented on this pull request.
> 
> > @@ -815,7 +815,7 @@ def _editor_page_changed(self, event):
>          self._editors.GetPage(event.GetSelection())
>          self._update_window_for_editor()
>          radio = self.current_editorset.radio
> -        radio_name = '%s %s%s' % (radio.VENDOR, radio.MODEL, radio.VARIANT)
> +        radio_name = '%s %s %s' % (radio.VENDOR, radio.MODEL, radio.VARIANT)
> 
> So, the problem with this is that some drivers have things like `R` or 
> `E` as their variant, for European or American versions of the radio, 
> which will make this look like `Yaesu FT-2D R` which is.. odd.
> 

Yes, but I thought I addressed that in the commit log -- most of the 
VARIANT strings in drivers are not suffixes like "R" but whole words like 
"Upper" or "UHF".

fthain:
On Thu, 2 Mar 2023, Dan Smith wrote:

> @kk7ds commented on this pull request.
> 
> > @@ -815,7 +815,7 @@ def _editor_page_changed(self, event):
>          self._editors.GetPage(event.GetSelection())
>          self._update_window_for_editor()
>          radio = self.current_editorset.radio
> -        radio_name = '%s %s%s' % (radio.VENDOR, radio.MODEL, radio.VARIANT)
> +        radio_name = '%s %s %s' % (radio.VENDOR, radio.MODEL, radio.VARIANT)
> 
> So, the problem with this is that some drivers have things like `R` or 
> `E` as their variant, for European or American versions of the radio, 
> which will make this look like `Yaesu FT-2D R` which is.. odd.
> 

`Yaesu FT-2D R` makes more sense to me than `FT-7100MVHF Band`.

kk7ds:
> Yes, but I thought I addressed that in the commit log -- most of the VARIANT strings in drivers are not suffixes like "R" but whole words like "Upper" or "UHF".

Yeah, the difference is that those variants are sub-devices of the main radio itself, so they don't get displayed in the status bar. Obscure, I know...

chirp/wxui/main.py
fthain: >the best thing for us to do is just drop the variant out of this string entirely
>

Removing the variant from the status bar would be the sort of compromise where everyone suffers equally. A much better solution is just do nothing! That is,  drop the patch and wait for someone to find a better solution.

kk7ds:
When resolve the conflict, please put this in the commit message for the driver commit:
```
Fixes: #10428
```
So that it gets linked properly. Then I'll merge.

PR #499, opened 2023-02-15
Fix offset problem on Asia version FT-25R/65R

HailerB:
Fixes #9713 and #10286

Asia region FT-25R / FT-65R radios treat repeater offsets differently than their US siblings. This patch adds capabilities to detect such radios and to correct the offset accordingly.

### DIFF
--- a/chirp/drivers/ft4.py
+++ b/chirp/drivers/ft4.py
@@ -241,7 +241,7 @@ def variable_len_resp(pipe):
             if i > toolong:
                 LOG.debug("Response too long. got" + util.hexprint(response))
                 raise errors.RadioError("Response from radio too long.")
-    return(response)
+    return response
 
 
 def sendcmd(pipe, cmd, response_len):
@@ -288,7 +288,7 @@ def enter_clonemode(radio):
             try:
                 if b"QX" == sendcmd(radio.pipe, b"PROGRAM", 2):
                     return
-            except:
+            except Exception:
                 continue
         sendcmd(radio.pipe, b"END", 0)
     raise errors.RadioError("expected QX from radio.")
@@ -477,6 +477,7 @@ class YaesuSC35GenericBankModel(chirp_common.BankModel):
             if retrieve_bit(self._radio._memobj.bankmask[bank.index], memndx):
                 banks.append(bank)
         return banks
+
 
 # the values in these lists must also be in the canonical UI list
 # we can re-arrange the order, and we don't need to have all
@@ -951,7 +952,7 @@ class YaesuSC35GenericRadio(chirp_common.CloneModeRadio,
                     setattr(_settings, name, value)
 
                 LOG.debug("Setting %s: %s" % (name, value))
-            except:
+            except Exception:
                 LOG.debug(element.get_name())
                 raise
 
@@ -1038,7 +1039,7 @@ class YaesuSC35GenericRadio(chirp_common.CloneModeRadio,
                     ndx = x[1].index(memref)
                     array = x[0]
                     break
-                except:
+                except Exception:
                     num += len(x[1])
             if array is None:
                 LOG.debug("unknown Special %s", memref)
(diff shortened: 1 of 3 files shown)


kk7ds:
Hey Bernard, thanks for looking into this! Can you put the "Fixes #xxx" stuff in the commit message itself? Redmine won't see anything about the PR, just the commits. The following is probably the git-fu you need:
```
$ git commit --amend
$ git push -f
```
That should amend your existing commit (let you change the message) and force-push to update this PR.

Thanks!

HailerB:
I'm not sure whether this came through correctly.

kk7ds:
Nope, you've got a merge commit in there now. I'll fix it for you.

HailerB:
Thanks. New Git client, new workflow, new repo - I hope I'll get this straight next time.

kk7ds:
Also, in the future, you should create a branch to do your work in and create the PR from that branch. Since you used 'master' I have to do a few extra dance steps to check it out, adjust and push for you.

Just git workflow stuff to get used to :)

Thanks!

HailerB:
Okay, makes a lot of sense to me.

chirp/drivers/ft4.py
```diff
@@ -308,16 +308,26 @@ def startcomms(radio, way):
     progressbar.max = radio.numblocks
     enter_clonemode(radio)
     id_response = sendcmd(radio.pipe, b'\x02', None)
+
+    # The last byte of the ID contains info about regional differences
+    radio.subtype = id_response[len(id_response) - 1]
```
kk7ds, line 313: Hmm, this will work for the case where you download the image, make changes and upload, but it won't work if I save the image and then re-open it right? Won't we be using the default to calculate the multiplier in that case?

I'm guessing that we need to have another radio class for the asia variant, and abort the clone here if the class they chose doesn't match the radio's ID string. Then we'll always know which multiplier to use based on the class, and we'll make sure we're talking to the right radio whenever we do.

HailerB:
Darn, I hadn't considered that. You're right. I wanted to avoid adding an extra radio and do it automatically. - Is the ID string somewhere encoded in the blob at the end of an IMG file?

kk7ds:
Yeah, chirp will be able to tell them apart if the vendor/model are different. It doesn't currently consider the variant as a strong differentiator, otherwise I'd say that's a reasonable approach here. There's also a way to persist info about the radio other than the vendor/model in the metadata blob. It's very recent, but maybe it's an appropriate thing to use here. If you stuff things into `radio._metadata` (which is a dict) they should be saved and restored in the `.img` file for you. Use some reasonable namespacing for the key, like `ft4_multiplier` or `ft4_identbyte`.

HailerB:
Ok, I'm going to rework this patch and add FT-25R (Asia) and FT-65R (Asia) radios to the selection instead.

kk7ds:
As opposed to the solution I just offered? I definitely think that putting " (Asia)" in the `MODEL` is not the best plan. It muddies things up a lot and we can use the metadata to store our flag I think.

HailerB:
Ah, hadn't seen your last comment. I'll go down the avenue you're suggesting. Will take a while, I need to go explore.

kk7ds:
I can also try to do that for you and pass it back to you for verification, just let me know.

HailerB:
I might take you up on that, but first let me try what I can do this coming week, I'll be on a "shutdown" (it never is no work). I'll contact you if I get stuck. Thanks!

kk7ds:
Yep, I'll try to get to it this week.

kk7ds:
I forgot I had written a clone test for this driver. I decided to make sure that validated the new behavior, and I'm glad I did because I apparently forgot how my own metadata stuff worked. With this now validated, I think I'm comfortable merging this without other confirmation (which seems unlikely anyway given the response on the bug).

PR #515, opened 2023-02-24
Trivial typo in FT-60 settings menu

fthain:
See also page 16 of FT-60R/E Operating Manual.


kk7ds:
Thanks. I'll hold off on merging this until there's something else meaty to go along with it so we don't build just for the typo.

PR #517, opened 2023-02-28
baofeng_wp970i.py: limit check the VFO A/B TXP settings - fixes #10403

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/baofeng_wp970i.py
+++ b/chirp/drivers/baofeng_wp970i.py
@@ -725,16 +725,24 @@ class WP970I(baofeng_common.BaofengCommonHT):
         rs.set_apply_callback(apply_offset, _mem.vfo.b)
         work.append(rs)
 
+        if _mem.vfo.a.txpower3 > 0x02:
+            val = 0x00
+        else:
+            val = _mem.vfo.a.txpower3
         rs = RadioSetting("vfo.a.txpower3", "VFO A Power",
                           RadioSettingValueList(
                               LIST_TXPOWER,
-                              LIST_TXPOWER[_mem.vfo.a.txpower3]))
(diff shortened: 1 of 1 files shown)


chirp/drivers/baofeng_wp970i.py
```diff
@@ -725,16 +725,24 @@ def apply_offset(setting, obj):
...
         work.append(rs)
 
+        if _mem.vfo.a.txpower3 > 0x02:
+            val = 0x00
+        else:
+            val = _mem.vfo.a.txpower3
         rs = RadioSetting("vfo.a.txpower3", "VFO A Power",
                           RadioSettingValueList(
                               LIST_TXPOWER,
-                              LIST_TXPOWER[_mem.vfo.a.txpower3]))
```
kk7ds, line 731: One way to do this with less code (if you're willing to compromise on the value chosen) is to just do:
```
min(_mem.vfo.a.txpower3, 0x02)
```
That won't default to zero but it will choose whatever power 0x02 is, if it's over range. It's pretty nested here so it might be worse for readability than just doing what you did, but it is a fast way to just write the code to be defensive.

No need to change, just an idea for next time.

kk7ds:
Thanks for doing this!

chirp/drivers/baofeng_wp970i.py
KC9HI: Sweet. I think I will go back through the new driver that I am working on and use this where appropriate. Thank you.

KC9HI: I was able to update my new driver in just a few minutes. This was very easy and shouldn't be any problem to do as I code in the future.

PR #520, opened 2023-03-01
BF-A58S: Update to support 2 TX power levels - fixes #10404

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/baofeng_wp970i.py
+++ b/chirp/drivers/baofeng_wp970i.py
@@ -68,6 +68,9 @@ LIST_TXPOWER = ["High", "Mid", "Low"]
 LIST_VOICE = ["Off", "English", "Chinese"]
 LIST_WORKMODE = ["Frequency", "Channel"]
 
+TXP_CHOICES = ["High", "Low"]
+TXP_VALUES = [0x00, 0x02]
+
 
 def model_match(cls, data):
     """Match the opened/downloaded image to the correct version"""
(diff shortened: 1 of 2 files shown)


chirp/drivers/baofeng_wp970i.py
```diff
@@ -940,6 +966,9 @@ class BFA58S(WP970I):
     VENDOR = "Baofeng"
     MODEL = "BF-A58S"
     LENGTH_NAME = 7
+    POWER_LEVELS = [chirp_common.PowerLevel("High", watts=5.00),
+                    chirp_common.PowerLevel("---", watts=1.01),
```
kk7ds, line 970: Erm, does this get exposed in this driver's valid_power_levels? If so, I don't think we should do this.

KC9HI: OK. I though that was a trick you told me about before. I must have been thinking about something else. Is there a way to suppress the middle choice or a better way to do this? It hasn't been a problem for the other Baofengs because the the radios with 2 TX power levels use a different memory location that used for radios with 3 TX power levels. For some reason in this radio they use 0x00 for High and 0x02 for Low.

kk7ds: Well, if you need to index into the list, making sure it always has enough entries is good, but not if some of those are invalid. Especially if there are enough bits to possibly mean this could get returned if memory was corrupted, that would be counter-productive. I haven't read through all of this driver, but why does this need to be three elements long? Are low and high 0 and 2 in memory? If it's possible that corrupted memory could select the middle entry, then I think it needs to be valid (perhaps just a duplicated Low entry). But, you'd want to filter that out in `get_features()`...

KC9HI: I'm good with setting to be High/Low/Low. I think there is a driver I can look at that has High/Mid/Low and Mid in the radio doesn't work so the driver forces it to High (or maybe Low, I don't remember). I can do that tomorrow.

PR #521, opened 2023-03-01
Adds IC-V86 test image and fix failing tests. #7231

karvanitis:
The test image in the orginal new model issue was never added;

fixes test failing due to out of index array, and 
removes unnecessary MUNCH_CLONE_RESP which prevented radio i/o.

# CHIRP PR Checklist

* All tests must be passing.


kk7ds:
Thanks Kosta. In the future please use a topic branch instead of `master` as it makes it much easier for me to push changes to your PR if needed.

I think this really needs a (new) bug to account for what looks like an obvious fix that prevents the radio from working at all. We also need to clean up the `MUNCH_CLONE_RESP` as this was the last driver still "using" that ill-fated flag. Normally I'd push both up here, but I'll go ahead and merge this as-is and push that change after.

I assume you could mark the V86 as tested by you in the matrix right? Perhaps another follow-on.

PR #522, opened 2023-03-01
Fix a number of bugs

kk7ds:
- Ensure browser panels are re-fit on selection
- Allow loaded modules to be reloaded
- Disable dev radio menu items when appropriate
- Remove MUNCH_CLONE_RESP from icf

### DIFF
--- a/chirp/detect.py
+++ b/chirp/detect.py
@@ -24,7 +24,6 @@ LOG = logging.getLogger(__name__)
 
 class DetectorRadio(chirp_common.Radio):
     """Minimal radio for model detection"""
-    MUNCH_CLONE_RESP = False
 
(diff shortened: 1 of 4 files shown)


kk7ds:
@KC9HI This should fix the scroll bar issue, the always-active reload menu items, as well as the reloading loaded modules issue. If you could try the loading and reloading stuff with tomorrow's installer build to make sure this didn't break any of that on windows (and solved the problem), I'd appreciate it :)

KC9HI:
Scroll bar issue - fixed
Reload Driver - Works {and I think I finally realize what it is supposed to do - I was thinking it did something different}
Editing/saving image with 'bad' setting using Browser - Works

PR #529, opened 2023-03-04
Add updated Turkish translation

TA1AUB:
Add updated Turkish translation


kk7ds:
Thanks, I will merge this as soon as we have some other functional change pending in the tree, just to avoid generating a build only for a translation update.

PR #534, opened 2023-03-07
Move alinco delays to earlier in the loop

kk7ds:
The alinco radios are very sensitive to replying too quickly after
a response. We currently delay late in the loop to avoid requesting
blocks too quickly. However it seems like -next (probably due to
python 3.10 perf improvements) is too quick on the first block
request after the ident response. So this moves the delay to the top
of the loop so we get it after the ident, and after each block.

Fixes #10414


kk7ds:
Azure mirror for 20.04 is busted, passes on the other distro, so clearly okay.

PR #537, opened 2023-03-09
Update Turkish language and add missing message

TA1AUB:
Update Turkish language
Add missing driver messages


TA1AUB:
What can I do to fix the error?

kk7ds:
That check is failing because your translation file(s) have strings in it that the regular locale build won't find and thus it looks like a problematic delta. Most of the strings in `chirp/drivers` are wrapping the `dedent()` call which is why `xgettext` isn't finding them (I didn't realize until now). What did you use to generate the translation files for all those strings such that it was able to find its way around them?

Unfortunately, I think we'll need to fix all of those to not be like this:
```
_(dedent("foo"))
```
and instead be like this:
```
dedent(_("foo"))
```

PR #539, opened 2023-03-10
Updata turkish language

TA1AUB:


TA1AUB:
I can't fix the problem. Please fix it for me.

kk7ds:
It's not trivial to fix because these strings aren't currently being extracted from the drivers because the marking is being done on a function and not on the string literals. I'm working on fixing that in #541 so you'll need to rebase on that once it's working and merged. I hope you'll be able to `msgmerge` what you have done here into that, but I'm not sure because of the way the whitespace is changing.

PR #545, opened 2023-03-11
Update turkish language

TA1AUB:


TA1AUB:
Can you please edit the file like this?

kk7ds:
Yeah I think it's okay now, I should make the check not so concerned about the file location lines. I'll pull this down and rebuild to check. Thanks for your work on this :)

kk7ds:
Can you confirm this looks okay? The build will create multiple strings to break at 80 columns which is another reason the check complains. I'm guessing the editor you are using rewraps them or something. Not a problem, it's easy to fix.

TA1AUB:
Everything looks okay.

PR #551, opened 2023-03-14
Update SHX8800 in GA-510 driver

lunapiercook:
# - remove language option from 8800 family (unused) 
# - correct spelling of Cignus XTR-5

# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/ga510.py
```diff
@@ -1074,12 +1084,6 @@ class TDH6Radio(RadioddityGA510Radio):
     VENDOR = "TIDRADIO"
     MODEL = "TD-H6"
 
-    def get_features(self):
-        rf = super().get_features()
-        rf.valid_bands = [(136000000, 174000000),
-                          (400000000, 520000000)]
-        return rf
-
```
kk7ds, line 1082: Why are you removing this? It was *just* added recently and removing it will regress that bug.

chirp/drivers/ga510.py
```diff
@@ -1102,9 +1106,9 @@ class RadioddityGS5BRadio(Senhaix8800Radio):
 
 
 @directory.register
-class SignusXTR5Radio(Senhaix8800Radio):
-    """Signus XTR-5"""
-    VENDOR = "Signus"
+class CignusXTR5Radio(Senhaix8800Radio):
+    """Cignus XTR-5"""
+    VENDOR = "Cignus"
```
kk7ds, line 1111: You can't just change a vendor/model like this as you will break compatibility with everyone's existing images for this radio. Was this really a typo (if so, that's frustrating) or does this radio exist with both spellings? Either way, we can't change it without accounting for the change in some way to avoid breaking users. There's a compatibility list in the loader, which we can use if we have to, but if it's spelled the old way in some places, we should just leave it.

chirp/drivers/ga510.py
```diff
@@ -533,6 +531,7 @@ def get_memory(self, num):
                                        self._decode_tone(_mem.rxtone))
         try:
             mem.extra = self._get_extra(_mem)
+
```
kk7ds, line 534: Please don't add random whitespace damage.

chirp/drivers/ga510.py
```diff
@@ -635,7 +634,15 @@ def get_settings(self):
...
 
         radioddity_settings = {
-            'language': ['English', 'Chinese'],
+            'savemode': ['Off', 'Mode 1', 'Mode 2', 'Mode 3'],
+            'cha_disp': ['CH+Name', 'CH+Freq'],
+            'chb_disp': ['CH+Name', 'CH+Freq'],
+            'txundertdr': ['Off', 'Band A', 'Band B'],
+            'rptnoiseclr': ['Off'] + ['%i' % i for i in range(100, 1001, 100)],
+            'rptnoisedet': ['Off'] + ['%i' % i for i in range(100, 1001, 100)],
+        }
```
kk7ds, line 643: Why are you duplicating all of these here? If they're common to both radios, they should go in `choice_settings` to avoid the repetition, no?

kk7ds (changes requested):
Why are you changing the test image? Please don't do that.

lunapiercook:
Hi Dan,
   
  
    
   
  
   Everything is noted, I'll answer your questions though ...
   
   
   @kk7ds requested changes on this pull request. 
   Why are you changing the test image? Please don't do that. 
   
  
   I assumed (wrongly, of course) that changing the driver required also changing the image. I'll not do that again.
   
  
    
   In chirp/drivers/ga510.py: 
   > @@ -1074,12 +1084,6 @@ class TDH6Radio(RadioddityGA510Radio):
     VENDOR = "TIDRADIO"
     MODEL = "TD-H6"
 
-    def get_features(self):
-        rf = super().get_features()
-        rf.valid_bands = [(136000000, 174000000),
-                          (400000000, 520000000)]
-        return rf
-
 
   Why are you removing this? It was just added recently and removing it will regress that bug. 
   
     
    
   
  
   All I can think of is that I somehow loaded an old copy of the module when transferring to a new laptop. I definitely need to ensure the module in-work is the latest.
   
  
    
   In chirp/drivers/ga510.py: 
   > @@ -1102,9 +1106,9 @@ class RadioddityGS5BRadio(Senhaix8800Radio):
 
 
 @directory.register
-class SignusXTR5Radio(Senhaix8800Radio):
-    """Signus XTR-5"""
-    VENDOR = "Signus"
+class CignusXTR5Radio(Senhaix8800Radio):
+    """Cignus XTR-5"""
+    VENDOR = "Cignus"
 
   You can't just change a vendor/model like this as you will break compatibility with everyone's existing images for this radio. Was this really a typo (if so, that's frustrating) or does this radio exist with both spellings? Either way, we can't change it without accounting for the change in some way to avoid breaking users. There's a compatibility list in the loader, which we can use if we have to, but if it's spelled the old way in some places, we should just leave it. 
   
     
    
   
  
   This was a model I added as part of the SenhaiX 8800 family, and misspelled myself. Again, I assumed (wrongly, of course) that this would be a simple replacement spelling.
   
  
    
   In chirp/drivers/ga510.py: 
   > @@ -533,6 +531,7 @@ def get_memory(self, num):
                                        self._decode_tone(_mem.rxtone))
         try:
             mem.extra = self._get_extra(_mem)
+
 
   Please don't add random whitespace damage. 
   
     
    
   
  
   Not sure where that came from, could be from using an older module file.
   
  
    
   In chirp/drivers/ga510.py: 
   > @@ -635,7 +634,15 @@ def get_settings(self):
         dtmf = RadioSettingGroup('dtmf', 'DTMF')
 
         radioddity_settings = {
-            'language': ['English', 'Chinese'],
+            'savemode': ['Off', 'Mode 1', 'Mode 2', 'Mode 3'],
+            'cha_disp': ['CH+Name', 'CH+Freq'],
+            'chb_disp': ['CH+Name', 'CH+Freq'],
+            'txundertdr': ['Off', 'Band A', 'Band B'],
+            'rptnoiseclr': ['Off'] + ['%i' % i for i in range(100, 1001, 100)],
+            'rptnoisedet': ['Off'] + ['%i' % i for i in range(100, 1001, 100)],
+        }
 
   Why are you duplicating all of these here? If they're common to both radios, they should go in choice_settings to avoid the repetition, no? 
   
  
   I'll certainly get this rearranged so it's assembled in a more logical manner.
   
  
    
   
  
   Thanks for the notes, Dan, they're greatly appreciated!!
   
  
    
   
  
   Dave

PR #554, opened 2023-03-15
Baofeng BF-T1: Support new FM frequency storage method - fixes #10445

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/bf_t1.py
+++ b/chirp/drivers/bf_t1.py
@@ -465,6 +465,7 @@ class BFT1(chirp_common.CloneModeRadio, chirp_common.ExperimentalRadio):
     _upper = 20
     _magic = BFT1_magic
     _id = BFT1_ident
+    _bw_shift = False
 
     @classmethod
     def get_prompts(cls):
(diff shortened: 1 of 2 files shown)


chirp/drivers/bf_t1.py
```diff
@@ -822,14 +823,30 @@ def get_settings(self):
...
+            LOG.debug("Setting fm_vfo = %s" % (value))
+            if self._bw_shift:
+                value = ((value & 0x00FF) << 8) | ((value & 0xFF00) >> 8)
+            setattr(obj, setting.get_name(), value)
+
+        # broadcast FM setting
+        value = _settings.fm_vfo
+        value_shifted = ((value & 0x00FF) << 8) | ((value & 0xFF00) >> 8)
+        if value_shifted <= 108.0 * 10 - 650:
+            # storage method 3 (discovered 2022)
```
kk7ds, line 836: Sorry for the delay here. I want to write some tests for this to make sure we're getting and setting all the various formats and that they're not getting confused.

You note that this is the third method here, but I only see this one and the original one (L840). What is the second method?

kk7ds: Ah, okay I thought this was the same as the recent one but a new method. Okay, so that makes sense. I'll extend my tests to include the middle method (on those radios) and include them in the coverage. I'll push something up here hopefully in a bit and you can check my work.

I think it'd be good to have this test coverage so we don't break method 3 when method 33 is discovered.

kk7ds: Hmm, the UV-5R image in tree seems to not load fm presets for me... Perhaps it's already broken for the format it uses?

KC9HI: I displays in legacy CHIRP. Perhaps I broke it?

kk7ds: Right, I think it's taking this path for the image in tree:
```
        else:
            # unknown (undiscovered method or no FM chip?)
            preset = False
```

This is why I think it's probably good to get some tests for this specifically on these models in the tree to make sure we're covering each of the possible cases.

I guess I can push up the tests for the BF-T1 and we can treat the UV-5R issue separately, unless you can spot the problem.

KC9HI: I can look it. It must that the 3rd type that I didn't have an image for.

KC9HI: If I go back to the last build in February, it works. I will figure it out an submit a patch.

kk7ds: Okay, I'll push the test for the BF-T1 on here and then I will extend it for the UV-5R once we have it fixed.

tests/unit/test_baofeng_fmradio.py
```diff
@@ -0,0 +1,166 @@
...
+                # We should not be able to set this out-of-range value
+                self.assertRaises(settings.InvalidValueError,
+                                  self.set_fmvfo, dial)
+
+                # Can't physically set this in memory
+                if dial <= 65:
+                    continue
+
+                # Set it anyway in memory directly
+                print(dial)
```
kk7ds, line 140: Oops, just realized I left this in.

chirp/drivers/bf_t1.py
kk7ds: Okay, so method three in the BF-T1 is this:
```
        if value_shifted <= 108.0 * 10 - 650:
            # storage method 3 (discovered 2022)
```
But in the UV-5R it's this:
```
        if value_shifted >= 65.0 * 10 and value_shifted <= 108.0 * 10:
            # storage method 3 (discovered 2022)
```

So a value of 0x5E01 (which should be 100.0) for the BF-T1 doesn't work for the UV-5R. Should the UV-5R be subtracting 650 there as well? Or are these actually different methods, both byte-swapped, but not both scaled by 650?

KC9HI: OK. I think I have the numbering mixed up. For some reason I juggled the order up when I did this on the 1st of March. Probably because I didn't understand what each was doing back then. I understand better tonight.

The way I understand it right now is up until March 1 there were 2 methods

M1
M2

Then on March there was a 3rd method added to uv-5r.py

shifted_M2

The BF-T1's original support was

M1

and now the new (4th) method is

shifted_M1

My head is spinning...

kk7ds: Okay, maybe we should standardize what we call these and also unify the functions in a single place. Would `baofeng_common.py` be a reasonable place?

I guess I should go ahead and merge this and the other one you have and I'll plan to try to finish this after. I'll push to remove the errant print first.

KC9HI: Yes, that sounds like a reasonable place.

I've got to shut down soon. I've been staying up too late every night doing this and with the recent 12+ hours work session for the btech.py driver limit checking and the time change on top of it, it became apparent that I was over doing it.

PR #555, opened 2023-03-16
uv5r.py: allow full-band programming support for UV-5G - fixes #10444

KC9HI:
The UV-5G (aka UV-5X) GMRS radios are preset from the factory (and after a RESET) with 30 GMRS channels and 11 NOAA WX radio channels. The programmed channels are not firmware locked from being changed or erased, TX is not firmware blocked for unsanctioned frequencies. In respect to channel programming, they are essentially the same as any common UV-5R (136-174 MHz & 400-520 MHz and TX/RX on all 128 channels).

This patch gives CHIRP the ability to match the radio's factory provided programming capability.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/uv5r.py
+++ b/chirp/drivers/uv5r.py
@@ -748,7 +748,6 @@ class BaofengUV5R(chirp_common.CloneModeRadio):
     _uhf_range = (400000000, 520000000)
     _aux_block = True
     _tri_power = False
-    _gmrs = False
     _bw_shift = False
     _mem_params = (0x1828  # poweron_msg offset
                    )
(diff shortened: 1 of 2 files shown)


KC9HI:
The driver is correct. The problem is that the image in the tree is not populated with a frequency value. It contains 0xFFFF so up until March 2, the driver has shown the specified default frequency (76.0 MHz) when a valid one was not provided.

It needs have a value that is somewhere between 0x0000 (65.0 MHz) and 0x1AE (108.0 MHz) inclusive.

kk7ds:
Ah, okay. So I'll test for that case as well and then the valid values.

KC9HI:
Oh. I replied to the wrong RP. LOL

PR #556, opened 2023-03-17
Unify baofeng fmradio decoding

kk7ds:
- Improvements to RadioSettings
- Unify fmradio decoding for baofengs

### DIFF
--- a/chirp/drivers/baofeng_common.py
+++ b/chirp/drivers/baofeng_common.py
@@ -16,6 +16,7 @@
 
 """common functions for Baofeng (or similar) handheld radios"""
 
+from collections import namedtuple
 import time
 import struct
 import logging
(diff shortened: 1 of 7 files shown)


kk7ds:
@KC9HI Man, what a nightmare.

So I think I've unified all the decoding and noted which methods we've seen on which models. However, in the process I think I've found a problem, in that there are some overlapping values (more on the UV-5R of course). Those are listed in the test file in the `OVERLAPS` list. Unfortunately, that means that there are cases where we won't be able to tell which method should be used for a given radio without some other sort of flag or version in the image to go on.

Not sure there's much we can do about it, but at least the overlapping values are listed there so that if someone complains about 65.1 being decoded as 90.6, we'll know it's a collision. I thought maybe all the collisions differed by even/odd steps such that you could be pretty sure the two won't exist in the same market, but there are some even/even collisions.

Anyway, let me know what you think of the encode/decode logic and if it all makes sense.

KC9HI:
I regret adding that setting to CHIRP. It isn't a "preset" like I called it. It is a VFO. The user can just key the frequency into the radio, or in case of a radio like the BF-T1 with no keypad, start the scan to find an active frequency.

Let me get to a place on my other project and I will load this up.

kk7ds:
Ack, well, TBH I don't really know why this is important to be able to set from the software. I've literally never (not once) wanted to control my VFOs (ham band or fm radio) from software. Maybe it would be better to just nuke them from both drivers entirely? Even though I just spent many hours untangling those, I'd be in favor ;)

Anyway, no rush at all, this is just latent cleanup really.

KC9HI:
What concerns me now is that when some guy with a fleet of Baofeng radios goes to program all with the same image, if the image used was from a radio that uses method 3, the frequency will be wrong or won't display in the radios that use method 1 or method 2. Changing the storage method wasn't an issue for Baofeng because their software doesn't have support loading/saving this setting. Another issue is that due to supply chain issues, a lot of these radios will be shipping without the FM chip and will not support that feature. I thought about taking time so that when a valid frequency wasn't detected, the "FM Settings" menu didn't even show up in CHIRP. But then I thought how much time to I really want to spend on this setting that I shouldn't have programmed to begin with. :(

kk7ds:
Well, I don't think there's anything we can do about the "using a method 3 image on a method 2 radio" issue regardless. But maybe you just mean that's an unfortunate consequence of them switching? It's definitely bad.

Sounds like you're leaning towards just ripping it out entirely and if so, I'm totes on board :)

PR #557, opened 2023-03-18
Fix editing entries for TH-D74

arodland:
Editing a memory in a D74 file in the GUI currently fails with a message "no such attribute 'dv_urcall'", because every memory is created as a DVMemory, but not every memory actually has the DV fields, and this blows up in some serialization code.

Do like other drivers do and only create a DVMemory for DV memories. Some of the number/extd_number was assigning to mem before we even called _get_raw_memory (and therefore before we could tell whether the mode is DV or not), so it had to be reorganized slightly.

Tested with my own D74.

### DIFF
--- a/chirp/drivers/thd74.py
+++ b/chirp/drivers/thd74.py
@@ -379,27 +379,34 @@ class THD74Radio(chirp_common.CloneModeRadio):
         return self._memobj.memgroups[number // 6].memories[number % 6]
 
     def get_memory(self, number):
-        mem = chirp_common.DVMemory()
-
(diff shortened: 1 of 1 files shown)


kk7ds:
Thanks, I dunno why I did it that way. Probably ten years between the D74 and the previous D-STAR radio I added :)

kk7ds:
In the future, it'd be nice to have a bug in the chirp tracker for stuff (per the checklist which you don't see if you use the CLI to push a PR). For user-visible fixed, that's where users are directed to find things that have already been fixed/reported.

PR #561, opened 2023-03-18
Use a PEP440/PEP508 compatible version to avoid errors on recent versions of setuptools

mrene:
[Setuptools 67.0.0](https://setuptools.pypa.io/en/stable/history.html#v67-0-0) introduces some breaking changes which prevent chirp from installing.

Because chirp uses `pydev3` as a version placeholder, this aborts the installation process because the version is invalid.

Opening this as a PR with a dummy `1.0.0` version since issues are disabled on the project (it fixes the build, but I'm not sure how you want to handle versioning)

### DIFF
--- a/chirp/__init__.py
+++ b/chirp/__init__.py
@@ -17,7 +17,7 @@ import os
 import sys
 from glob import glob
 
-CHIRP_VERSION = "py3dev"
+CHIRP_VERSION = "1.0.0"
 
 module_dir = os.path.dirname(sys.modules["chirp"].__file__)
 __all__ = []


kk7ds:
Fair warning, this is a hotbutton issue for me. I don't care about PEP440 nor do I care that python packaging people think they should be able to dictate my version numbers. The reason it's set to 0 in the `setup.py` file is so that install works directly from the git tree. The build process of stamping out tarballs puts just the date part of the build into the setup.py when it does those, so that they too are compliant.

If you're hitting some new thing, please point it out and we can try to work around it, but nothing should be looking at the `CHIRP_VERSION` other than chirp itself anymore.

mrene:
Oh appoplogies, I didn't see that fix setting the version to 0. I don't particularly care about the versioning issue, I was just hitting a failure when trying to run the NixOS package for chirp. 

Updating the package to the latest version builds fine.

PR #568, opened 2023-03-27
UV-5R read_block: read requested size from the pipe, not hardcoded 0x40

acabey:
Small clean-up I noticed while working through the UV-5R code. The `_read_block` function is currently only ever called with a size value of 0x40, so functionally this will not make any difference, but it may prevent some future frustration and debugging if it is ever called with a different value. I may also just misunderstand the protocol here if the chunk size must always be 0x40 (or at most 0x40); just let me know if that is the case.


kk7ds:
cc @KC9HI

KC9HI:
This looks good to me. Good catch.

PR #573, opened 2023-03-28
Fix memory dialog initialization of power level

kk7ds:
For radios with per-memory power levels, the memory property dialog
was not determining the proper "current" index and defaulting to
index=0. This fixes that by str()'ing the choices before testing for
membership of the str()'d current value.

Fixes: 10487


kk7ds:
cc @mrtjr1

PR #578, opened 2023-04-03
Add Radtel RT-470 - fixes #10354

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1070 @@
...
+    # check if we had EXITO
+    if exito is False:
+        msg = "The radio did not accept program mode after five tries.\n"
+        msg += "Check you interface cable and power cycle your radio."
+        raise errors.RadioError(msg)
+
+    try:
+        serial.write(b"F")
+        ident = serial.read(8)
+    except errors.RadioError:
```
kk7ds, line 281: These serial calls can't ever raise the chirp-specific `errors.RadioError` so this will never happen. This should probably be `except Exception` to cache OS errors reading or writing to the pipe.

chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1070 @@
...
+    if ident not in radio._fingerprint:
+        LOG.debug(util.hexprint(ident))
+        raise errors.RadioError("Radio returned unknown identification string")
+
+
+def _exit_programming_mode(radio):
+    serial = radio.pipe
+    try:
+        serial.write(b"E")
+    except errors.RadioError:
```
kk7ds, line 293: Same here.

chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1070 @@
...
+    LOG.debug("Reading block %04x..." % (block_addr))
+
+    try:
+        serial.write(cmd)
+        response = serial.read(4 + block_size)
+        if response[:4] != expectedresponse:
+            raise Exception("Error reading block %04x." % (block_addr))
+
+        block_data = response[4:]
+    except errors.RadioError:
```
kk7ds, line 311: And this. I'm guessing maybe this pattern was copied from something that wrapped the reads and writes in another function that *could* raise that error? In this case, you probably want to `raise errors.RadioError` on L308, catch that specifically, re-`raise` it, and then catch `Exception` to grab anything else and raise an `errors.RadioError`.

chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1070 @@
...
+    data = radio.get_mmap()[block_addr:block_addr + block_size]
+
+    LOG.debug("Writing Data:")
+    LOG.debug(util.hexprint(cmd + data))
+
+    try:
+        serial.write(cmd + data)
+        if serial.read(1) != CMD_ACK:
+            raise Exception("No ACK")
+    except errors.RadioError:
```
kk7ds, line 330: This as well, same as above.

chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1070 @@
...
+    def process_mmap(self):
+        self._memobj = bitwise.parse(MEM_FORMAT, self._mmap)
+
+    def sync_in(self):
+        """Download from radio"""
+        try:
+            data = do_download(self)
+        except errors.RadioError:
+            # Pass through any real errors we raise
+            raise
```
kk7ds, line 461: This  is the pattern I referenced above.

chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1070 @@
...
+            raise errors.RadioError('Unexpected error communicating '
+                                    'with the radio')
+        self._mmap = data
+        self.process_mmap()
+
+    def sync_out(self):
+        """Upload to radio"""
+        try:
+            do_upload(self)
+        except Exception:
```
kk7ds, line 475: It'd be better to pass through any `RadioError` here like you do for `sync_in()` so you get the better error reporting if it happens.

chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1070 @@
...
+
+        # Signal (DTMF Encoder Group #)
+        rs = RadioSettingValueList(PTTIDCODE_LIST, PTTIDCODE_LIST[_mem.scode])
+        rset = RadioSetting("scode", "PTT ID Code", rs)
+        mem.extra.append(rset)
+
+        # # Encrypt
+        # rs = RadioSettingValueList(ENCRYPT_LIST, ENCRYPT_LIST[_mem.encrypt])
+        # rset = RadioSetting("encrypt", "Encrypt", rs)
+        # mem.extra.append(rset)
```
kk7ds, line 614: I really hate to have commented-out code like this when we can avoid it. Is this stuff you expect to enable later? It'd be nice to just keep that out in a separate commit in a branch that we can go back to later if/when you're ready. Reading the code without syntax highlighting on, it can be difficult to notice that this isn't actually live code, and since git can keep this sort of thing for us until later, we don't really need to "preserve" it here. Anyway, not too critical, but.. definitely my preference.

KC9HI:
Looks like I missed a couple of things. Standby for another push..

KC9HI:
Oops. I guess it was OK and I somehow convinced myself otherwise. Let me know if this is OK. I know have have one or more drivers to go back and revisit.

chirp/drivers/mml_jc8810.py
KC9HI: Yeah... I meant to remove these before making the PR but doing it fell through the cracks.

chirp/drivers/mml_jc8810.py
```diff
@@ -0,0 +1,1050 @@
...
+    cmd = struct.pack(">cHb", b'W', block_addr, block_size)
+    data = radio.get_mmap()[block_addr:block_addr + block_size]
+
+    LOG.debug("Writing Data:")
+    LOG.debug(util.hexprint(cmd + data))
+
+    try:
+        serial.write(cmd + data)
+        if serial.read(1) != CMD_ACK:
+            raise Exception("No ACK")
```
kk7ds, line 329: Not a big deal but you'll never see this because it will get swallowed and raised as a radioerror. You could `LOG.error()` in the handler below if you want to see it.

PR #582, opened 2023-04-09
Update es.po

MELERIX:


kk7ds:
Thanks, sorry to add two new strings right after your previous catch-up :)

PR #590, opened 2023-04-15
Add Wouxun KG-UV9K (8.33k step) support

mrtjr1:
Fixes: #6895
Add support for KG-UV9K with 8.33k step option

# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/kguv9dplus.py
```diff
@@ -2445,6 +2449,14 @@ class KGUV9PXRadio(KGUV9DPlusRadio):
     _rev = b"02"  # default rev for the radio I know about...
     _file_ident = b"kg-uv9px"
     NEEDS_COMPAT_SERIAL = False
+    _valid_steps = STEPS
+    _step_list = STEP_LIST
+
+    # Force Match Model to False to prevent Tox error for multiple drivers
```
kk7ds, line 2455: It's not tox. Tox is just running our tests. If we change to using something else (we haven't always used tox, and may not forever) then this won't make any sense :)

Further, the test isn't just there to make life difficult, it's failing because chirp itself wouldn't be able to tell the difference between the two images. Meaning a user might save radio B and then when they re-open the image, it might be detected as radio A.

Jim is pretty good about using a standard comment for all his radios, so maybe it'd be good to just replicate that:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/btech.py#L4049

chirp/drivers/kguv9dplus.py
```diff
@@ -2882,3 +2894,31 @@ class KGUV9GXRadio(KGUV9PXRadio):
...
+    VENDOR = "Wouxun"
+    MODEL = "KG-UV9K"
+    _model = b"KG-UV9D"
+    _file_ident = b"kg-uv9k"
+    _rev = b"02"  # default rev for the radio I know about...
+    NEEDS_COMPAT_SERIAL = False
+    _step_list = STEP_LIST_9K
+    _valid_steps = STEPS_9K
+
+    # Force Match Model to False to prevent Tox error for multiple drivers
```
kk7ds, line 2920: Here too.

kk7ds (commented):
Apologies for my OCD, but can you tweak these comments?

mrtjr1:
No problem.  I understand and will make changes later tonight.

kk7ds:
I just made the change so I can merge this and something else together today.

mrtjr1:
Awesome thanks.  Much appreciated

PR #609, opened 2023-05-18
Convert to PEP-517

vishwin:
Also use setuptools_scm for dynamic versioning


vishwin:
Date-based version automatically set during build when using a full git clone. Otherwise (like git archives for system packaging or shallow clones), supply `PKG-INFO`:
```
Version: 20230520
```

kk7ds (commented):
I'm not super enthusiastic about these changes (from the community, I appreciate you doing the work here) but alas. Is there some coming deadline with pip or python that makes it important for us to do this right now? Other than the fad aspect I mean :)

As before, I need some time to make sure I can get the build stuff working with this before we merge.

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+chirp = "chirp.wxui:chirpmain"
+chirpc = "chirp.cli.main:main"
+
+[tool.versioningit.tag2version]
+rmprefix = "release_"
+
+[tool.versioningit.format]
+distance = "{committer_date:%Y%m%d}"
+dirty = "{committer_date:%Y%m%d}+d{build_date:%Y%m%d}"
+distance-dirty = "{committer_date:%Y%m%d}+d{build_date:%Y%m%d}"
```
kk7ds, line 43: Ah, I didn't pay attention here. I don't think I want to do this. I want to know when people are running from git versus one of the specific builds. Someone running in the middle of two commits that went out on a specific day will look like they're running the same thing as someone on a formal build but they're not. Also, some of our agreements with network services rely on our ability to enumerate builds within a certain window for anti-spam and anti-leech abilities. This will basically make it so that every day becomes a build number even though that's not true.

As mentioned, there's more to the versioning that we use internally than just display and it really matters to me (and others) that we're able to be specific.

chirp/__init__.py
```diff
@@ -17,7 +17,12 @@
 import sys
 from glob import glob
 
-CHIRP_VERSION = "py3dev"
+if sys.version_info < (3, 8):
+    import importlib_metadata
+else:
+    import importlib.metadata as importlib_metadata
+
+CHIRP_VERSION = importlib_metadata.version("chirp")
```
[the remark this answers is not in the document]
vishwin: I just noticed that it was hardcoded to py3dev when I opened the about box so it seemed odd, especially with image formats and whatnot.

pyproject.toml
vishwin: This only works with a fully-cloned repository, ie full history. Otherwise this errors out since `committer_date` relies on a full history metadata to retrieve the commit date. When using an archive, shallow clone, etc, whatever is in `PKG-INFO` overrides any computed version.

Can also add the git hash in there to further delineate something built directly from git however.

vishwin:
> I'm not super enthusiastic about these changes (from the community, I appreciate you doing the work here) but alas. Is there some coming deadline with pip or python that makes it important for us to do this right now? Other than the fad aspect I mean :)

[Not a fad, but policy](https://peps.python.org/pep-0440/), just not enforced until now for some reason.

pyproject.toml
kk7ds: So we could make this just the githash then? or maybe `git-$githash`? That'd make me happy.

chirp/__init__.py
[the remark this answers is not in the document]
kk7ds: It's just to keep the "I'm running this from git" separate from anything that looks like a build number. `git-$hash` will be fine and more descriptive.

kk7ds:
> > I'm not super enthusiastic about these changes (from the community, I appreciate you doing the work here) but alas. Is there some coming deadline with pip or python that makes it important for us to do this right now? Other than the fad aspect I mean :)
> 
> [Not a fad, but policy](https://peps.python.org/pep-0440/), just not enforced until now for some reason.

The *policy* is only for things that exist in the python packaging ecosystem and they very definitely don't have the ability to enforce that on us. Just because we're written in python does not mean we need to exist in and abide by all the overly-draconian rules. I'm really only interested in doing as much as is required to avoid making it difficult for people that have to install via pip.

pyproject.toml
kk7ds: Hash is far more useful.

vishwin: `{revision}` is the hash.

vishwin:
> The _policy_ is only for things that exist in the python packaging ecosystem and they very definitely don't have the ability to enforce that on us. Just because we're written in python does not mean we need to exist in and abide by all the overly-draconian rules. I'm really only interested in doing as much as is required to avoid making it difficult for people that have to install via pip.

As long as you are using any standards-compliant Python package build backend, you are in the Python packaging ecosystem, even if not in PyPI. And in the case of pip, some of us operating systems/distributions cannot use pip for this purpose, but other standards-compliant tooling.

vishwin:
No longer needed, can already be built in PEP-517 mode since the standard assumes setuptools if `pyproject.toml` does not exist or does not otherwise specify a build backend.

PR #612, opened 2023-05-22
FT-450 without FUTURE

Old-Phart:
# CHIRP PR Checklist

* All tests must be passing.


CppProperties.json
```diff
@@ -0,0 +1,21 @@
...
+      "defines": [
+        "WIN32",
+        "_DEBUG",
+        "UNICODE",
+        "_UNICODE"
+      ],
+      "intelliSenseMode": "windows-msvc-x64"
+    }
+  ]
+}
```
kk7ds, line 21: This file should not be committed to the tree.

kk7ds (changes requested):
Hi Rick, as I noted in the previous one, I really can't review a diff this large. Just like with the patches on email, please separate out the changes to just what is needed to fix python3, and then any other logical changes. Cleanups should be separate if they don't change anything functional. Also, we should not need to update the image file for any of this. If there's some reason to change that, then we can discuss it, but we should make sure that all the tests pass on the existing image file to avoid introducing changes that work *because* the image was changed.

Thanks!

PR #616, opened 2023-06-01
Update es.po and a small typo fix

MELERIX:


kk7ds:
Thanks for both. In the future, if you can put a little more in the commit message for the functional change, that'd be good. The first line goes out to the users' list in the build notification, so "Update mml_jc8810 seconds to ms" even would be better :)

PR #620, opened 2023-06-01
Remove unused system imports

dforsi:
I used `autoflake` to remove unused imports of modules from the standard library of Python, but I kept all `pass` statements by adding triple quotes around two docstrings and not applying the deletion of other `pass` statements.

Running `pytest` I get the same number of pass/fail and 7 warnings less:

after the changes in this PR I get:
> 33 failed, 7348 passed, 1891 skipped, 11567 warnings

on master I get:
> 33 failed, 7348 passed, 1891 skipped, 11574 warnings

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -1379,11 +1379,11 @@ class CloneModeRadio(FileBackedRadio, ExternalMemoryProperties):
         return cls._memsize and len(filedata) == cls._memsize
 
     def sync_in(self):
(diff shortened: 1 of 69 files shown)


kk7ds:
If you're working towards this, it would be great if you could remove files from `tools/cpep8.manifest` once they pass the stricter flake8 check to keep them that way. Will rebase and merge in a sec.

PR #627, opened 2023-06-04
RT-470: support firmware update v1.22 - fixes #10572

KC9HI:
This patch updates the settings that were changed/added.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/mml_jc8810.py
+++ b/chirp/drivers/mml_jc8810.py
@@ -107,8 +107,8 @@ struct {
      rpste:4;         //      RPT Noise Clear
   u8 unused_9016:4,   // 9016
      rptrl:4;         //      RPT Noise Delay
-  u8 unused_9017:7,   // 9017
-     roger:1;         //      Roger
+  u8 unused_9017:6,   // 9017
+     roger:2;         //      Roger
   u8 unknown_9018;    // 9018
   u8 unused_9019:7,   // 9019
      fmradio:1;       //      FM Radio
@@ -205,6 +205,7 @@ PTTIDCODE_LIST = ["%s" % x for x in range(1, 16)]
 PTTLT_LIST = ["None", "100 ms"] + \
              ["%s ms" % x for x in range(200, 1200, 200)]
 QTSAVE_LIST = ["All", "RX", "TX"]
+ROGER_LIST = ["OFF", "BEEP", "TONE1200"]
 RPSTE_LIST = ["Off"] + ["%s ms" % x for x in range(100, 1100, 100)]
 SAVE_LIST = ["Off", "Normal", "Super", "Deep"]
 SCREV_LIST = ["Time (TO)", "Carrier (CO)", "Search (SE)"]
@@ -924,7 +925,7 @@ class JC8810base(chirp_common.CloneModeRadio):
                                     _settings.skey3_lp)
             basic.append(rset)
 
-        if self.MODEL in ["RT-470L"]:
+        if self.MODEL in ["RT-470", "RT-470L"]:
             # Menu 25: TOP KEY (RT-470L)
             def apply_skeytop_listvalue(setting, obj):
                 LOG.debug("Setting value: " + str(setting.value) +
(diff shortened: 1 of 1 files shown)


chirp/drivers/mml_jc8810.py
```diff
@@ -205,6 +205,8 @@
 PTTLT_LIST = ["None", "100 ms"] + \
              ["%s ms" % x for x in range(200, 1200, 200)]
 QTSAVE_LIST = ["All", "RX", "TX"]
+ROGER_LIST = ["OFF", "BEEP", "TONE1200"]
+ROGER470_LIST = ["OFF", "BEEP", "TONE1200 [fw v1.22+]"]
```
kk7ds, line 209: I would really rather we not do this (make an option with a firmware version in it), especially since it's not like the user needs to select one or the other. It looks to me like below you're able to determine which model by looking at `MODEL`, why do we need different lists with slightly different names? Can't you do the right thing for the newer firmware if `MODEL==RT470L` in the set without the user having to see the difference?

KC9HI: I would really rather not do this either. But I can't help the fact that the RT-470 previous choices were ON/OFF (which I handled using RadioSettingValueBoolean()) and  (currently latest released) fw v1.22 changed them to OFF/BEEP/TONE1200.

The RT-470L is not an RT-470 with newer firmware. It is a totally different radio. It has different rf.valid_power_levels, different rf.valid_bands and different supported menus.

The RT-470 user has to "see" the difference because if the radio has not had its firmware upgraded to v1.22, then how will the user know that the choosing 'TONE1200' will do nothing (most likely behave the same as selecting OFF)? Not knowing creates the potential for the user to open a bug report because choosing 'TONE1200' didn't result in the radio emitting 'TONE1200' when the PTT button is released. 

It is the same situation for the TOP KEY (Short Press) selection. The RT-470 fw v1.22 added 'hidden' support for the TOP KEY (Short Press). If the CHIRP user with an RT-470 having a firmware version prior to v1.22 attempts to make any changes to the TOP KEY (Short Press) setting in CHIRP, nothing will happen in the radio. The TOP KEY (Short Press) is not supported prior to v1.22 so short pressing the Top Key does nothing.

It would be great if the RT-470 image contained the firmware version. I would have used this to cleanly display what settings that use had available. But I have searched the RT-470 image (even after expanding it to 64k) for anything that resembles the firmware version and have not been able to locate anything. I have also reached out to some RT-470 owners that have been actively hacking the radio to see if they have discovered any bytes that would seem to reveal the firmware version and so far nothing.

My only other solution would be to have the firmware version in the "MODEL =" that would allow to choose the RT-470 model based on the radio's firmware version. I don't consider this an acceptable solution.

So what I have done, and should be pushing later today, is to rip out all the "v1.22" options from the driver. Everything for the RT-470 is now based on the radio having v1.22 firmware. I have added "Notice to Developer" comments that state which firmware version that the driver support for each model is based on. I have also added comments at the setting that mention what the v1.22 firmware added/changed and that these additions/changes are not compatible with previous RT-470 firmware versions.

chirp/drivers/mml_jc8810.py
```diff
@@ -934,7 +935,16 @@ def apply_skeytop_listvalue(setting, obj):
...
 
-            if self.MODEL in ["RT-470L"]:
+            if self.MODEL in ["RT-470"]:
+                # ==========
+                # Notice to developers:
+                # The RT-470 v1.22 firmware added 'hidden' support for the
+                # Top Key (Short Press) feature. RT-470 radios with a firmware
+                # version prior to v1.22 will not honor the Top Key (Short
+                # Press) setting in CHIRP.
+                # ==========
```
kk7ds, line 945: This is unfortunate for sure, but I prefer this. Is the firmware update-able by the user or just newer radios shipping with newer stuff?

What about adding a info or experimental prompt for the user explaining that the driver expects the newer firmware and some options will not be honored on older ones?

PR #635, opened 2023-06-09
FT1, FT2 Settings and specials

Darieb:
# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/ft1d.py
```diff
@@ -1147,31 +1147,32 @@ def _get_aprs_beacons(self):
...
+                            "aprs_beacon.dst_callsign%d" % index,
+                            "DST Callsign %d" % index, val)
+                    menu.append(rs)
 
-            if int(aprs_meta[index].sender_callsign[0]) != 255:
                 date = "%02d/%02d/%02d" % (
                     aprs_meta[index].date[0],
                     aprs_meta[index].date[1],
                     aprs_meta[index].date[2])
+                LOG.warn("\nDate=%s", date)
```
kk7ds, line 1468: This isn't really a good use for being a warning, as far as I can tell. I assume this is just debug right? Also please don't put newlines in log messages unless absolutely necessary.

chirp/drivers/ft1d.py
```diff
@@ -1826,6 +1979,195 @@ def _get_scan_settings(self):
...
+    def _get_digital_settings(self):
+        topmenu = RadioSettingGroup("digital_settings", "Digital")
+        menu = RadioSettingGroup("settings", "Digital Modes")
+        topmenu.append(menu)
+        GMmenu = RadioSettingGroup("first_settings", "Group Monitor(GM)")
+        topmenu.append(GMmenu)
+        WXmenu = RadioSettingGroup("WiresX_settings", "Wires-X")
+        topmenu.append(WXmenu)
+#        LOG.debug("Starting digital settings:\nMenu: '%s'"
+#                  "GMmenu: '%s'" % (topmenu, GMmenu))
```
kk7ds, line 2132: Can you remove these commented out lines? I appreciate you removing some old stale ones in the previous commit :)

chirp/drivers/ft1d.py
```diff
@@ -1826,6 +1979,195 @@ def _get_scan_settings(self):
...
+
+        # 07  BEP.STB    Standby Beep in the digital C4FM mode. On/Off
+
+        val = RadioSettingValueList(
+            self._STANDBY_BEEP,
+            self._STANDBY_BEEP[digital_settings.standby_beep])
+        rs = RadioSetting("digital_settings.standby_beep",
+                          "Standby Beep", val)
+        menu.append(rs)
+#        LOG.debug("Finished digital settings: '%s'" % str(menu))
```
kk7ds, line 2198: This too (and any others).

chirp/drivers/ft1d.py
```diff
@@ -1826,6 +1979,195 @@ def _get_scan_settings(self):
...
+                                      r.Rooms[j])
+                WXmenu.append(rs)
+                cn = ''
+                for l in range(0, 16):
+                    s = r.Rooms[j].name[l]
+                    if int(s) != 0xff:
+                        cn = cn + str(s)
+#                if int(r.Rooms[j].name[1]) != 0xFF:
+#                    cn = bytes(str(r.Rooms[j].name).rstrip(' '),
+#                               "ascii").rstrip(b"0xFF")
```
kk7ds, line 2271: This too

chirp/drivers/ft1d.py
```diff
@@ -204,17 +205,29 @@
      autostep:1,
      automode:1,
      unknown9:3;
-} memory[%d];
+};
 
-#seekto 0x280A;
-struct {
+#seekto 0x2D4A;
+struct  memslot  memory[%d];
```
kk7ds, line 211: So if I may make a suggestion here. Instead of using these anonymous position-based format specifiers, it would be nice to use named ones. So this line would be:
```
struct memslot memory[%(max_mem)d];
```
and then your code below can be something like:
```
mem_params = {
    'max_mem': 900,
    'max_flag': 900,
    'dtmf_loc': 0xE4A,
}
mem_format % self._mem_params
```
Doing that will be much easier to reason about and debug...

chirp/drivers/ft2d.py
```diff
@@ -74,9 +73,9 @@ class FT2D(ft1d.FT1Radio):
 
     _model = b"AH60M"  # Get this from chirp .img file after saving once
     _has_vibrate = True
-    _mem_params = (0x94a,          # Location of DTMF storage
-                   999,            # size of memories array
-                   999,            # size of flags array
+    _mem_params = (900,            # size of memories array
+                   900,            # size of flags array
+                   0xe4a,          # location of DTMF storage
```
kk7ds, line 78: I haven't looked yet, but does anything need changing for the FT3? I have an FT3 and can test this later.

chirp/drivers/ft1d.py
```diff
@@ -899,34 +902,111 @@ def _strip_ff_pads(cls, messages):
...
+        elif memref > self.MAX_MEM_SLOT:         # numbered special
+            ndx = memref - (self.MAX_MEM_SLOT + 1)
+            for x in self.class_specials:
+                if ndx < len(x[1]):
+                    array = x[0]
+                    sname = x[1][ndx]
+                    break
+                ndx -= len(x[1])
+            if array is None:
+                Log.debug("memref number %d out of range", memref)
```
kk7ds, line 961: `LOG`?

kk7ds (commented):
Hi Declan,

I'm working my way through these but I've got to stop for some weekend activities. I wanted to post the comments I have so far so you can have something to chew on.

I pretty much stopped at the special memories one, which I think probably will take some focus. Do you want to maybe pull out the commits before that one into a separate PR, make the changes I describe and put them into a separate PR so we can get them merged and done with? That would reduce the size of this one and make some incremental progress.

Thanks!

kk7ds:
Also, thanks very much for splitting it up. It's much easier to review that way. If you need help with the workflow for making changes to the relevant commits instead of just tacking things on the end, let me know and I can help with that.

Darieb:
Thanks so much for having checked and reported back. I'll try to fix and resubmit promptly. I especially like the mem_params technique... it's MUCH more portable. Sorry to have left in the debug lines (both with and without comments.) 

Yes, I'd like a tutorial for fixing individual commits instead of all-at-the-end, please. Unless these changes are trivial enough that all-at-the end is manageable (clearly that's what I did for the last commit.

kk7ds:
So the first thing to know is `git commit -a --amend`... That lets you "amend" some changes into the top commit. So if you were to commit, then realize you needed a style fix or adjust the commit message, just amend that commit. The second piece is how to go back to a commit that was a few commits ago. That's an "interactive rebase". You need to give it a target so it knows how far back to go. If your set is based on `origin/master` then you can use that. Doing `git rebase -i origin/master` will pop up an editor showing you all your commits. Change "pick" to "edit" on the one(s) you want to fix. Then it will go back in time, with that commit as the most recent. You can make a change and then `git commit -a --amend` to make a few changes to it. To roll to the next one (or all the way to the top) run `git rebase --continue`. Once you've rewritten history like that, git will refuse to push to this PR like normal, since history doesn't match what it expects. Just add `-f` (force) and it will update this PR with what you've got.

I'd definitely prefer to have even small style fixes applied to the actual commit that you need fixing if it's doable.

Thanks!

chirp/drivers/ft1d.py
```diff
@@ -1097,52 +1089,55 @@ def _debank(self, mem):
...
+            flag.used = not mem.empty
+            flag.valid = True
+            flag.skip = mem.skip == "S"
+            flag.pskip = mem.skip == "P"
+        freq = mem.freq
+        if regtype in ["VFO", "Home"]:
+            freq = self.enforce_band(_mem, mem.freq, ndx, regtype, flag)
+        _mem.freq = int(freq / 1000)
+        _mem.offset = int(mem.offset / 1000)
         _mem.label = self._encode_label(mem)
```
kk7ds, line 1166: I just wan to be sure..all the special memories also have a label field? In a lot of other radios, you can't give the "home" or "scan edge" channels a name. If these can, then that's fine, but if not we should mark those as immutable.

chirp/drivers/ft1d.py
```diff
@@ -1089,6 +1104,21 @@ def _debank(self, mem):
...
+        raise an exception to cause UI to pop up an error message
+        """
+        frange = VALID_BANDS[num]
+        if freq >= frange[0] and freq <= frange[1]:
+            mem.freq = freq
+            return freq
+        mem.freq = frange[0]
+        raise errors.RadioError("Frequency out of range"
+                                " for %s register." % regtype)
+        return frange[0]
```
kk7ds, line 1122: It makes no sense to return something after a raise.

chirp/drivers/ft1d.py
```diff
@@ -1089,6 +1104,21 @@ def _debank(self, mem):
         for bank in bm.get_memory_mappings(mem):
             bm.remove_memory_from_mapping(mem, bank)
 
+    def enforce_band(self, mem, freq, num, regtype, flag):
```
kk7ds, line 1109: I don't see where you're ever calling this. Am I missing it? If I try this whole PR locally and I try to set 144MHz into the 50MHz home channel, it does not raise an error and changes to 500MHz.

The right way to do this is override `validate_memory()` in your class and return error messages if/when necessary, which the UI will show to the user.

chirp/drivers/ft1d.py
Darieb: You are right. Only memory-memory and Home can take a label.  I'll fix that.

chirp/drivers/ft1d.py
```diff
@@ -899,34 +902,111 @@ def _strip_ff_pads(cls, messages):
...
+        array = None
+        num = memref
+        sname = memref
+        mstr = isinstance(memref, str)
+        specials = ALLNAMES
+        extr = False
+        if extref is not None:
+            extr = extref in specials
+        if mstr or extr:        # named special?
+            num = -1            # I'm using negative for specials
```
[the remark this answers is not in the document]
Darieb: Positives work mostly, but still need some help in Home specials.

chirp/drivers/ft2d.py
Darieb: FT2 and FT3 were faking access to the skip memories by claiming 999 memories (900 real-memory and 99 skip-memory; they're contiguous in the radio.) Re-factoring so Skip-memory as special allows return to 900 memories and associated flags.

chirp/drivers/ft1d.py
Darieb: Great idea. It'll require change to the FTM-3200D driver, but darn well worth it. Indeed, for the time being I still define that max_flag, but don't use it; I can use max_mem in each case, rather than duplicate the positional parameter.

chirp/drivers/ft1d.py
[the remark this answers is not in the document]
Darieb: Ack. 
I'm using positive integers, but now something weird is happening... when I set a special memory, my radio structures get updated, and the display seems ok, but CHIRP complains as below. It looks to me as if _special_numbers doesn't have the index for that special register?! [1100 is the index for the first HOME register]. I don't know what I've done to bollix the _special_numbers structure. (using the "interact with driver" doesn't allow me to see what's in the wxui structures at all.
Help, please


> ERROR: <function ChirpMemEdit._memory_edited at 0x11176dee0> raised unexpected exception

> Traceback (most recent call last):
>   File "/Users/declan/chirp/chirp/wxui/common.py", line 452, in run_safe
>     return fn(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 1172, in _memory_edited
>     self.set_memory(mem)
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 852, in set_memory
>     self.do_radio(set_cb, 'set_memory', chirp_common.FrozenMemory(mem))
>   File "/Users/declan/chirp/chirp/wxui/common.py", line 192, in do_radio
>     cb(job)
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 846, in set_cb
>     self.do_radio(extra_cb, 'set_memory_extra', mem)
>   File "/Users/declan/chirp/chirp/wxui/common.py", line 192, in do_radio
>     cb(job)
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 837, in extra_cb
>     self.refresh_memory(mem.number)
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 827, in refresh_memory
>     executor(get_cb, 'get_memory', number)
>   File "/Users/declan/chirp/chirp/wxui/common.py", line 192, in do_radio
>     cb(job)
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 823, in get_cb
>     executor(extra_cb, 'get_memory_extra', job.result)
>   File "/Users/declan/chirp/chirp/wxui/common.py", line 192, in do_radio
>     cb(job)
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 806, in extra_cb
>     self._refresh_memory(number, job.result)
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 736, in _refresh_memory
>     row = self.mem2row(number)
>           ^^^^^^^^^^^^^^^^^^^^
>   File "/Users/declan/chirp/chirp/wxui/memedit.py", line 700, in mem2row
>     return self._special_rows[self._special_numbers[number]]
>            ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> KeyError: 1100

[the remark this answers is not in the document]
kk7ds: You've likely got a gap or overlap in some of your special channels. The UI remembers the virtual numbers you give to your specials the first time it pulls them, so if they ever change you'll hit this.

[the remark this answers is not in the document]
Darieb: OK, thanks. I suppose that's possible. What field contains the "virtual numbers"??? Is it chirp.common.Memory().number ??? How does chirp.common.Memory().extd_number fit in? AFAICT, my get_memory and probably set_memory can be called with a number OR a specials name, so I must be able to handle either form and get the correct correlation.

I did a quick print test in get_memory:
With negative indices I get .number=1..900 memories, -1... -199 for specials. (0 is missing!) I get no _special_numbers problems AFAICT; it's easy to check the boundaries for the different specials. The value coming in to get_memory includes the special name, instead of a numeral or number, so the boundaries are obvious.

With positive indices, .number=1..900 memories, 901-1110 for specials. No overlaps that I see, no holes. but still the KeyError at 1100.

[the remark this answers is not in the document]
kk7ds: `mem.number` needs to be an `int`, `mem.extd_number` is the string special name (only for special memories). `get_memory()` is called with either an `int` for regular memories or a `str` for specials. Since `set_memory()` only takes a `Memory` object, the former rules apply. By "virtual numbers" I meant the integer numbers for the specials that are greater than the end of the regular memory bounds.

If you push your code up I can possibly take a quick look, but it's hard to speculate about code I can't see...

Darieb:
Too many changes to handle. I'll send separate PRs.

PR #638, opened 2023-06-11
First of many steps towards refactoring the anytone.py driver...

polyglot-jones:
First of many steps towards refactoring the anytone.py driver to extract the core commonality among the variant models (e.g. Anytone 5888UV vs. the Powerwerx DB-750X).

This step encapsulates the communication protocol (via the "pipe.")

This new logic will first be used in the upcoming powerwerx.py driver. The older anytone*.py drivers will be refactored after that.

### DIFF
--- a/chirp/drivers/anytone_clone.py
+++ b/chirp/drivers/anytone_clone.py
@@ -1,0 +1,269 @@
+# Copyright 2023 Dan Smith <dsmith@danplanet.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from builtins import bytes
+import struct
+import logging
+from abc import ABC, abstractmethod
+
+from chirp import util, errors
+
+LOG = logging.getLogger(__name__)
+
+
+# ############################################################################
+#                                                 Radio Communication Protocol
+# ############################################################################
+
+# TODO Consider lifting this to common for all drivers to utilize
+class RadioCommunicationProtocol(ABC):
+    """Abstract base class for encapsulating the logic that communicates with
+    the radio -- while remaining agnositic as to what content is being
+    communicated.
+
+    Usage example (within the download code, similar for upload):
+        with MyProtocol(pipe, block_size, file_ident) as protocol:
+            # `protocol.start_session()` is automatically called.
+            # `protocol.verify_radio_ident()` is automatically called.
+
+            for ...
+                for ...
+                    block = protocol.read_block(addr)
+                    data += block
+
+            # `protocol.end_session()` is automatically called.
+
+    """
+
+    def __init__(
+        self, pipe, block_size: int, file_ident, echos_write=True
+    ) -> None:
+        """@file_ident can be a single ID or a list of multiple possible
+        IDs."""
+        assert block_size > 0 and block_size % 2 == 0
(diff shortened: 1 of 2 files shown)


polyglot-jones:
1. The style check will pass after PR #637 is merged.
2. I don't know why the PR Checks is complaining about a new use of `future`. There isn't any.

chirp/drivers/anytone_clone.py
```diff
@@ -0,0 +1,269 @@
...
+                    block = protocol.read_block(addr)
+                    data += block
+
+            # `protocol.end_session()` is automatically called.
+
+    """
+
+    def __init__(
+        self, pipe, block_size: int, file_ident, echos_write=True
+    ) -> None:
```
kk7ds, line 52: I think I've mentioned this on the list, but surely many times in other PRs. I absolutely detest black formatting style (as do many others). I'm sorry you've done this work in this way and will have to reformat it, but that's how it is. I have to deal with it at work, but I get to choose here. These parens on their own lines are one of the ugliest aspects of it, IMHO.

Most of the rest of the chirp code is fairly conventional python formatting. Please reformat this to look like the rest of the codebase.

kk7ds (commented):
You asked about extending to longer lines and I said I wasn't going to change that. I hate black formatting even more, and even if I didn't, none of the rest of the code looks like this.

chirp/drivers/anytone_clone.py
```diff
@@ -0,0 +1,269 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from builtins import bytes
```
kk7ds, line 16: The "no new uses of future" is because you're using this compatibility layer, but shouldn't.

kk7ds:
I assume this is on hold until you refactor the existing anytone driver as another commit here, or add the new driver? Point being, there's no reason to merge this until something is using it. Thanks for adding model-specific tests - I wish I could get more people to do that, of course.

chirp/drivers/anytone_clone.py
polyglot-jones: If not Black, then which tool should I use? Any special settings?

chirp/drivers/anytone_clone.py
polyglot-jones: Oh, I didn't realize. That just carried over form the previous code and I didn't even notice it. Thanks.

polyglot-jones:
I started committing this is in chunks per the PR rules to keep it to one conceptual unit per PR. Are you saying that you want me instead to commit the full anytone_clone.py file (including the bank model and the radio base class) and the powerwerx.py file all together in one PR?

polyglot-jones:
BTW, I only own a Powerwerx DB750X, not any of the other Anytones. So when I go to check in those refactored drivers, I'll need someone who does to beta test them. I was planning on asking the dev and user mailing lists when the time comes.

chirp/drivers/anytone_clone.py
kk7ds: I can't offer you a tool to automatically format the code in more conventional python and chirp style, as you know. You can start by writing code that the current checker (flake8) will allow, and not adding arbitrary vertical stacking of things that will fit perfectly fine on fewer lines. Anywhere there's a closing paren on a line by itself is an easy regex for not-okay-ness :)

kk7ds:
Well, I would call "refactoring an existing anytone driver to allow for reuse" as one conceptual thing. Or a commit with common bits, followed by a commit of your new driver to use them even would be fine. What I *really* don't want is single commits that make multiple changes, and to a lesser degree, PRs that do a bunch of random gardening in multiple areas. Multiple related commits in a PR are totally fine.

I have a 5888UV I can test with, if that's one of the ones that will be affected. Something else I did during the py3 conversion was write a clone simulator for a radio that will pass with the current code and then change the code itself. That seemed pretty accurate, and you've already demonstrated the ability to do that here, so maybe that's a strategy in the absence (or in addition to) a volunteer.

PR #641, opened 2023-06-12
Allow opening child windows and moving tabs

kk7ds:
This adds the ability to File->New Window, drag tabs between windows,
and open files in specific windows.

### DIFF
--- a/chirp/locale/de.po
+++ b/chirp/locale/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-06-12 15:03-0700\n"
+"POT-Creation-Date: 2023-06-13 18:59-0700\n"
 "PO-Revision-Date: 2012-10-02 22:11+0100\n"
 "Last-Translator: Benjamin, HB9EUK <hb9euk@hb9d.org>\n"
 "Language-Team: German\n"
@@ -35,22 +35,22 @@ msgstr ""
 msgid "%(value)s must be between %(min)i and %(max)i"
 msgstr ""
 
-#: ../wxui/memedit.py:1388
+#: ../wxui/memedit.py:1466
 #, fuzzy, python-format
 msgid "%i Memories"
 msgstr "Speicher"
 
-#: ../wxui/memedit.py:1395
+#: ../wxui/memedit.py:1473
 #, fuzzy, python-format
 msgid "%i Memories and shift all up"
 msgstr "Löschen (und nach oben verschieben)"
 
-#: ../wxui/memedit.py:1391
+#: ../wxui/memedit.py:1469
 #, fuzzy, python-format
 msgid "%i Memories and shift block up"
 msgstr "Löschen (und nach oben verschieben)"
 
-#: ../wxui/main.py:1102
+#: ../wxui/main.py:1233
 #, fuzzy, python-format
 msgid "%s has not been saved. Save before closing?"
 msgstr "Datei wurde geändert, speichern Sie die Änderungen vor dem Schliessen?"
@@ -591,17 +591,17 @@ msgid ""
 " click <b>Then"
 msgstr ""
 
-#: ../wxui/main.py:1481
+#: ../wxui/main.py:1617
 msgid ""
 "A new CHIRP version is available. Please visit the website as soon as "
 "possible to download it!"
 msgstr ""
 
-#: ../wxui/main.py:694
+#: ../wxui/main.py:817
 msgid "About"
 msgstr ""
 
-#: ../wxui/main.py:1343
+#: ../wxui/main.py:1479
 msgid "About CHIRP"
 msgstr ""
 
@@ -609,12 +609,12 @@ msgstr ""
 msgid "All"
 msgstr "Alle"
 
-#: ../wxui/main.py:969
+#: ../wxui/main.py:1100
 #, fuzzy
 msgid "All Files"
 msgstr "CSV Datei"
 
-#: ../wxui/main.py:980
+#: ../wxui/main.py:1111
 msgid "All supported formats|"
 msgstr ""
 
@@ -634,7 +634,7 @@ msgstr ""
 msgid "Available modules"
 msgstr ""
 
-#: ../wxui/main.py:1427
+#: ../wxui/main.py:1563
 msgid "Bandplan"
 msgstr ""
 
@@ -642,7 +642,7 @@ msgstr ""
 msgid "Bands"
 msgstr ""
 
-#: ../wxui/main.py:138
+#: ../wxui/main.py:196
 msgid "Banks"
 msgstr "Bank"
 
@@ -650,7 +650,7 @@ msgstr "Bank"
 msgid "Bin"
 msgstr ""
 
-#: ../wxui/main.py:149
+#: ../wxui/main.py:207
 msgid "Browser"
 msgstr ""
 
@@ -662,13 +662,13 @@ msgstr ""
 msgid "Canada"
 msgstr ""
 
-#: ../wxui/memedit.py:1086
+#: ../wxui/memedit.py:1164
 #, python-format
 msgid ""
 "Channels with equivalent TX and RX %s are represented by tone mode of \"%s\""
 msgstr ""
 
-#: ../wxui/main.py:968
+#: ../wxui/main.py:1099
 msgid "Chirp Image Files"
 msgstr ""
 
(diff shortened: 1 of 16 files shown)


kk7ds:
@KC9HI I think you tend to like a side-by-side workflow. Mind trying this out to see if it's worthwhile? It could definitely could use some stress testing, as I haven't even tried it with a real radio. The tabs *are* draggable between the windows, but wx doesn't make it very clear. I don't love that part of the experience, but maybe it's something that can be improved later.

kk7ds:
Don't start a new copy, just go File->New Window

kk7ds:
There also needs to be some sort of "detach this tab as its own window" function but I haven't decided how to represent it in the menu.

KC9HI:
That is downright awesome! I have opened a second instance of CHIRP before to have different images in each instance so I could see both tabs as I copy-and-pasted between them. This takes that to a whole new level!

kk7ds:
Woohoo, you're a hard guy to get one excited exclamation point out of, let alone two! So yeah, if you can poke around and try to find any weirdness, that'd be cool. I think I need to squash the usual restore-the-window-size-and-position logic in the child windows because otherwise they just open right on top of the first one and thus are hard to tell apart. But let me know if you find anything else that seems wrong.

KC9HI:
Just thinking out loud... would there be a way to "right-click" and drag and then get a prompt with Copy or Move choices? Or Ctrl + click and drag for copy, etc.

kk7ds:
On which? Tabs or memories? The way the per-file tabs are drawn I don't get a whole lot of control over them, which is pretty annoying. I was hoping for a right-click ... detach option, but I'm not sure I can do that. I'll keep looking though.

I assume you've seen that you can drag per-file tabs from one window to the other. It's just not obvious that it's working until you're hovering over the tab bar on the other side. That's the thing I don't like, but I just don't have much control over it :/

Drag-and-drop of memories is possible for sure, I've just never been big on drag-and-drop of data like that, partially because it always seems to behave slightly differently on each platform. But if you think it'd be widely used I could look into it.

KC9HI:
I was just thinking the tabs. But it should be a big deal to load the same file in the "new" window or download from the radio again.

Yes. It took me a little bit to understand that I needed to hover over the tab bar and release to complete the move. But if you close all of the tabs in a window, the tab bar is gone. You can close the window and open a new one or simply open a new file which adds the tab bar back.

I'm good with copy-and-paste of the memories.

kk7ds:
Yeah, the tab drag is not very intuitive, which is a bummer. I could restore the welcome tab if you close the last one so the tab bar remains, but that also kinda sucks. I'll do some thinking on it.

kk7ds:
Okay, I think this is much improved:

- Dragging a tab always works and you can drop it anywhere in the window
- Dragging a tab gives you feedback that it is happening before you drop it
- You can drop it in a specific place in the target window's tab order and it will go there
- Reordering tabs in the same window works the same way, but slightly different than it was before
- Right-click on a tab gives an "open in new window" pop-up
- Dragging memories between windows now works (but you have to grab it from a cell, not a row header, and only one at a time, which seems to be some limitation)
- Dragging memories within a single sheet also works
- Closing the first window no longer exits the whole program (not sure if you noticed this before)

On the memory dragging: The way I have it right now is dragging within a single memedit moves the memory around. Dragging between windows copies the memory. That's overriding the kinda default behavior, which is to always move, regardless of whether or not you're crossing windows. I feel like people might be surprised if they try to drag between windows and realize they've moved a memory. However, this override behavior means that the typical "hold control for copy instead of move" doesn't work properly when crossing windows (but does if within a single page).

Any thoughts on whether or not making it consistent (default move, control to copy) across windows would be better or worse?

kk7ds:
Okay, with this update, you can drag multiple memories by first selecting them then using Alt+Drag from the row labels to grab them all at once.

KC9HI:
>-  Dragging a tab always works and you can drop it anywhere in the window

Not exactly that way with Windows. It has to be dropped in the Memories/Settings tabs area and above or in the very bottom status bar area. Anywhere else in the windows and the cursor is an ISO 'Do Not Enter' symbol (circle with a back slash through it) to indicate areas where it is not permissible to drop the tab.

>-  Dragging a tab gives you feedback that it is happening before you drop it

Do you mean the rectangular outline added under the cursor? I probably wouldn't have noticed it if you hadn't said something.

>-  You can drop it in a specific place in the target window's tab order and it will go there

OK. This didn't seem to be working when I tried to drop it exactly between 2 tabs. Then I finally realized that the tab is inserted to the left of the tab it is dropped on.

>-  Reordering tabs in the same window works the same way, but slightly different than it was before

I didn't try this last night. Verified.

>-  Right-click on a tab gives an "open in new window" pop-up

Verified.

>-  Dragging memories between windows now works (but you have to grab it from a cell, not a row header, and only one at a time, which seems to be some limitation)

Verified. To me grabbing from a cell (it doesn't matter which cell) is probably better. Especially if you are grabbing the memory based on data in a particular cell. Once you find it you grab the cell and move/copy it to a new location.

If you only need 2 or 3 memories moved/copied, one at a time if fine. For more than that simply bulk copy/move using cut-and-paste. 

>-  Dragging memories within a single sheet also works

Verified

>-  Closing the first window no longer exits the whole program (not sure if you noticed this before)

Verified. You are correct, I didn't notice.

Things that I noticed...

- A new child window takes on the View properties of the parent window.
- Changing a View property only affects the window it is changed in. Depending on what you are changing, this can be handy or annoying. But I think this is the correct way it should work.

So far I have not encountered any errors or glitches. Every thing seem to work in a manner that makes sense to me. I makes what I used to do by running multiple versions of CHIRP so much easier and is certainly more powerful. Nice job!

kk7ds:
> Not exactly that way with Windows. It has to be dropped in the Memories/Settings tabs area and above or in the very bottom status bar area. Anywhere else in the windows and the cursor is an ISO 'Do Not Enter' symbol (circle with a back slash through it) to indicate areas where it is not permissible to drop the tab.

Ah, well, it *did* work the way I described it before I added the draggable memories. Once the memory editor became a drop target for memories, it stops accepting the tab drag because it's the wrong data type. Interesting that it behaves that way on Windows and not mac and linux. However, I might be able to change it so that works. That said, it does let you drag it to anywhere in an empty window, right? Even if I can't make it draggable into the memory editor itself, the "target" areas are a lot larger than the previous attempt.

> Do you mean the rectangular outline added under the cursor? I probably wouldn't have noticed it if you hadn't said something.

Yes, that's what I mean. Before there was *zero* indication that it was doing anything until you got all the way over to the other tab bar. I think I can actually do more than just the box like make the tab seem to be grabbed, but not sure. I'll see what I can do.

> OK. This didn't seem to be working when I tried to drop it exactly between 2 tabs. Then I finally realized that the tab is inserted to the left of the tab it is dropped on.

Yeah, I *might* be able to tell whether you drop on the left or right of a tab and have it go the right direction, but I'm not sure. Either way, at least it's *somewhat* controllable. That function isn't implemented in the linux/GTK wx library so it's not even possible there.

> Things that I noticed...
> 
>     * A new child window takes on the View properties of the parent window.

Yeah, those settings are global (stored in the conf). Were you surprised that they were in a good way or bad way?

>     * Changing a View property only affects the window it is changed in. Depending on what you are changing, this can be handy or annoying. But I think this is the correct way it should work.

Yeah, I guess you might want the font stuff to propagate, but not the others. I'd probably not worry about making them propagate unless someone really complains just because of how complex it would be to do that.

> So far I have not encountered any errors or glitches. Every thing seem to work in a manner that makes sense to me. I makes what I used to do by running multiple versions of CHIRP so much easier and is certainly more powerful. Nice job!

Thanks, and thanks for the testing. I think this is probably good enough to merge for users to play with and I can work on the drag target stuff and maybe further indication from the cursor during the drag.

KC9HI:
> > Not exactly that way with Windows. It has to be dropped in the Memories/Settings tabs area and above or in the very bottom status bar area. Anywhere else in the windows and the cursor is an ISO 'Do Not Enter' symbol (circle with a back slash through it) to indicate areas where it is not permissible to drop the tab.
> 
> That said, it does let you drag it to anywhere in an empty window, right?
Yes. Empty or not.

> Even if I can't make it draggable into the memory editor itself, the "target" areas are a lot larger than the previous attempt.
Yes. I am totally happy with how it works.

> Yeah, those settings are global (stored in the conf). Were you surprised that they were in a good way or bad way?

I was just pointing out what I noticed. I guess since I new the View settings were global, I was expecting a change to affect all open windows. How it behaves is perfectly fine with me. I would expect that for most users, these settings are set and forget (most probably won't ever change them from their initial defaults) and will never notice.

> Yeah, I guess you might want the font stuff to propagate, but not the others. I'd probably not worry about making them propagate unless someone really complains just because of how complex it would be to do that.

I wouldn't mess with it. I just happened to have 3 windows open. I was testing drag and drop of memories and noticed that I didn't have any empty memories to drop them into. This was because I had 'Hide empty memories' enabled (which I only enabled yesterday for testing purposes). I just noticed that when I disabled it, only the empty memories in the current window appeared. I thought I should mention it. But I wouldn't expect the typical CHIRP user to notice this.

> Thanks, and thanks for the testing. I think this is probably good enough to merge for users to play with and I can work on the drag target stuff and maybe further indication from the cursor during the drag.

Yes. By all means. This is definitely something I could see myself using on a very regular basis. Thanks for implementing it!!!!!

PR #642, opened 2023-06-13
Update Spanish Translation

MELERIX:


kk7ds:
Any faster and you'll be translating these before I write them... :)

PR #653, opened 2023-06-15
Some fixes for TH-D7 and TH-D7G

W8BSD:
- Disable hardware flow control and DTR/RTS This is a 3wire serial connection, so these aren't possible, and with them enabled, it doesn't work on FreeBSD.
- Set memory bounds to 0..199 from 1..200 This matches the radio, but causes a glitch when loading where row labels are initially 1..200, then become 0..199 as the memories load.  No obvious way to fix this, but the old way you couldn't set memory 0, and attempting to set memory 200 resulted in an error.
- Suppress warning on (normal) empty DCS value
- Strings must not be padded to max length, or setting fails
- MYCALL is 9 characters, not 8
- Add "All" APRS beep setting for TH-D7G

Tested on my TH-D7G

### DIFF
--- a/chirp/drivers/kenwood_live.py
+++ b/chirp/drivers/kenwood_live.py
@@ -83,7 +83,7 @@ def _command(ser, cmd, *args):
     result = ""
     while not result.endswith(LAST_DELIMITER[0]):
         result += ser.read(COMMAND_RESP_BUFSIZE).decode('cp1252')
-        if (time.time() - start) > 0.5:
+        if (time.time() - start) > 1:  # TXH sometimes takes longer on TH-D7G
             LOG.error("Timeout waiting for data")
             break
 
@@ -395,8 +395,10 @@ class KenwoodOldLiveRadio(KenwoodLiveRadio):
 class THD7Radio(KenwoodOldLiveRadio):
     """Kenwood TH-D7"""
     MODEL = "TH-D7"
+    HARDWARE_FLOW = False
 
     _kenwood_split = True
+    _upper = 199
 
     _BEP_OPTIONS = ["Off", "Key", "Key+Data", "All"]
     _POSC_OPTIONS = ["Off Duty", "Enroute", "In Service", "Returning",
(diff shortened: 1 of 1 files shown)


chirp/drivers/kenwood_live.py
```diff
@@ -395,8 +395,12 @@ def set_memory(self, memory):
 class THD7Radio(KenwoodOldLiveRadio):
     """Kenwood TH-D7"""
     MODEL = "TH-D7"
+    HARDWARE_FLOW = False
+    WANTS_DTR = False
+    WANTS_RTS = False
```
kk7ds, line 400: Do you actually need all three of these or just `HARDWARE_FLOW`? The `WANTS_` flags are kinda weird and not actually related to hardware flow signaling. Without them in some cases some USB chips won't do *anything*. Any chance I could get you to try this with a linux machine to confirm?

kk7ds:
>     * Set memory bounds to 0..199 from 1..200 This matches the radio, but causes a glitch when loading where row labels are initially 1..200, then become 0..199 as the memories load.  No obvious way to fix this, but the old way you couldn't set memory 0, and attempting to set memory 200 resulted in an error.

Does the radio number them from 0 to 199 or 1 to 200? CHIRP should match what the radio uses. For the "no obvious fix" are you referring to the row labels changing as it loads?

W8BSD:
> Does the radio number them from 0 to 199 or 1 to 200? CHIRP should match what the radio uses.

0 to 199.

> For the "no obvious fix" are you referring to the row labels changing as it loads?

Yeah.

chirp/drivers/kenwood_live.py
W8BSD: Testing without the WANTS_* now ... IIRC, I threw all three in together because it was waiting for CD to be asserted.

kk7ds: Okay, that *should* only be tied to `HARDWARE_FLOW`. The `WANTS` ones control whether or not DTR and RTS are asserted when the port is open, which some chips require to transmit anything, even if hardware flow control is not in use. FreeBSD could definitely be different, but hopefully not.

kk7ds:
I think the row label renumbering thing is a general quirk amongst the live radios. I'll have a look.

chirp/drivers/kenwood_live.py
W8BSD: Yeah, it appears to work fine with just `HARDWARE_FLOW = False`... pushed.

kk7ds:
Thanks, glad some people still care about these old "real" radios :)

I've repro'd the renumbering thing with another live driver, so not D7 related. I'll see if I can fix that.

PR #654, opened 2023-06-15
Fix zero-origin live driver row label mismatch

kk7ds:
Zero-origin live drivers see the rows getting renamed during loading.
This fixes that.

Fixes #10641


kk7ds:
cc: @W8BSD

PR #659, opened 2023-06-16
Yaesu FTnD: Issue 6151 and 10629, WiresX and special registers.

Darieb:
FeatureAdd #6151 WiresX settings
FeatureAdd #10629 Handle special registers. This also requires changes to Yaesu FTM3200D_R, which depends upon FT1D_R. Style and driver-test fixes are included rather than in separate commits.

# CHIRP PR Checklist

The following must be true before PRs can be merged:

* All tests must be passing.
* Commits should be squashed into logical units.
* Commits should be rebased (or simply rebase-able in the web UI) on current master. Do not put merge commits in a PR.
* Commits in a single PR should be related.
* Major new features or bug fixes should reference a [CHIRP issue](https://chirp.danplanet.com/projects/chirp/issues).
* New drivers should be accompanied by a test image in `tests/images` (except for thin aliases where the driver is sufficiently tested already).

Please also follow these guidelines:

* Keep cleanups in separate commits from functional changes.
* Please write a reasonable commit message, especially if making some change that isn't totally obvious (such as adding a new model, adding a feature, etc).
* Do not add new py2-compatibility code (No new uses of `six`, `future`, etc).
* All new drivers should set `NEEDS_COMPAT_SERIAL=False` and use `MemoryMapBytes`.
* New drivers and radio models will affect the Python3 test matrix. You should regenerate this file with `tox -emakesupported` and include it in your commit.

PR #660, opened 2023-06-17
Fix broken FTM-7250 driver for recent FTx changes

kk7ds:
The changes to the ft1d (et al) drivers in 420df3e3 broke the
ftm7250 driver, which also depends on that. This fixes that driver
to behave like the base class expects.

Related to #6151
Related to #10629


kk7ds:
The changes in #659 broke the build this morning because they broke the ftm7250 driver. I didn't realize because the tests here run the fast-driver check, which doesn't properly detect drivers that subclass other drivers. I'll have to fix that.

FYI @Darieb

PR #664, opened 2023-06-20
Simple import support

kk7ds:
- Limit paste error dialog to 20 lines
- Implement simple import function
- Make radio browser read-only in frozen builds


kk7ds:
cc: @KC9HI

PR #665, opened 2023-06-21
Add many more settings to TH-D7G

W8BSD:
A large number of settings were missing, including dedicated memories like call memories and programmable scan memories.

There was also a fairly large number of options missing from TH-D7 (and TM-D700).

This is tested on TH-D7G, but should really be tested on TH-D7 and TM-D700

### DIFF
--- a/chirp/drivers/kenwood_live.py
+++ b/chirp/drivers/kenwood_live.py
@@ -22,7 +22,8 @@ import logging
 from chirp import chirp_common, errors, directory, util
 from chirp.settings import RadioSetting, RadioSettingGroup, \
     RadioSettingValueInteger, RadioSettingValueBoolean, \
-    RadioSettingValueString, RadioSettingValueList, RadioSettings
+    RadioSettingValueString, RadioSettingValueList, \
+    RadioSettingValueMap, RadioSettings
 
 LOG = logging.getLogger(__name__)
 
@@ -83,7 +84,7 @@ def _command(ser, cmd, *args):
     result = ""
     while not result.endswith(LAST_DELIMITER[0]):
         result += ser.read(COMMAND_RESP_BUFSIZE).decode('cp1252')
-        if (time.time() - start) > 1:  # TXH sometimes takes longer on TH-D7G
+        if (time.time() - start) > 0.5:
             LOG.error("Timeout waiting for data")
             break
 
@@ -168,6 +169,57 @@ def iserr(result):
     return result in ["N", "?"]
 
 
+# NOTE: There's actually a set_validate_callback() thing, but that
+#       callback doesn't get the object, so it would need to be
+#       a lambda or something to get the min/max values... this seemed
+#       more obvious.
+class KenwoodSettingProgrammableVFO(RadioSettingValueString):
+
+    """A string setting"""
+
+    def __init__(self, minval, maxval, current):
+        minlen = len("%d" % minval)
+        maxlen = len("%d" % maxval)
+        start, end = current.split(',', 1)
+        current = "%d-%d" % (int(start), int(end))
+        LOG.debug("Current = '%s'" % current)
+        RadioSettingValueString.__init__(self, minlen * 2 + 1,
+                                         maxlen * 2 + 1, current,
+                                         False, "0123456789-")
+        self._minval = minval
+        self._maxval = maxval
+        self.set_value(current)
+
+    def validate(self):
+        value = self.get_value()
+        LOG.debug("Validating... %s" % value)
+        try:
+            start, end = value.split("-", 1)
+            start = int(start)
+            end = int(end)
+            if start > end:  # TODO: Set current value back on error...
+                raise errors.RadioError("Start must be less than or equal "
+                                        "to end")
+            if (start < self._minval):
+                raise errors.RadioError("Start must be greater than or equal "
+                                        "to %d" % self._minval)
+            if (start > self._maxval):
+                raise errors.RadioError("Start must be less than or equal "
+                                        "to %d" % self._maxval)
+            if (end < self._minval):
+                raise errors.RadioError("End must be greater than or equal "
+                                        "to %d" % self._minval)
+            if (end > self._maxval):
+                raise errors.RadioError("End must be less than or equal to " +
+                                        "%d" % self._maxval)
+        except errors.RadioError:
+            raise
+        except:
+            raise errors.RadioError("Programmable VFO must be Start-End " +
+                                    "format (ie: 144-147)")
+        return "%05d,%05d" % (start, end)
+
+
 class KenwoodLiveRadio(chirp_common.LiveRadio):
     """Base class for all live-mode kenwood radios"""
     BAUD_RATE = 9600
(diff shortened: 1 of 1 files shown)


W8BSD:
So this isn't well tested enough to merge I think, but I'm not sure I've done everything "right", and I'm not sure how testing on TH-D7 and TM-D700 could be arranged.

chirp/drivers/kenwood_live.py
```diff
@@ -83,7 +84,7 @@ def _command(ser, cmd, *args):
     result = ""
     while not result.endswith(LAST_DELIMITER[0]):
         result += ser.read(COMMAND_RESP_BUFSIZE).decode('cp1252')
-        if (time.time() - start) > 1:  # TXH sometimes takes longer on TH-D7G
```
kk7ds, line 86: Are you removing this comment because it references the D7G and you don't think it's required? I'd have to go look at the history to see when and why it was added, but changing this now worries me a bit because of all the (many) drivers that could depend on it, despite what the comment says. Can you explain why you want to change this?

chirp/drivers/kenwood_live.py
```diff
@@ -168,6 +169,53 @@ def iserr(result):
...
+                                        "to %d" % self._minval)
+            if (end > self._maxval):
+                raise errors.RadioError("End must be less than or equal to " +
+                                        "%d" % self._maxval)
+        except errors.RadioError:
+            raise
+        except:
+            raise errors.RadioError("Programmable VFO must be Start-End " +
+                                    "format (ie: 144-147)")
+        return "%05d,%05d" % (start, end)
```
kk7ds, line 220: Er, this format is different than what you say it needs to be above in the error and in your splitting. I feel like it would be better to expose this as two settings, an upper and lower, like the other radios do...

chirp/drivers/kenwood_live.py
```diff
@@ -198,39 +246,72 @@ def __init__(self, *args, **kwargs):
 
             command(self.pipe, "AI", "0")
 
-    def _cmd_get_memory(self, number):
-        return "MR", "%i,0,%03i" % (self._vfo, number)
+    def _cmd_get_memory(self, number, cmd="MR"):
+        if isinstance(number, str):
+            return cmd, "%i,0,%s" % (self._vfo, number)
```
kk7ds, line 255: So these look like `U1` or something is that right? I'm not sure I like making these take a string or an int, but I acknowledge it looks like the easiest path.

chirp/drivers/kenwood_live.py
```diff
@@ -198,39 +246,72 @@ def __init__(self, *args, **kwargs):
...
+            read_cmd = "MR"
         if number in self._memcache and not NOCACHE:
             return self._memcache[number]
 
-        result = command(self.pipe, *self._cmd_get_memory(number))
+        result = command(self.pipe, *self._cmd_get_memory(cmdnum, read_cmd))
         if result == "N" or result == "E":
             mem = chirp_common.Memory()
+            if self._has_name:
+                if name is not None:
```
kk7ds, line 317: You're using this wrong. The `name` property is owned by the user and is only the alpha tag. If you can't have an alpha tag on these then you should set it to the empty string and make it immutable. What you're looking for is "special channels". You need to expose `valid_special_chans` in your `RadioFeatures`. You still need to give them a virtual `mem.number` like you are here, but set the name (like "Call 1") as `mem.extd_number`. That tells the UI that it's special and not part of the regular number memories. Check out some of the other drivers in the tree that expose those, and maybe open up an image file (in `tests/images`) of one to get an idea.

chirp/drivers/kenwood_live.py
```diff
@@ -616,18 +1211,778 @@ class THD7GRadio(THD7Radio):
...
+                 ("MP 3", "My Position 3", 10, MAIDENHEAD_CHARSET)],
+    }
+
+    EXTRA_MAP_SETTINGS = {
+        'main': [("BEL 0", "Tone Alert Band A"),
+                 ("BEL 1", "Tone Alert Band B"),
+                 ("SQ 0", "Band A Squelch"),
+                 ("SQ 1", "Band B Squelch")],
+        'aprs': [("TXD", "Transmit Delay")],
+    }
```
kk7ds, line 1985: I wonder if it makes sense to pull the D7 out to another file at this point? You've added a lot to this file, which was already too big. You can just import `kenwood_live` and do the same thing, but keep all this verbose settings stuff to one file. What do you think? If you do, it'd be great if you could make one commit that splits the existing driver to a new file (unchanged) and then another commit to make your changes here, just so it's still easy to review the delta. I know that's more work, but hopefully you'll agree it's worthwhile. This file has just grown organically beyond what it really should have in it :)

Maybe move the D700 in there as well?

kk7ds (changes requested):
Thanks for all this work. I have very little time to review this today and then I'm out for the next four days. But, I skimmed and gave you some initial feedback. When I get back, I can test on a D700 if you want.

chirp/drivers/kenwood_live.py
W8BSD: I added this in commit 4ff976b, and the problem was absolutely a bad cable I was using.

kk7ds: LOL, sorry. I guess I'm showing how small my buffer is. In my defense, there are a lot of drivers :)

chirp/drivers/kenwood_live.py
W8BSD: When modifying on the radio, it's presented as "XXX-XXXMHz", and you can change between the two fields.  Since the two values are set together with the same command, I wanted it to stay together since it seems that every time you change a setting, all the settings that have been changed are also updated, so I didn't want weird ordering "stuff" to happen.

chirp/drivers/kenwood_live.py
W8BSD: Yeah, "L0", "U0" matches what's passed in that field, and what is displayed, so I did it the hackish way.

chirp/drivers/kenwood_live.py
W8BSD: Thanks, will do... it was very hard to get it working this way, so I suspect I was doing this wrong. :D

chirp/drivers/kenwood_live.py
kk7ds: Okay, so you'll get *all* the settings at once, always. So I think you can handle it in a stable way even if they're separate. But yeah, let me think on this after I play with it on a D700.

chirp/drivers/kenwood_live.py
W8BSD: Makes sense, I was already considering it... at the same time, I'll make an abstract base class that they all share and hopefully come up with something better than the _SETTINGS_OPTIONS{}/_XXX_OPTIONS override mechanism that's currently in use.

The maidenhead conversion stuff doesn't *really* belong here either, it was just the easiest way to get lat/lon data into a single text box.  The radio is set in XX°YY.YY′ format, but having four boxes all controlling the same set command felt more awkward to me.

W8BSD:
> When I get back, I can test on a D700 if you want.

It's great that you have one!  I looked for one at Hamvention this year and the only one in the flea market was priced at $525 and my upper limit was $350 with an expected price of around $300, so I didn't even bother trying to haggle him down.

Finding a TH-D7A is pretty tough too it seems.

kk7ds:
Yeah, I have *two* of them, even! I'll poke at one with this when I'm back. Thanks!

chirp/drivers/kenwood_live.py
kk7ds: Okay, cool. This definitely needs some cleanup, we just need to be careful not to disturb all the drivers under this that will be hard to find testers for. I have a D700, D710G, D72, D74, and TM271 I can use for verification and I *think* @KC9HI has an F6A, IIRC.

kk7ds:
I quickly broke out the D700 to validate your split PR so we can merge that and let you get underway with your fixes. While I did that I quickly tested this one and got:
```
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_live.py", line 333, in get_memory
    mem = self._parse_mem_spec(spec, read_cmd)
TypeError: TMD700Radio._parse_mem_spec() takes 2 positional arguments but 3 were given
```

I haven't looked at why, but should be an easy fix. Just recording for posterity while I have it out.

W8BSD:
Closed in favour of #674

PR #670, opened 2023-06-22
Split TH-D7, TH-D7G, and TH-D700 into a separate file

W8BSD:
This is in preparation of a settings overhaul.


kk7ds:
You can ignore the PR checks failure that complains about a missing image (obviously). The style checks are probably failing because your new file is being held to the stricter standards than the old one, which is in the exclusion list. Just add yours to `tools/cpep8.manifest` and it should pass that style check.

kk7ds:
Works with my D700. Let's get the style thing fixed so I can merge this before I leave to unblock your work.

kk7ds:
Actually, the list of failures is small. Let's just fix them and keep strict mode on for the new file. I'll push a fix for this, just a sec.

W8BSD:
> I'll push a fix for this, just a sec.

Whoops, I clobbered yours. :(

chirp/drivers/kenwood_d7.py
```diff
@@ -0,0 +1,344 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from chirp.drivers.kenwood_live import KenwoodOldLiveRadio, MODES, STEPS, \
+    DUPLEX, LOG, get_tmode
```
kk7ds, line 17: Importing LOG is technically not correct because that logger will log as `kenwood_live` and not `kenwood_d7` but it's not critical. Just fix that up in your settings series.

kk7ds:
No problem :)

chirp/drivers/kenwood_d7.py
W8BSD: Heh, holding off on another force push with the proper fix. :)

kk7ds: I already merged this as you had it. It's fine, just fix it up in your next one. I wanted to make sure we merged this split before I leave because it'll be easier for you to iterate on the settings stuff with it done. Thanks for the swift action :)

PR #672, opened 2023-06-22
Apply some fixes for flake8

dforsi:
This PR

- removes some unused imports and variables (such as ` SETTINGS_LISTS` and the `e` in `except Exception as e:`)
- removes a couple of duplicate methods
- fixes some escape sequences flagged as invalid (either by using raw strings of fixing the escape characters trying to be consistent with similar uses in other drivers)

 and makes 16 files "clean".


chirp/chirp_common.py
```diff
@@ -395,7 +395,7 @@ def format_freq(self):
     def parse_freq(self, freqstr):
         """Set the frequency from a string"""
         self.freq = parse_freq(freqstr)
-        return self.freqx
+        return self.freq
```
kk7ds, line 398: Yikes! Not sure this ever got used, but better that it's right. Looks like an errant typo on my part :(

chirp/drivers/icx90.py
```diff
@@ -602,10 +602,6 @@ def set_settings(self, settings):
     def process_mmap(self):
         self.memobj = bitwise.parse(ICX90_MEM_FORMAT, self._mmap)
 
-    def get_raw_memory(self, number):
```
kk7ds, line 606: Oh wow, this one is interesting... Maybe a bad merge at some point.

kk7ds:
This is a huge amount of work, thanks very much for doing this!

I'm fine with what is here, but I want to check with @KC9HI about the `SETTING_LISTS` things. Jim, you've added a lot of that. Were you just doing that for your own documentation or did those lists get used at some point somewhere? Perhaps just copy-and-paste stuff? They're not expected by any of the CHIRP infrastructure. Perhaps just a pattern that got copied from a driver that *was* using it?

KC9HI:
Mainly copy-and-paste. Whoever developed the initial uv5r.py driver had a SETTING_LISTS thing and it just got migrated to other drivers. I don't remember if I ever asked what they were for. I just assumed it was a "CHIRP" thing. I don't use them for anything. If CHIRP doesn't need it, I don't need it and would be happy to stop adding/maintaining them.

kk7ds:
Okay, that might've been me, so I'll take the blame :)

I'll rebase this and merge then, thanks for the confirmation!

PR #673, opened 2023-06-23
dupe() memory in set_memory() for memcache

W8BSD:
The memory passed to set_memory() is frozen, so if it's saved in memcache, the next time you try to change the memory, you'll get a lot of "fix this driver!" warnings from memedit and friends.


kk7ds:
> The memory passed to set_memory() is frozen, so if it's saved in memcache, the next time you try to change the memory, you'll get a lot of "fix this driver!" warnings from memedit and friends.

Yep, that was my trap to crowdsource fixing these. Thanks! :)

PR #674, opened 2023-06-23
Overhaul of Kenwood D7 Driver

W8BSD:
Removes dependency on kenwood_live, and implements a separate
radio class for D7 series radios.

Moves globals into the classes
- This allows finer control with subclassing

Clean up I/O
- Don't require a timeout after each command, speeding up access
  significantly
- Read all "garbage" bytes at the start rather than "just" 25

Add generic support for special channels
- Used for scan memories and call memories in the D7s

Makes settings easier to implement
- Data-driven description of settings with built-in support for
  bools, ints, strings, lists, and maps, which is enough for most
  settings
- Also supports custom classes for more complex settings, used for
  positions, DTMF memories, and programmable VFOs currently
- Allows for settings to depend on other settings
  Used for automatic simplex check and tone alert settings

Add full menu structure from the manuals.
- All settings listed in the manuals that can be controled via
  serial port are now exposed in Chirp

### DIFF
--- a/chirp/drivers/kenwood_d7.py
+++ b/chirp/drivers/kenwood_d7.py
@@ -13,332 +13,2319 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from chirp.drivers.kenwood_live import KenwoodOldLiveRadio, MODES, STEPS, \
-    DUPLEX, LOG, get_tmode
-from chirp import chirp_common, directory, util
+import logging
+import math
+import os
+import threading
+
+from chirp import chirp_common, directory, errors, util
 from chirp.settings import RadioSetting, RadioSettingGroup, \
     RadioSettingValueInteger, RadioSettingValueBoolean, \
-    RadioSettingValueString, RadioSettingValueList, RadioSettings
+    RadioSettingValueString, RadioSettingValueList, RadioSettingValueMap, \
+    RadioSettings, RadioSettingValueFloat
+
+LOG = logging.getLogger(__name__)
 
 
-@directory.register
-class THD7Radio(KenwoodOldLiveRadio):
-    """Kenwood TH-D7"""
-    MODEL = "TH-D7"
+class KenwoodD7Position(RadioSettingGroup):
+    def __init__(self, name, shortname, radio, *args, **kwargs):
+        super().__init__(name, shortname, *args, **kwargs)
+        self._radio = radio
+        if ' ' in name:
+            index = name.split(" ", 1)[1]
+        else:
+            index = ''
+        rs = RadioSetting("latd%s" % index, "Latitude degrees",
+                          RadioSettingValueInteger(0, 90, 0))
+        self.append(rs)
+        rs = RadioSetting("latm%s" % index, "Latitude minutes",
+                          RadioSettingValueFloat(0, 59.99, 0,
+                                                 resolution=0.01,
+                                                 precision=2))
+        self.append(rs)
+        rs = RadioSetting("south%s" % index, "South",
+                          RadioSettingValueBoolean(False))
+        self.append(rs)
+        rs = RadioSetting("longd%s" % index, "Longitude degrees",
+                          RadioSettingValueInteger(0, 180, 0))
+        self.append(rs)
+        rs = RadioSetting("longm%s" % index, "Longitude minutes",
+                          RadioSettingValueFloat(0, 59.99, 0,
+                                                 resolution=0.01, precision=2))
+        self.append(rs)
+        rs = RadioSetting("west%s" % index, "West",
+                          RadioSettingValueBoolean(False))
+        self.append(rs)
+
+    def kenwood_d7_read_value(self):
+        index = self._get_index()
+        name = self.get_name()
+        rawval = self._radio._kenwood_get(name)[1]
+        curr_value = int(rawval[0:2])
+        self["latd%s" % index].value.set_value(curr_value)
+        curr_value = int(rawval[2:6]) / 100
+        self["latm%s" % index].value.set_value(curr_value)
+        curr_value = int(rawval[6:8])
+        self["south%s" % index].value.set_value(curr_value)
+        curr_value = int(rawval[8:11])
+        self["longd%s" % index].value.set_value(curr_value)
+        curr_value = int(rawval[11:15]) / 100
+        self["longm%s" % index].value.set_value(curr_value)
+        curr_value = int(rawval[15:17])
+        self["west%s" % index].value.set_value(curr_value)
+        self['latd%s' % index].value._has_changed = False
+        self['latm%s' % index].value._has_changed = False
+        self['south%s' % index].value._has_changed = False
+        self['longd%s' % index].value._has_changed = False
+        self['longm%s' % index].value._has_changed = False
+        self['west%s' % index].value._has_changed = False
+
+    def _get_index(self):
+        name = self.get_name()
+        if ' ' in name:
+            return name.split(" ", 1)[1]
+        return ''
+
+    def changed(self):
+        index = self._get_index()
+        if self['latd%s' % index].changed():
+            return True
+        if self['latm%s' % index].changed():
+            return True
+        if self['south%s' % index].changed():
+            return True
+        if self['longd%s' % index].changed():
+            return True
+        if self['longm%s' % index].changed():
+            return True
+        if self['west%s' % index].changed():
+            return True
+        return False
+
+    def kenwood_d7_set_value(self):
+        index = self._get_index()
+        latd = self['latd%s' % index].value.get_value()
+        latm = self['latm%s' % index].value.get_value()
+        latmd = int(round(math.modf(latm)[0] * 100))
+        latm = int(round(latm))
+        south = int(self['south%s' % index].value.get_value())
+        longd = self['longd%s' % index].value.get_value()
+        longm = self['longm%s' % index].value.get_value()
+        longmd = int(round(math.modf(longm)[0] * 100))
+        longm = int(round(longm))
+        west = int(self['west%s' % index].value.get_value())
+        args = '%s%02d%02d%02d%02d%03d%02d%02d%02d' % (index,
+                                                       latd, latm, latmd,
+                                                       south,
+                                                       longd, longm, longmd,
+                                                       west)
+        if index != '':
+            index = ' ' + index
+        self._radio._kenwood_set('MP%s' % index, args)
+        # TODO: Hacking up _has_changed is bad... but we don't want
+        #       to re-send all the set commands each time either.
+        self['latd%s' % index].value._has_changed = False
+        self['latm%s' % index].value._has_changed = False
+        self['south%s' % index].value._has_changed = False
+        self['longd%s' % index].value._has_changed = False
+        self['longm%s' % index].value._has_changed = False
+        self['west%s' % index].value._has_changed = False
+
+
+class KenwoodD7DTMFMemory(RadioSettingGroup):
+    DTMF_CHARSET = "ABCD#*0123456789 "
+
+    def __init__(self, name, shortname, radio, *args, **kwargs):
+        super().__init__(name, shortname, *args, **kwargs)
+        self._radio = radio
+        self._name = name
+        index = self._get_index()
+        rs = RadioSetting("DMN %s" % index, "Memory Name",
+                          RadioSettingValueString(0, 8, '', False))
+        self.append(rs)
+        rs = RadioSetting("DM %s" % index, "Memory Value",
+                          RadioSettingValueString(0, 16, '', False,
+                                                  self.DTMF_CHARSET))
+        self.append(rs)
+
+    def _get_index(self):
+        return "%02d" % int(self._name.split(" ", 1)[1])
+
+    def kenwood_d7_read_value(self):
+        index = self._get_index()
+        value = self._radio._kenwood_get("DM %s" % index)[1]
+        value = value.replace("E", "*").replace("F", "#")
+        vname = self._radio._kenwood_get("DMN %s" % index)[1]
+        self["DMN %s" % index].value.set_value(vname)
+        self["DMN %s" % index].value._has_changed = False
+        self["DM %s" % index].value.set_value(value)
+        self["DM %s" % index].value._has_changed = False
+
+    def changed(self):
+        for element in self:
+            if element.changed():
+                return True
+        return False
+
+    def kenwood_d7_set_value(self):
+        for element in self:
+            if not element.changed():
+                continue
+            newval = element.value.get_value()
+            newval = newval.replace("*", "E").replace("#", "F")
+            self._radio._kenwood_set(element.get_name(),
+                                     newval)
+            element.value._has_changed = False
+
+
+class KenwoodD7ProgrammableVFOs(RadioSettingGroup):
+    def __init__(self, name, shortname, radio, *args, **kwargs):
+        self._radio = radio
+        super().__init__(name, shortname, *args, **kwargs)
+        for optname, index, minimum, maximum in self._radio._PROGRAMMABLE_VFOS:
+            group = RadioSettingGroup("PV %d" % index, optname)
+            self.append(group)
+            rs = RadioSetting("PV %dL" % index, "Lower Limit",
+                              RadioSettingValueInteger(minimum, maximum,
+                                                       minimum))
+            group.append(rs)
+            rs = RadioSetting("PV %dU" % index, "Upper Limit",
+                              RadioSettingValueInteger(minimum, maximum,
+                                                       maximum))
+            group.append(rs)
+
+    def _get_index(self, element):
+        name = element.get_name()
+        return int(name.split(" ", 1)[1])
+
+    def kenwood_d7_read_value(self):
+        for element in self:
+            index = self._get_index(element)
+            rv = self._radio._kenwood_get("PV %d" % index)[1]
+            lower, upper = rv.split(',', 1)
+            lower = int(lower)
+            upper = int(upper)
+            pvdl = "PV %dL" % index
+            pvdu = "PV %dU" % index
+            element[pvdl].value.set_value(lower)
+            element[pvdl].value._has_changed = False
+            element[pvdu].value.set_value(upper)
+            element[pvdu].value._has_changed = False
+
+    def changed(self):
+        for element in self:
+            for bound in element:
+                if bound.changed():
+                    return True
+        return False
+
+    # TODO: Custom validator things...
+
+    def kenwood_d7_set_value(self):
+        for element in self:
+            index = self._get_index(element)
+            pvdl = "PV %dL" % index
+            pvdu = "PV %dU" % index
+            if element[pvdl].changed() or element[pvdu].changed():
+                lower = element[pvdl].value.get_value()
+                upper = element[pvdu].value.get_value()
+                args = "%05d,%05d" % (lower, upper)
+                self._radio._kenwood_set(element.get_name(), args)
+                element[pvdl].value._has_changed = False
+                element[pvdu].value._has_changed = False
+
+
+class KenwoodD7Family(chirp_common.LiveRadio):
+    VENDOR = "Kenwood"
+    MODEL = ""
+    NEEDS_COMPAT_SERIAL = False
     HARDWARE_FLOW = False
 
-    _kenwood_split = True
-    _upper = 199
+    _ARG_DELIMITER = " "
+    _BAUDS = [9600]
+    _CMD_DELIMITER = "\r"
+    _DISABLE_AI_CMD = ('AI', '0')
+    _DUPLEX = {0: "", 1: "+", 2: "-"}
+    _HAS_NAME = True
+    _LOCK = threading.Lock()
+    _MODES = {0: "FM", 1: "AM"}
+    _NOCACHE = "CHIRP_NOCACHE" in os.environ
+    _SPLIT = True
+    _STEPS = (5.0, 6.25, 10.0, 12.5, 15.0, 20.0, 25.0, 30.0, 50.0, 100.0)
+    _TONES = chirp_common.OLD_TONES
+    _LOWER = 0
+    _UPPER = 199
 
-    _BEP_OPTIONS = ["Off", "Key", "Key+Data", "All"]
-    _POSC_OPTIONS = ["Off Duty", "Enroute", "In Service", "Returning",
-                     "Committed", "Special", "Priority", "Emergency"]
+    _CALL_CHANS = ("VHF Call", "UHF Call")
+    _SPECIAL_CHANS = ("L0", "U0",
+                      "L1", "U1",
+                      "L2", "U2",
+                      "L3", "U3",
+                      "L4", "U4",
+                      "L5", "U5",
+                      "L6", "U6",
+                      "L7", "U7",
+                      "L8", "U8",
+                      "L9", "U9",
+                      *_CALL_CHANS)
 
-    _SETTINGS_OPTIONS = {
-        "BAL": ["4:0", "3:1", "2:2", "1:3", "0:4"],
-        "BEP": None,
-        "BEPT": ["Off", "Mine", "All New"],  # D700 has fourth "All"
-        "DS": ["Data Band", "Both Bands"],
-        "DTB": ["A", "B"],
-        "DTBA": ["A", "B", "A:TX/B:RX"],  # D700 has fourth A:RX/B:TX
-        "DTX": ["Manual", "PTT", "Auto"],
-        "ICO": ["Kenwood", "Runner", "House", "Tent", "Boat", "SSTV",
-                "Plane", "Speedboat", "Car", "Bicycle"],
-        "MNF": ["Name", "Frequency"],
-        "PKSA": ["1200", "9600"],
-        "POSC": None,
-        "PT": ["100ms", "200ms", "500ms", "750ms",
-               "1000ms", "1500ms", "2000ms"],
-        "SCR": ["Time", "Carrier", "Seek"],
-        "SV": ["Off", "0.2s", "0.4s", "0.6s", "0.8s", "1.0s",
-               "2s", "3s", "4s", "5s"],
-        "TEMP": ["F", "C"],
-        "TXI": ["30sec", "1min", "2min", "3min", "4min", "5min",
-                "10min", "20min", "30min"],
-        "UNIT": ["English", "Metric"],
-        "WAY": ["Off", "6 digit NMEA", "7 digit NMEA", "8 digit NMEA",
-                "9 digit NMEA", "6 digit Magellan", "DGPS"],
-    }
+    # These are used more than once in _COMMON_SETTINGS
+    _BOOL = {'type': 'bool'}
+    _DATA_BAND = {'type': 'list',
+                  'values': ("A", "B", "A:TX/B:RX", "A:RX/B:TX")}
+    _DTMF_MEMORY = {'type': KenwoodD7DTMFMemory}
+    _ON_OFF_DUAL_MAP = {'type': 'map',
+                        'map': (("Off", "0"), ("On", "1,0")),
+                        'set_requires': (('DL', True),),
+                        'get_requires': (('DL', True),)}
+    _POSITION = {'type': KenwoodD7Position}
+    _SSTV_CHARSET = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789 !?-/"
+    _SSTV_COLOURS = {'type': 'list',
+                     'values': ("Black", "Blue", "Red", "Magenta", "Green",
+                                "Cyan", "Yellow", "White")}
+    _STATUS_TEXT = {'type': 'string', 'max_len': 32}
+    _SQUELCH = {'type': 'list', 'digits': 2,
+                'values': ("Open", "1", "2", "3", "4", "5")}
+    # Settings that are shared by at least two radios.
+    # If a third radio does not have the setting at all, the presence
+    # here won't matter.  If it has it but with differrent parameters,
+    # it can be overridden in the subclass _SETTINGS dict.
+    _COMMON_SETTINGS = {
+        "AIP": _BOOL,
+        "ARL": {'type': 'map',
+                'map': (('Off', '0000'),) + tuple(
+                    [('%i' % x, '%04i' % x)
+                     for x in range(10, 2510, 10)])},
+        "ARLM": {'type': 'string', 'max_len': 45},
+        "AMGG": {'type': 'string', 'max_len': 45,
+                 'charset': "ABCDEFGHIJKLMNOPQRSTUVWXYZ*,-0123456789"},
+        "AMR": _BOOL,
+        "APO": {'type': 'list',
+                'values': ("Off", "30min", "60min")},
+        "ARO": _BOOL,
+        # ASC x requires that the specified VFO is enabled on TH-D7G
+        # (ie: dual is on or VFO is current) for both fetch and set
+        # Easiest to just enable DL before set/read
+        "ASC 0": _ON_OFF_DUAL_MAP,
+        "ASC 1": _ON_OFF_DUAL_MAP,
+        "BAL": {'type': 'list',
+                'values': ("4:0", "3:1", "2:2", "1:3", "0:4")},
+        "BC": {'type': 'list',
+               'values': ("A", "B")},
+        "BCN": _BOOL,
+        # BEL x requires that the specified VFO is enabled on TH-D7G
+        # (ie: dual is on or VFO is current) for both fetch and set
+        # Easiest to just enable DL before set/read
+        "BEL 0": _ON_OFF_DUAL_MAP,
+        "BEL 1": _ON_OFF_DUAL_MAP,
+        "BEP": {'type': 'list',
+                'values': ("Off", "Key", "Key+Data", "All")},
+        "BEPT": {'type': 'list',
+                 'values': ("Off", "Mine", "All New", "All")},
+        "CH": _BOOL,
+        "CKEY": {'type': 'list',
+                 'values': ("Call", "1750Hz")},
+        "CNT": {'type': 'integer', 'min': 1, 'max': 16},
+        "DL": _BOOL,
+        "DS": {'type': 'list',
+               'values': ("Data Band", "Both Bands")},
+        "DTB": _DATA_BAND,
+        "DTBA": _DATA_BAND,
+        "dtmfmem 00": _DTMF_MEMORY,
+        "dtmfmem 01": _DTMF_MEMORY,
+        "dtmfmem 02": _DTMF_MEMORY,
+        "dtmfmem 03": _DTMF_MEMORY,
+        "dtmfmem 04": _DTMF_MEMORY,
+        "dtmfmem 05": _DTMF_MEMORY,
+        "dtmfmem 06": _DTMF_MEMORY,
+        "dtmfmem 07": _DTMF_MEMORY,
+        "dtmfmem 08": _DTMF_MEMORY,
+        "dtmfmem 09": _DTMF_MEMORY,
+        "DTX": {'type': 'list',
+                'values': ("Manual", "PTT", "Auto")},
+        "ELK": {'type': 'bool'},
+        "GU": {'type': 'list',
+               'values': ("Not Used", "NMEA", "NMEA96")},
+        # This isn't exactly shared, but each radio adds to this list
+        "ICO": {'type': 'map',
+                'map': (('Kenwood', '0,0'),
+                        ('Runner', '0,1'),
+                        ('House ', '0,2'),
+                        ('Tent', '0,3'),
+                        ('Boat', '0,4'),
+                        ('SSTV', '0,5'),
+                        ('Plane', '0,6'),
+                        ('Speedboat', '0,7'),
+                        ('Car', '0,8'),
+                        ('Bicycle', '0,9'),
+                        ('TRIANGLE(DF station)', '0,A'),
+                        ('Jeep', '0,B'),
+                        ('Recreational Vehicle', '0,C'),
+                        ('Truck ', '0,D'),
+                        ('Van', '0,E'))},
+        "KILO": {'type': 'list',
+                 'values': ("Miles", "Kilometers")},
+        "LK": _BOOL,
+        "LMP": _BOOL,
+        "MAC": _SSTV_COLOURS,
+        "MES": {'type': 'string', 'max_len': 8},
+        # NOTE: Requires memory mode to be active
+        "MNF": {'type': 'list',
+                'values': ("Name", "Frequency")},
+        "MP 1": _POSITION,
+        "MP 2": _POSITION,
+        "MP 3": _POSITION,
+        "MYC": {'type': 'string', 'max_len': 9},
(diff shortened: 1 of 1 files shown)


W8BSD:
With the bare import of kenwood_live apparently needed to modify LAST_BAUD, I'm seriously considering disconnecting this driver from kenwood_live completely.  At this point the amount of shared code is low enough that I'm not even really sure how much there even is left.

W8BSD:
> I'm seriously considering disconnecting this driver from kenwood_live completely.

Done.

W8BSD:
I'm actually wondering if this should be renamed kenwood_newlive.py or something... I'm starting to consider doing my TS-B2000 and maybe the TS-940S with this driver.  Using the menus on the B2000 is a huge pain and since it's always connected to my computer, it would be nice to be able to set it up using Chirp instead.  The 940 I'm not as sure about, I tend to use HF memories a lot differently than I do VHF+ ones.

Likely a change for a follow-on commit though.

W8BSD:
TS-2000 conversion in my [TS-B2000-Update)](https://github.com/W8BSD/chirp/tree/TS-B2000-Update) branch.

W8BSD:
Ok, I think that's "done" now barring TM-D700 testing, It's now enough to support the wildly different TS-2000, so it seems like the features are nailed down nicely.  I'll wait to mark it as ready to review until I actually write a useful commit message.

W8BSD:
I've also started the work splitting out generic functionality... first was pulling out the setting "stuff" into a separate mixin (which is certainly in the wrong place) in [this branch](https://github.com/W8BSD/chirp/tree/Settings-Mixin)

kk7ds:
Okay I did a test of this with my D700 and got a few issues. I created a bug which you can use to track this work:

https://chirp.danplanet.com/issues/10683

and attached my debug logs there for each issue. Note that without this patch applied I can paste the same memories and they all get into the radio (albeit much slower).

Another thing - the radio starts with memory zero with the patch applied, memory one is the actual first channel in the radio (as with the previous driver).

W8BSD:
Maybe fixed now.

kk7ds:
Not sure if you were looking at the memory issue or the settings one. Memory issue remains:
```
DEBUG: PC->RADIO: MW 0,0,156,00462725000,0,1,0,0,1,00,09,0010,05,005000000,0,0
DEBUG: RADIO->PC: 'N'
ERROR: Failed to run <RadioJob@0>set_memory(Memory 156: 462.725000+5.000000 FM (Mt Tabor) r88.5  c77.0* d023 NN [5.00],)=None
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1159, in set_memory
    raise errors.InvalidDataError("Radio refused %s" % memid)
chirp.errors.InvalidDataError: Radio refused 156
DEBUG: Radio finished <RadioJob@0>set_memory(Memory 156: 462.725000+5.000000 FM (Mt Tabor) r88.5  c77.0* d023 NN [5.00],)=InvalidDataError('Radio refused 156')
```
Settings is different but not better:
```
ERROR: Failed to run <RadioJob@10>get_settings(,)=None
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1144, in get_settings
    ret = self._create_group(self._SETTINGS_MENUS)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 638, in _create_group
    grp.append(self._create_group(entry[1], subgrp))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 638, in _create_group
    grp.append(self._create_group(entry[1], subgrp))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 674, in _create_group
    grp.append(rs)
  File "/Users/dan/Documents/chirp.git/chirp/settings.py", line 367, in append
    self[element.get_name()] = element
  File "/Users/dan/Documents/chirp.git/chirp/settings.py", line 403, in __setitem__
    raise KeyError("Duplicate item %s" % name)
KeyError: 'Duplicate item SSQ 0'
```

W8BSD:
> Not sure if you were looking at the memory issue or the settings one.

Both of course, both attempted to be fixed again, and force pushed to cb9f6b5b

kk7ds:
Nice, looks a lot better now. I was able to paste in all my channels and they went in no problem this time.

Settings loads and I toggled a whole bunch of them without any trouble. Once I toggled the "channel display" I get this:
```
ERROR: Failed to run <RadioJob@0>set_settings([<chirp.settings.RadioSettingGroup object at 0x1522fa3b0>, <chirp.settings.RadioSettingGroup object at 0x1522f90c0>, <chirp.settings.RadioSettingGroup object at 0x1522f9a50>, <chirp.settings.RadioSettingGroup object at 0x14fe5db70>, <chirp.settings.RadioSettingGroup object at 0x14fe5d4b0>],)=None
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1182, in set_settings
    self.set_settings(element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1182, in set_settings
    self.set_settings(element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1186, in set_settings
    self._set_setting(name, sdata, element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1039, in _set_setting
    self._kenwood_set_bool(name, value)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 865, in _kenwood_set_bool
    return self._kenwood_set(cmd, str(int(value)))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 862, in _kenwood_set
    raise errors.RadioError("Radio refused to set %s" % cmd)
chirp.errors.RadioError: Radio refused to set CH
```
But, it *does* work.. the radio changes to channel display mode but I guess the response isn't what you expect.

But yeah man, awesome and really quick.

Just tried one other thing, which is a real hard thing for live radios to do... I selected two memories and then *Edit > Move Up* and get this:

```
DEBUG: Transplanting Memory 9: 0.000000/0.600000 FM () r88.5  c88.5  d023 NN [5.00] to 11 for move indexes: [9, 10] -> [8, 9]
DEBUG: Moving Memory 10: 147.120000+0.600000 FM (Timberli) r103.5  c100.0* d023 NN [5.00] to 9
DEBUG: Moving Memory 11: 444.225000+5.000000 FM (Timberli) r88.5  c100.0* d023 NN [5.00] to 10
DEBUG: PC->RADIO: MW 0,0,011
DEBUG: RADIO->PC: 'MW'
DEBUG: PC->RADIO: MNA 0,011,
DEBUG: RADIO->PC: 'N'
ERROR: Failed to run <RadioJob@0>set_memory(Memory 11: 0.000000/0.600000 FM () r88.5  c88.5  d023 NN [5.00],)=None
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1166, in set_memory
    raise errors.InvalidDataError("Radio refused name %s: %s" %
chirp.errors.InvalidDataError: Radio refused name 011: ''
DEBUG: Radio finished <RadioJob@0>set_memory(Memory 11: 0.000000/0.600000 FM () r88.5  c88.5  d023 NN [5.00],)=InvalidDataError("Radio refused name 011: ''")
```

But it seemed like it otherwise worked in the radio. If I close and re-open it looks right I think. Move up/down causes a bunch of get, delete, and set operations to happen all at once. Does that work on your D7?

Note I haven't even looked at the code yet, just testing in between other tasks.

W8BSD:
> But, it _does_ work.. the radio changes to channel display mode but I guess the response isn't what you expect.

Curious what debug logging shows is sent/received there.

> Just tried one other thing, which is a real hard thing for live radios to do... I selected two memories and then Edit > Move Up and get this:

It didn't even occur to me to try that, I'll take a look.  I'll test/hack some more tonight.

W8BSD:
Pushed ba5863f with a workaround for the memory move issue, it appears that set_memory() is being passed a memory where memory.empty == True.  I would expect erase_memory() to be used for that... just have set_memory() calling erase_memory() and debug logging now.

W8BSD:
One issue I'm seeing is when I try to copy from a regular memory to a special memory (VHF call in this case), the overwrite prompt uses the number, not the extd_number, but the extd_number seems to clobber the number for the target, which fails, and I get this:

```python
ERROR: Failed to paste: '<' not supported between instances of 'str' and 'int'
  File "/usr/home/admin/src/chirp/chirp/wxui/memedit.py", line 1761, in _cb_paste_memories
    mem = import_logic.import_mem(self._radio, srcrf, mem)
  File "/usr/home/admin/src/chirp/chirp/import_logic.py", line 263, in import_mem
    msgs = dst_radio.validate_memory(dst_mem)
  File "/usr/home/admin/src/chirp/chirp/chirp_common.py", line 1231, in validate_memory
    return rf.validate_memory(mem)
  File "/usr/home/admin/src/chirp/chirp/chirp_common.py", line 980, in validate_memory
    (mem.number < lo or mem.number > hi) and \
```

Not sure if that's a driver issue, or a special memory issue.

kk7ds:
Yeah, so `erase_memory()` is older and basically everyone expects to just get a memory with empty set to true nowadays. However, there are some live drivers that really need to have erase be separate. There might be something in the UI that is just doing set and not erase, so I can look into that. I'll also look into the special memory thing and test the latest, but it'll be tomorrow.

Thanks for your work on these!

W8BSD:
Just force-pushed 33432e2 with a theoretical fix for the CH issue... added an RX drain before sending a command so if a command sends multiple responses, or the driver state gets messed up, it can at least be sure it's the right response for the command that was sent.

Really, just a theory as to to may have happened, but it's a good thing to have in there regardless.

kk7ds:
Okay, latest change fixes the bulk moves, so that's sweet. It doesn't fix the CH thing but does make it better. I still get the "NAKed" error when I enable CH mode, but not when I disable it and it seems to recover fine. Before it would continue giving me the failure even after I re-disabled it and went on to other settings.

I also started it loading memories and then switched to settings to make loading settings and memories fight it out and everything seemed to be happy - definitely nice.

The paste-into-special problem is not related to your driver, and is also Gnarly (capital G). I'll have to chew on that one for a bit.

W8BSD:
I'm really curious what (if anything) the radio is sending in response to the `CH 1`... if it's simply not replying, I can just tack on an ID command to force a response and deal with it, if it's replying `N` or `?` it's a bit weirder and will need a think, and if it's doing something like a `UMC 2` it should be fixable with the prerequisite stuff... and I bet it will work properly if the radio is in memory mode already.

kk7ds:
Sorry, should have included a log snippet:
```
DEBUG: PC->RADIO: CH 1
DEBUG: RADIO->PC: 'SM 0,00'
ERROR: Failed to run <RadioJob@0>set_settings([<chirp.settings.RadioSettingGroup object at 0x1201e19f0>, <chirp.settings.RadioSettingGroup object at 0x1201e1ae0>, <chirp.settings.RadioSettingGroup object at 0x1201e17b0>, <chirp.settings.RadioSettingGroup object at 0x1174f4c70>, <chirp.settings.RadioSettingGroup object at 0x1174f46d0>],)=None
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1191, in set_settings
    self.set_settings(element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1191, in set_settings
    self.set_settings(element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1195, in set_settings
    self._set_setting(name, sdata, element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1043, in _set_setting
    self._kenwood_set_bool(name, value)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 869, in _kenwood_set_bool
    return self._kenwood_set(cmd, str(int(value)))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 866, in _kenwood_set
    raise errors.RadioError("Radio refused to set %s" % cmd)
chirp.errors.RadioError: Radio refused to set CH
DEBUG: Radio finished <RadioJob@0>set_settings([<chirp.settings.RadioSettingGroup object at 0x1201e19f0>, <chirp.settings.RadioSettingGroup object at 0x1201e1ae0>, <chirp.settings.RadioSettingGroup object at 0x1201e17b0>, <chirp.settings.RadioSettingGroup object at 0x1174f4c70>, <chirp.settings.RadioSettingGroup object at 0x1174f46d0>],)=RadioError('Radio refused to set CH')
INFO: Editor status: Saved settings
```

W8BSD:
That's *very* interesting, SM is the S-Meter reading, implying that AI mode got enabled somehow... do you know if squelch was open at the time?

W8BSD:
I've pushed a change the specifically ignores "SM " lines, but I suspect that either AI is briefly enabled, so pretty much anything can be received.

kk7ds:
Same deal with 10ceafdc:
```
DEBUG: PC->RADIO: CH 1
DEBUG: RADIO->PC: 'SM 0,00'
ERROR: Failed to run <RadioJob@0>set_settings([<chirp.settings.RadioSettingGroup object at 0x155d09a50>, <chirp.settings.RadioSettingGroup object at 0x155d09840>, <chirp.settings.RadioSettingGroup object at 0x155d09690>, <chirp.settings.RadioSettingGroup object at 0x14a4f4cd0>, <chirp.settings.RadioSettingGroup object at 0x14a4f4730>],)=None
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1197, in set_settings
    self.set_settings(element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1197, in set_settings
    self.set_settings(element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1201, in set_settings
    self._set_setting(name, sdata, element)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1049, in _set_setting
    self._kenwood_set_bool(name, value)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 875, in _kenwood_set_bool
    return self._kenwood_set(cmd, str(int(value)))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 872, in _kenwood_set
    raise errors.RadioError("Radio refused to set %s" % cmd)
chirp.errors.RadioError: Radio refused to set CH
DEBUG: Radio finished <RadioJob@0>set_settings([<chirp.settings.RadioSettingGroup object at 0x155d09a50>, <chirp.settings.RadioSettingGroup object at 0x155d09840>, <chirp.settings.RadioSettingGroup object at 0x155d09690>, <chirp.settings.RadioSettingGroup object at 0x14a4f4cd0>, <chirp.settings.RadioSettingGroup object at 0x14a4f4730>],)=RadioError('Radio refused to set CH')
INFO: Editor status: Saved settings
```
No S-Meter deflection for sure.

W8BSD:
Heh...
```python
if result[0:2] == 'SM ':
```

Unlikely... fixed.

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
+LOG = logging.getLogger(__name__)
 
-@directory.register
-class THD7Radio(KenwoodOldLiveRadio):
-    """Kenwood TH-D7"""
-    MODEL = "TH-D7"
+
+# It would be nice to have this as an array so it's grouped in the right
+# Unfortunately, when we do that, we can't have labels for each line
+class KenwoodD7Position(RadioSettingGroup):
```
kk7ds, line 32: These aren't usually subclassed like this. It's not really a problem and they generally don't change much, but none of the other drivers do it (that I know of) so it kinda puts you in being-different-land. The usage is pretty straightforward, so I don't anticipate problems, but it's also one of those things where we won't know until someone tries to use it because it's a live driver. So just FYI...

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
-        rf.valid_tuning_steps = STEPS
-        rf.memory_bounds = (0, self._upper)
-        return rf
+    _ARG_DELIMITER = " "
+    _BAUDS = [9600]
+    _CMD_DELIMITER = "\r"
+    _DISABLE_AI_CMD = ('AI', '0')
+    _DUPLEX = {0: "", 1: "+", 2: "-"}
+    _HAS_NAME = True
+    _LOCK = threading.Lock()
```
kk7ds, line 254: Is this necessary? The RadioThread should be handling dispatching one request at a time to the driver...

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
+                        ("2220", "2220"), ("2230", "2230"), ("2240", "2240"),
+                        ("2250", "2250"), ("2260", "2260"), ("2270", "2270"),
+                        ("2280", "2280"), ("2290", "2290"), ("2300", "2300"),
+                        ("2310", "2310"), ("2320", "2320"), ("2330", "2330"),
+                        ("2340", "2340"), ("2350", "2350"), ("2360", "2360"),
+                        ("2370", "2370"), ("2380", "2380"), ("2390", "2390"),
+                        ("2400", "2400"), ("2410", "2410"), ("2420", "2420"),
+                        ("2430", "2430"), ("2440", "2440"), ("2450", "2450"),
+                        ("2460", "2460"), ("2470", "2470"), ("2480", "2480"),
+                        ("2490", "2490"), ("2500", "2500"))},
```
kk7ds, line 383: We could probably write a generator to make this in a lot fewer lines of code. Looks like it's 0-2500 in steps of ten? Something like:
```
[('Off', '0000')] + [('%i' % x, '%04i' % x) for x in range(10, 2510, 10)]
```

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
+                         ("94.8", "11"), ("97.4", "12"), ("100.0", "13"),
+                         ("103.5", "14"), ("107.2", "15"), ("110.9", "16"),
+                         ("114.8", "17"), ("118.8", "18"), ("123.0", "19"),
+                         ("127.3", "20"), ("131.8", "21"), ("136.5", "22"),
+                         ("141.3", "23"), ("146.2", "24"), ("151.4", "25"),
+                         ("156.7", "26"), ("162.2", "27"), ("167.9", "28"),
+                         ("173.8", "29"), ("179.9", "30"), ("186.2", "31"),
+                         ("192.8", "32"), ("203.5", "33"), ("210.7", "34"),
+                         ("218.1", "35"), ("225.7", "36"), ("233.6", "37"),
+                         ("241.8", "38"), ("250.3", "39"))},
```
kk7ds, line 481: Similar here from `KENWOOD_TONES`?

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
+    _PROGRAMMABLE_VFOS = (
+        ("Band A, 118MHz Sub-Band", 1, 118, 135),
+        ("Band A, VHF Sub-Band", 2, 136, 173),
+        ("Band B, VHF Sub-Band", 3, 144, 147),
+        ("Band B, UHF Sub-Band", 6, 400, 479))
+
+    _baud = 9600
+    _cur_setting = 0
+    _memcache = {}
+    _setcache = {}
```
kk7ds, line 544: It's a bad idea to do this (existing examples in the code notwithstanding). This is a mutable-type variable on the class, which is shared amongst all instances of this class until you do something like L552. better to leave these out and depend on the init code to create them so they never get accidentally shared.

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
+            sdata = self._get_setting_data(prereq)
+            self._set_setting(prereq, sdata, entry)
+
+    def _drain_input(self):
+        oldto = self.pipe.timeout
+        self.pipe.timeout = 0
+        # Read until a timeout
+        while len(self.pipe.read(1)):
+            pass
+        self.pipe.timeout = oldto
```
kk7ds, line 713: So, FYI, on Windows changing the timeout apparently clears the input buffer. I can test this on windows too to be sure, but just be super careful about when you do this when there might be data pending you want to read.

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
+        if len(spec) == 14:  # Call memory
+            mem.offset = int(spec[2])
+        else:
+            mem.offset = int(spec[3])
+
+    def _refresh_setting(self, entry):
+        name = entry.get_name()
+        setting = self._get_setting_data(name)
+        # TODO: It would be nice to bump_wait_dialog here...
+        #       Also, this is really only useful for the cli. :(
```
kk7ds, line 994: Yeah, unfortunately, .. hmm, what if you make this a generator? I might be able to make the UI iterate that and do the bumping for you as you read the settings out.

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
-        for setting, group, name in lists:
-            value = self._kenwood_get_int(setting)
-            options = self._get_setting_options(setting)
-            rs = RadioSetting(setting, name,
-                              RadioSettingValueList(options,
-                                                    options[value]))
-            group.append(rs)
+    def set_memory(self, memory):
+        index, memid = self._validate_memory(memory)
+        if (memory.empty):
```
kk7ds, line 1171: No need for parens here.

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3404 @@
...
-        "BEPT": ["Off", "Mine", "All New", "All"],  # Added "All"
-    }
+    _SETTINGS = {
+        "DS": {'type': 'list',
+               'values': ("Data Band", "Both Bands", "Ignore DCD")},
+        "DSPA": {'type': 'list',
+                 'values': ("Entire Display", "One Line")},
+        "ICO": {'type': 'map',
+                'map': (('Kenwood', '0,0'),
+                        ('Runner', '0,1'),
```
kk7ds, line 1338: Don't you want to re-use as much of this as you can from the sibling?

kk7ds:
I'll test this updated one against the radio in a bit..

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,3384 @@
...
 
+LOG = logging.getLogger(__name__)
 
-@directory.register
-class THD7Radio(KenwoodOldLiveRadio):
-    """Kenwood TH-D7"""
-    MODEL = "TH-D7"
+
+# It would be nice to have this as an array so it's grouped in the right
+# Unfortunately, when we do that, we can't have labels for each line
```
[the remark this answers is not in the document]
W8BSD: Yeah, that's what I figured, I'll remove the comment since it's fine the way it is.

chirp/drivers/kenwood_d7.py
W8BSD: Ditto

chirp/drivers/kenwood_d7.py
W8BSD: I only have it there because it was in there before (and is still in kenwood_live).  Happy to remove it.

kk7ds:
Probably just a typo, but from f2a996ae:

```
|          | 1.0% Fetching Dual                          DEBUG: PC->RADIO: DL
DEBUG: RADIO->PC: 'DL 1'
DEBUG: PC->RADIO: DL 1
DEBUG: RADIO->PC: 'DL 1'
ERROR: Failed to run <RadioJob@10>get_settings(,)=None
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1161, in get_settings
    self._refresh_settings()
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1029, in _refresh_settings
    self._refresh_setting(entry)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1017, in _refresh_setting
    value = self._kenwood_get(name)[1]
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 837, in _kenwood_get
    self._do_prerequisite(cmd, True, True)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 705, in _do_prerequisite
    self._set_setting(prereq, sdata, entry)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1054, in _set_setting
    self._kenwood_set_bool(name, value)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 880, in _kenwood_set_bool
    return self._kenwood_set(cmd, str(int(value)))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 875, in _kenwood_set
    if self._kenwood_set_success(cmd, modcmd, value, response):
NameError: name 'response' is not defined
DEBUG: Radio finished <RadioJob@10>get_settings(,)=NameError("name 'response' is not defined")
INFO: Editor status: Retrieved settings
ERROR: Context raised unexpected_exception
  File "/Users/dan/Documents/chirp.git/chirp/wxui/settingsedit.py", line 47, in _initialize
    raise job.result
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1161, in get_settings
    self._refresh_settings()
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1029, in _refresh_settings
    self._refresh_setting(entry)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1017, in _refresh_setting
    value = self._kenwood_get(name)[1]
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 837, in _kenwood_get
    self._do_prerequisite(cmd, True, True)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 705, in _do_prerequisite
    self._set_setting(prereq, sdata, entry)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1054, in _set_setting
    self._kenwood_set_bool(name, value)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 880, in _kenwood_set_bool
    return self._kenwood_set(cmd, str(int(value)))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 875, in _kenwood_set
    if self._kenwood_set_success(cmd, modcmd, value, response):
NameError: name 'response' is not defined
  File "/opt/homebrew/lib/python3.10/site-packages/wx/core.py", line 3427, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File "/Users/dan/Documents/chirp.git/chirp/wxui/settingsedit.py", line 47, in _initialize
    raise job.result
  File "/Users/dan/Documents/chirp.git/chirp/wxui/radiothread.py", line 72, in dispatch
    self.result = getattr(radio, self.fn)(*self.args, **self.kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1161, in get_settings
    self._refresh_settings()
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1029, in _refresh_settings
    self._refresh_setting(entry)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1017, in _refresh_setting
    value = self._kenwood_get(name)[1]
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 837, in _kenwood_get
    self._do_prerequisite(cmd, True, True)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 705, in _do_prerequisite
    self._set_setting(prereq, sdata, entry)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 1054, in _set_setting
    self._kenwood_set_bool(name, value)
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 880, in _kenwood_set_bool
    return self._kenwood_set(cmd, str(int(value)))
  File "/Users/dan/Documents/chirp.git/chirp/drivers/kenwood_d7.py", line 875, in _kenwood_set
    if self._kenwood_set_success(cmd, modcmd, value, response):
NameError: name 'response' is not defined
```

chirp/drivers/kenwood_d7.py
kk7ds: Ah, sorry, it's been a long time since I looked at this stuff. Should be good to remove it.

chirp/drivers/kenwood_d7.py
W8BSD: Perfect, I'm not a python developer, so I didn't think of that.

chirp/drivers/kenwood_d7.py
W8BSD: Sounds good, not being a python developer, these are the type of nuances I'm unaware of.

chirp/drivers/kenwood_d7.py
W8BSD: That's fine here, since clearing the input buffer is the intent, but thanks for the heads-up.

chirp/drivers/kenwood_d7.py
W8BSD: Damn, I thought I managed to catch myself on all of these, thanks.

chirp/drivers/kenwood_d7.py
kk7ds: Yep, it's not something I did a very good job of guarding against in the early days, despite being a python guy. It's obscure for sure, but learn from my scars :)

chirp/drivers/kenwood_d7.py
kk7ds: Let me play with this idea (tomorrow) and if it seems good I'll show you the python magic to make this effectively a coroutine that generates pieces of the list as the UI consumes them, thus avoiding hanging up the UI while you pull these all out.

chirp/drivers/kenwood_d7.py
W8BSD: I do, but I don't actually know how to do it well... the only difference between the D7G and the D700 is a single extra entry in the D700.  I'm not too worried about the D7 since it doesn't have very many options.

Basically, I'm not sure how/when storage is allocated for these, and I'm not sure how to grab class values from other classes.

chirp/drivers/kenwood_d7.py
W8BSD: Yeah, I'll take a look.

chirp/drivers/kenwood_d7.py
kk7ds: It's not really a big deal, it's just a lot of duplication. I think you should be able to do something like:
```
'map': KenwoodD7._SETTINGS['ICO']['map'] + (('WEATHER Station (blue)', '1,/#'), ...)
```

chirp/drivers/kenwood_d7.py
W8BSD: Turns out it's needed so I can click the settings tab while the memories are loading.

chirp/drivers/kenwood_d7.py
W8BSD: Oh, this is about the RadioSetting subclasses, not the comment... :smile: 

Yeah, this seemed like the cleanest way to bundle all the code together and keep the _SETTINGS dict as the only thing the individual drivers needed.  I've actually made the settings "stuff" a mixin that I'll be using for a TS-2000 overhaul I've hacked up that I'm very happy with (which is also much faster, and has all the menu settings).  I'll try to get as many live radios using it as possible so it's noticed quickly. :wink:

chirp/drivers/kenwood_d7.py
W8BSD: This one was tricky...

```python
        "SKTN": {'type': 'map',
                 'map': tuple([(str(x), "%02d" % (ind + 1))
                               for ind, x in enumerate(_TONES) if x != 69.3])},
```

W8BSD:
> Probably just a typo, but from [f2a996a](https://github.com/kk7ds/chirp/commit/f2a996ae6fefd07bed00c6120eb925cee4e164ff):

Fixed, thanks.

chirp/drivers/kenwood_d7.py
W8BSD: Done

chirp/drivers/kenwood_d7.py
W8BSD: Done.

W8BSD:
I'm really not sure how the style checker wants me to indent those lines it's flagging... it *looks* properly visually indented to me...

chirp/drivers/kenwood_d7.py
```diff
@@ -13,332 +13,2319 @@
...
+                'values': ("Open", "1", "2", "3", "4", "5")}
+    # Settings that are shared by at least two radios.
+    # If a third radio does not have the setting at all, the presence
+    # here won't matter.  If it has it but with differrent parameters,
+    # it can be overridden in the subclass _SETTINGS dict.
+    _COMMON_SETTINGS = {
+        "AIP": _BOOL,
+        "ARL": {'type': 'map',
+                'map': (('Off', '0000'),) +
+                       tuple([('%i' % x, '%04i' % x)
```
kk7ds, line 297: I think it's complaining because it thinks this line belongs to the same level as `'map'`, but that would look confusing and weird, so don't do that. Convention would be to put the `tuple(` at the end of the line above and then go in four spaces and I think it'll be happy:
```
'map': (('Off', '0000'),) + tuple(
    [('%i' % x, '%04i' % x)
     for x ...
```

W8BSD: Yeah, that takes care of that one, but the other two aren't in a tuple(), they're just straight concatenation.  I'll try going in four spaces instead of visual alignment and see what it does.

W8BSD: Still complaining, moving back to what looks right to me for now.

kk7ds:
Rebased and fixed the style thing. AFAICT, this is better than what we have right now, so I think it makes sense to merge this and iterate from here, especially since the diff is large enough to cause github's web UI heartache. I think the `_LOCK` thing can be removed now because of how chirp-next works, but I'll let you validate with your radios before we do it.

Anything else you know of outstanding that we need to resolve before we move to this?

Also, were you saying the TS-2000 stuff could be based on the newer of these? I've never had hands on one of those before but I know there are still some issues: https://chirp.danplanet.com/issues/10262

W8BSD:
> Rebased and fixed the style thing. AFAICT, this is better than what we have right now, so I think it makes sense to merge this and iterate from here, especially since the diff is large enough to cause github's web UI heartache.

Sounds good to me.  I've been amending and force-pushing a single branch through this review because I've kept telling myself I was almost done, so it will be good to get away from that.

> I think the `_LOCK` thing can be removed now because of how chirp-next works, but I'll let you validate with your radios before we do it.

It's still needed so you can load the settings tab while the memories are loading.

> Anything else you know of outstanding that we need to resolve before we move to this?

Nope, everything else I'm considering is new features, so there's no need to hold this up for that.

> Also, were you saying the TS-2000 stuff could be based on the newer of these? I've never had hands on one of those before but I know there are still some issues: https://chirp.danplanet.com/issues/10262

Yeah, I've actually got a quick hack for it working, but I want to test it a lot more since I was mostly concentrating on the settings.  I have a TS-B2000 as my daily driver, so it was a natural thing to play with.  I'll have to update it for the latest changes, but I can likely have it in shape for a PR in a week or two.

To do the TS-2000, I'm actually pulling out the settings stuff as a mixin, and overriding a bunch of methods.  I'll likely do a PR for the mixin fairly soon, but then add a superclass (kenwood_newlive) that will be shared by this new D7 and the TS-2000 which will help clean up both... aside from being request/response in ASCII, the two protocols are quite different, so basing them on the same class is a bit silly.

kk7ds:
> It's still needed so you can load the settings tab while the memories are loading.

Are you saying that because of the comment or because you tested and found it broken? If the latter, I should look into why the UI isn't using the thread for that because it likely means other live drivers are broken.

> Yeah, I've actually got a quick hack for it working, but I want to test it a lot more since I was mostly concentrating on the settings. I have a TS-B2000 as my daily driver, so it was a natural thing to play with. I'll have to update it for the latest changes, but I can likely have it in shape for a PR in a week or two.

Sweet, I know some folks will be happy. Please charge anything against that issue. I forgot to amend your commit message for this one, but "Fixes: #10262" will link the commit to the bug in the tracker.

Thanks!

W8BSD:
> Are you saying that because of the comment or because you tested and found it broken? If the latter, I should look into why the UI isn't using the thread for that because it likely means other live drivers are broken.

I added the comment after I tested it.

I assume two threads are being used, one for the memories and the other for the settings... when one thread drains the pipe and sends the command, the other thread needs to hold off until after the first one reads the reply.

kk7ds:
> I added the comment after I tested it.

Ah, okay I assumed it was there already.

> I assume two threads are being used, one for the memories and the other for the settings... when one thread drains the pipe and sends the command, the other thread needs to hold off until after the first one reads the reply.

Yeah, but there's a prioritized queue and thread consumer that the UI uses for memories and should be using to do things like `get_settings()` as well, that assumes one thing happens against the radio at a time. Anyway, I'll look into it, I was just curious if you had actually tried.

kk7ds:
Ah, we are *actually* starting multiple threads, but we shouldn't be. Most live radios do not have settings so this has gone unnoticed in the new UI, but that's broken. It also explains another problem I was having with a different radio a few weeks ago which I never got back to (it's a weird radio for other reasons). I'll work on that.

PR #688, opened 2023-06-28
Update es.po

MELERIX:
Updated few new strings to Spanish


kk7ds:
Now you're just showing off ;)

PR #689, opened 2023-06-29
Fix copy/paste between specials

kk7ds:
- ic2820: Fix invalid immutable fields
- Make fake driver support specials for testing
- Fix some issues copy/pasting between specials


kk7ds:
@W8BSD This fixes the copy and paste between specials on a few radios I tried, including on top of your change.

PR #693, opened 2023-06-30
Retevis RB27: support full band TX/RX - fixes #10686

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/tests/unit/test_chirp_common.py
+++ b/tests/unit/test_chirp_common.py
@@ -751,8 +751,7 @@ class TestOverrideRules(base.BaseTest):
         'Radioddity_DB25-G',
         'Retevis_RA85',
         'Retevis_RA685',
-        'Retevis_RB17P',
-        'Retevis_RB27',
+        'Retevis_RB17P'
     ]
 
     def _test_radio_override_immutable_policy(self, rclass):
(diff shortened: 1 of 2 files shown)


tests/unit/test_chirp_common.py
```diff
@@ -751,8 +751,7 @@ class TestOverrideRules(base.BaseTest):
         'Radioddity_DB25-G',
         'Retevis_RA85',
         'Retevis_RA685',
-        'Retevis_RB17P',
-        'Retevis_RB27',
+        'Retevis_RB17P'
```
kk7ds, line 754: Don't change this, but if you always make these end with a trailing comma, you can add and remove items from the list with only one line of change. It's an OCD thing :P

KC9HI: OK. The only reason I removed it was because you told not to do it on something else. Must have been in a different context. I agree, I find the trailing comma handy in this case.

kk7ds: Probably on a function call. It's technically allowed there too, but very unconventional. Something like:
```
some_thing(foo, bar,)
```
or on single-line lists:
```
foo = [1, 2, ]
```
is not generally done. But multi-line lists it's better to keep the trailing comma, just because each one is a line, and you end up having to make less change to add/remove things.

But as you can see, its importance is pretty microscopic :)

KC9HI: From what I recall, I believe that you are exactly right. Got it. Thanks!

PR #694, opened 2023-07-01
Move settings infrastructure into a separate mixin

W8BSD:
The settings "stuff" added in commit 3506133 is generally useful to other live drivers that want to expose settings, so split it out into a generic mixin rather than having it as part of the KenwoodD7Family class.

### DIFF
--- a/chirp/drivers/kenwood_d7.py
+++ b/chirp/drivers/kenwood_d7.py
@@ -23,6 +23,7 @@ from chirp.settings import RadioSetting, RadioSettingGroup, \
     RadioSettingValueInteger, RadioSettingValueBoolean, \
     RadioSettingValueString, RadioSettingValueList, RadioSettingValueMap, \
     RadioSettings, RadioSettingValueFloat
+from chirp.drivers.settings_mixin import SettingsMixin
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


W8BSD:
I'll fix up the style stuff, but I'm mostly worried about where I put the file... I haven't looked around to see how mixins are handled.

kk7ds:
> I'll fix up the style stuff, but I'm mostly worried about where I put the file... I haven't looked around to see how mixins are handled.

I'd have to look at what other live drivers have in the way of settings to see if this is actually useful for them in the way you have it, but I'm not sure we even have any other than kenwood ones. So I'd probably put it in `kenwood_live` (because some of the other drivers import/inherit from that base) or create a `kenwood_live_settings` or just keep it in `kenwood_d7` until and unless you're planning to refactor some of the other drivers to use it. But yeah, in the `drivers` module is where it should go. That's all the "radio specific stuff" and there is other infrastructure that has different rules for that module than the rest of chirp.

W8BSD:
> So I'd probably put it in `kenwood_live` (because some of the other drivers import/inherit from that base) or create a `kenwood_live_settings` or just keep it in `kenwood_d7` until and unless you're planning to refactor some of the other drivers to use it. But yeah, in the `drivers` module is where it should go. That's all the "radio specific stuff" and there is other infrastructure that has different rules for that module than the rest of chirp.

So yeah, my TS-2000 is using it as well, and I plan to detach that from d7, adding a new parent class for the two... but I can put it with the new parent class easily enough.

PR #695, opened 2023-07-01
Ensure exclusive access to radios per editorset

kk7ds:
A ChirpLiveEditorSet represents a conversation with a single radio
endpoint, but may involve multiple independent event streams. We
really don't need multiple threads and queues for talking to a single
radio, however radios with multiple sub-devices will end up with
different Radio classes by nature. So, instead of collapsing editors
into a single thread, we can just make them share a lock, established
at the EditorSet level, which is what this patch does.

This also means the other drivers (that I know of) that do their
own locking (kenwood_live, kenwood_d7, because settings) and ic9x
(because multiple sub-devices) can simplify and rely on this being
handled in the UI.

### DIFF
--- a/chirp/drivers/ic9x.py
+++ b/chirp/drivers/ic9x.py
@@ -52,7 +52,7 @@ CHARSET = chirp_common.CHARSET_ALPHANUMERIC + \
     "!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~"
 
 
-class Lock:
+class IC9xState:
     """Maintains the state of an ic9x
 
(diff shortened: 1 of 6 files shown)


kk7ds:
FYI @W8BSD this should make any threads for a single radio exclusive so we don't try to have more than one simultaneous conversation with the radio at a time. I've tested on a D700 for your new module, a TM-281 to confirm kenwood_live, and an IC-92 to get the other one. If you have a chance to look this over and/or test before we merge, that'd be cool. No rush since it's just cleanup.

PR #697, opened 2023-07-02
Add warnings and volatile flag to settings

kk7ds:
This adds two functions for settings. A warning message can be added,
which will either always be displayed to the user when they change the
value, or only when they change it away from some default. They are
given the opportunity to cancel the change if the warning makes them
reconsider.

Also, a setting can be marked "volatile" which will cause the settings
tab to be fully reloaded when it has been changed. This resets the
view as well, so it's not ideal, but this can be used when one setting
necessarily impacts another.


kk7ds:
@KC9HI FYI, this is what I was thinking for "careful with this setting" warning dialogs. I haven't tested this on windows yet, but here it is. I made the modified firmware toggle for the UV-K5 use it. We should probably similarly warn for other settings where the user can cause mayhem by changing it.

PR #698, opened 2023-07-02
Save backup images after every download

kk7ds:
This makes us save a backup image of every radio we download from
in the CHIRP profile directory, under backups/. This will help people
who don't capture a first-time download image from their radio, as
well as people who make some change that they end up not liking.

We'll need to prune these somehow going forward, but that can be done
later.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,6 +8,7 @@ The following must be true before PRs can be merged:
 * Commits in a single PR should be related.
 * Major new features or bug fixes should reference a [CHIRP issue](https://chirp.danplanet.com/projects/chirp/issues).
 * New drivers should be accompanied by a test image in `tests/images` (except for thin aliases where the driver is sufficiently tested already).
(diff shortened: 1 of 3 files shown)


kk7ds:
@KC9HI Thoughts on this? This might help with the "I hope you kept a pristine copy of your radio before you did something bad" problems right? Original idea from @goldstar611 a while back, but the recent rash of issues makes me think it's worth it.

I could make chirp automatically prune anything more than 365 days old, or maybe as much as necessary to keep ithe backup directory under 100MB or something.

kk7ds:
Okay, made it prune at 365 days, configurable by hidden setting in `chirp.config`.

goldstar611:
This is awesome. It won't help out if you download one then upload to many radios (one of my original hesitations on sending a patch earlier this year) but I think it will definitely help out in other cases like when manufacturers update the firmware years later without changing the radio ident for example.

kk7ds:
Yeah, won't help if you have never downloaded a radio and upload to it blindly, but there's not much we can do about that. But, if you have two and have *ever* downloaded both, at least there's a hope of finding one of the original ones for the radio you want :)

PR #700, opened 2023-07-03
Add subclass for Baofeng's 1900 series

cetinajero:
Hello,

Currently Chirp doesn't support the Baofeng BF-1900 series, this PR was created to support them, including:

- BF-1901(5W)
- BF-1902(5W)
- BF-1903(5W)
- BF-1904(10W)

The PR updates the settings2 struct to reflect the scan mode options used by this models.

Baofeng decided to remove the programmable side key function and used that reserved address to add the "Search" scan mode option (having 3 scan options requires more than 1 memory bit).

My testing and debugging was based on the original Baofeng software (BF_1901_FH) and comparing back and forth.

Fixes: #8551

✌🏼


kk7ds:
Thanks for this, but it's very little delta from the `h777` driver so it really needs to be a subclass for just the difference to avoid the duplication. We just need to refactor the base driver a tiny bit to support the difference. If you need help with this I could take a stab at it for you based on what you have here and then you can test and iterate from there.

cetinajero:
Let me take a shot at it, thank you Dan.

chirp/drivers/baofeng_1900.py
```diff
@@ -0,0 +1,596 @@
...
+
+def _bf1900_enter_programming_mode(radio):
+    serial = radio.pipe
+    # increase default timeout from .25 to .5 for all serial communications
+    serial.timeout = 0.5
+
+    try:
+        serial.write(b"\x02")
+        time.sleep(0.1)
+        serial.write(b"PWPG970")
```
kk7ds, line 99: To handle this, on the base class, define the program command like this:
```
class H777Radio(chirp_common.CloneModeRadio):
 ...
 PROGRAM_CMD = b'PROGRAM'
```
and on your subclass:
```
class Baofeng1900(H777Radio):
 ...
 PROGRAM_CMD = b'PWPG970'
```
Then make the original function like this:
```
def _h777_enter_programming_mode(radio):
    ...
    serial.write(radio.PROGRAM_CMD)
```
So that it uses the right command for the class being used.

chirp/drivers/baofeng_1900.py
```diff
@@ -0,0 +1,596 @@
...
+#seekto 0x03C0;
+struct {
+    u8 unused:6,
+       batterysaver:1,
+       beep:1;
+    u8 squelchlevel;
+    u8 scanmode;
+    u8 timeouttimer;
+    u8 unused2[4];
+} settings2;
```
kk7ds, line 70: Break this out into its own string:
```
H777_SETTINGS2 = """
 struct {
   ...
 } settings2;
```
and add one for yours:
```
BAOFENG1900_SETTINGS2 = """
struct {
 ...
} settings2;
```
Then see below in `process_mmap()`

chirp/drivers/baofeng_1900.py
```diff
@@ -0,0 +1,596 @@
...
+        rf.memory_bounds = (1, 16)
+        rf.valid_bands = [(400000000, 520000000)]
+        rf.valid_power_levels = BF1900_POWER_LEVELS
+        rf.valid_tuning_steps = [2.5, 5.0, 6.25, 10.0, 12.5, 15.0, 20.0, 25.0,
+                                 50.0, 100.0]
+
+        return rf
+
+    def process_mmap(self):
+        self._memobj = bitwise.parse(MEM_FORMAT, self._mmap)
```
kk7ds, line 300: Make this:
```
self._memobj = bitwise.parse(MEM_FORMAT + H777_SETTINGS2, self._mmap)
```
Then override `process_mmap()` in your subclass and use your settings2 variant instead.

chirp/drivers/baofeng_1900.py
```diff
@@ -0,0 +1,596 @@
...
+    _ranges = [
+        (0x0000, 0x0110),
+        (0x0380, 0x03E0),
+        (0x02B0, 0x02C0),
+    ]
+    _memsize = 0x03E0
+    _has_fm = True
+    _has_sidekey = False
+    _has_scanmodes = True
+    _has_scramble = False
```
kk7ds, line 267: I think you get the idea, but use flags like this to control the different behavior in things like `get_settings()` for your subclass so that the right things will get exposed (or not).

kk7ds (commented):
Great, thanks very much! Here are some hints to get you started. Also, please include a test image in `tests/images` for the radio to exercise the new subclass. You're welcome to put it into the `h777.py` file, or keep your `baofeng_1900.py` file and just `import h777` so you can subclass.

chirp/drivers/h777.py
```diff
@@ -665,3 +689,22 @@ class BFM4Radio(H777Radio):
     def match_model(cls, filedata, filename):
         # This model is only ever matched via metadata
         return False
+
+@directory.register
+class BF1900Radio(H777Radio):
+    VENDOR = "Baofeng"
+    MODEL = "BF-1901"
+    PROGRAM_CMD = b'PWPG970'
+    ALIASES = [BaofengBF1904]
```
kk7ds, line 698: We're actually trying to avoid adding new Aliases nowadays. Can you convert that into just a subclass of this and `@directory.register` it like this one? I assume the 1904 is the one with the different power levels? If so, you might also want to make that subclass expose a different set of values so it's correct?

cetinajero:
Thanks for your guidance Dan, WIP yet.

Still pending tasks:

- Alarms as lists
- `Search` scan mode
- Valid band range up to 520 MHz
- Vox level up to 9

kk7ds:
Looks pretty close, that was fast :)

chirp/drivers/h777.py
```diff
@@ -665,3 +670,34 @@ class BFM4Radio(H777Radio):
...
+    def match_model(cls, filedata, filename):
+        # This model is only ever matched via metadata
+        return False
+
+@directory.register
+class BF1904Radio(BF1901Radio):
+    VENDOR = "Baofeng"
+    MODEL = "BF-1904"
+    ALIASES = []
+
```
kk7ds, line 699: Do you want separate power levels for this? I think this is the 10W unit? Not a huge deal, but it should be easy.

cetinajero: Different power levels but same subset of features... what do you suggest?

kk7ds:
Just need to fix the style things to be mergeable. Also, can you fix up the commit message? I think it still has content from a previous commit in it. Also put "Fixes: #8551" in the commit message so the bug system will close that bug and point to this when it merges. Looks great otherwise.

I can rebase for you when ready, if it helps.

chirp/drivers/h777.py
kk7ds: It would be better to have different power levels exposed by that driver so that when importing memories it selects the closest one. You can just do the same thing you did for the VALID_BANDS but with a list of power levels.

cetinajero: Got it

cetinajero:
Yes, I'm running with `tox -e style`

Is there an auto fix style option?

I plan to add 3 total commits (other two for the scan list struct and alarm option list), is that ok?

kk7ds:
> Yes, I'm running with `tox -e style`
> 
> Is there an auto fix style option?

No, but I can fix for you if needed.

> I plan to add 3 total commits (other two for the scan list struct and alarm option list), is that ok?

Sure. However, I'm going to be unavailable for the next week starting tomorrow. How about we just get this merge-able now and you can follow up with the extra features later?

cetinajero:
Ok, almost there

cetinajero:
It should be mergeable by now Dan, let me know your thoughts.

kk7ds:
Looks good to me. Can I tweak your commit message for the first commit to reference the bug? Other than that I'm good.

cetinajero:
Oh yeah, I forgot that.

I can just amend it.

kk7ds:
Also, I assume we want to tell users with a BF-1902 to use BF-1901 right? There's mapping in the code that will update the website after merge so it says "BF-1902 (use BF-1901)".

cetinajero:
> There's mapping in the code that will update the website after merge so it says "BF-1902 (use BF-1901)".

Yeah, that will be helpful, where can we add those relationships?

kk7ds:
I added it here, to `model_alias_map.yaml`. The build process will update the web page tomorrow when it creates the build.

cetinajero:
Awesome, thanks Dan!

It was great working on this

kk7ds:
Excellent thanks very much!

PR #702, opened 2023-07-04
SP calling frequencies stock config

szporwolik:
Created stock config with calling frequencies and frequently used frequencies for Poland (SP).


kk7ds:
Thanks. Are you able to run chirp direct from git? If so, I can get you to test the przemienniki code before I merge.

szporwolik:
Yes, with no problem - I've actually set up a full environment.

kk7ds:
Great thanks, I'll mention you on the PR so you can look.

In the meantime, can you squash these two commits into one so we don't have the rename in the history? I can do it for you if not.

szporwolik:
Closing, will open new one with squashed commits.

PR #703, opened 2023-07-04
Re-add przemienniki.net query source

kk7ds:
This is very rough and lightly tested so far.

Fixes #10700

### DIFF
--- a/chirp/locale/de.po
+++ b/chirp/locale/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-07-02 14:20-0700\n"
+"POT-Creation-Date: 2023-07-04 11:04-0700\n"
 "PO-Revision-Date: 2012-10-02 22:11+0100\n"
 "Last-Translator: Benjamin, HB9EUK <hb9euk@hb9d.org>\n"
 "Language-Team: German\n"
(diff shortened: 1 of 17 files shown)


kk7ds:
@szporwolik this is a suuuuper fast implementation this morning while eating breakfast. Please test thoroughly ;D

kk7ds:
Also, if you have better text to put into the "info" blob for this, that would be good. The info on each of these sources is suppose to be a short description, so maybe something like "A european repeater database" or something?

kk7ds:
Oops, forgot to commit the new file. I blame breakfast ;)

szporwolik:
@kk7ds thank your for this quick implementation! 

Things that I would suggest to change, prior merging (the minimum functionality):
- after pulling the repeaters there is "dummy" line at the top : 146.0100 - this could be skipped, seems like a bug
- switching to digital repeaters (mode = DV) gives a parsing error; I would skip the DV (digital) repeaters for the beginning if this will make the whole feature go online sooner 
2023-07-04 18_33_15-CHIRP (przemienniki net)
- the distance radius does not work, even when giving all the information still I got all the repeaters in the country (screen) (same comment as with DV, even without this feature getting the list of repeaters from country would be beneficial)
2023-07-04 18_31_54-CHIRP (przemienniki net)
- the distance shall have the unit, like "Distance [km]" or something
- I would add the following information (just a suggestion ;-) )  - "FREE repeater database, which provide most up-to-date information about repeaters in Europe . No account is required."
- Slovenia "si", Netherlands "nl", Iceland "is", Russia "ru" are missing in the country drop down list comparing to website

kk7ds:
>     * after pulling the repeaters there is "dummy" line at the top : 146.0100 - this could be skipped, seems like a bug

Fixed.

>     * switching to digital repeaters (mode = DV) gives a parsing error; I would skip the DV (digital) repeaters for the beginning if this will make the whole feature go online sooner

Hmm, yeah, DV returns nothing for me in any country (that I tried). I just added this because it was in the other implementation. Perhaps these are all gone, or no longer supported in the database?

>     * the distance radius does not work, even when giving all the information still I got all the repeaters in the country (screen) (same comment as with DV, even without this feature getting the list of repeaters from country would be beneficial)

Yep, I was passing 'Latitude' instead of 'latitude' (et al) and the API seems to be case-sensitive. I think it's fixed now.

>     * the distance shall have the unit, like "Distance [km]" or something

This is in the tooltip. I can change this at some point, but it will invalidate all the translation strings for the other dialogs that use the same,.

>     * I would add the following information (just a suggestion ;-) )  - "FREE repeater database, which provide most up-to-date information about repeaters in Europe . No account is required."

Sounds fine to me.

>     * Slovenia "si", Netherlands "nl", Iceland "is", Russia "ru" are missing in the country drop down list comparing to website

Done.

Will push an update in a sec. I'm going to be unavailable for the rest of the week, so if you can again test real quick we should be able to get this merged before I go. Thanks!

szporwolik:
Quickly tested. Functionally I think this is a perfect start! 

If possible it would be good the have the query repeaters sorted by name (now it seems a bit random, when coordinates are not given). This is minor feature request.

Great thank you!

kk7ds:
Okay, cool, thanks for your help. I will merge this soon.

The sort order is whatever is returned from the database server. You think sort by name (not frequency) is best?

kk7ds:
This update sorts by name if either lat and/or lon is not provided. I think the server sorts by distance when those are provided, so this gives you either server sort order by distance, or name if not. Easy to change to frequency if you think that's better, otherwise this should be good.

PR #704, opened 2023-07-04
Created stock config file for Poland (SP)

szporwolik:
Created stock config file for Poland (SP) with frequently used frequencies and simplex calling frequencies.


kk7ds:
Thanks, I just rebased this on master and will merge.

PR #707, opened 2023-07-05
Update es.po

MELERIX:
Translated new strings to Spanish


kk7ds:
Sorry, I *do* try to avoid too much damage to translated strings when I do this stuff, but there has been a lot of activity lately. Thanks as always :)

PR #708, opened 2023-07-05
Polish (PL) translation update

szporwolik:
Major update, all strings not related to drivers are now translated.


szporwolik:
Please excuse, will create new one without merge,

PR #710, opened 2023-07-05
Przemienniki.net query source updates

szporwolik:
After initial implementation by @kk7ds  (thanks again!) I've fixed some minor items:
- added parameter to only show working repeaters provided by przemienniki.net API
- added support for digital repeaters and various other modes provided by przemienniki.net API
- added possibility do select multiple bands at the same time
- alphabetically reordered country list in the dropdown

Related to https://chirp.danplanet.com/issues/1369

Notes:
- band selection is not written into config - the reason is that if we will not show at the main window the previously chosen  bands I can expect that some users will be confused about this, instead we will force user to open the modal band selection window each time; BUT 2m and 70cm are always pre-selected, for user convince and to cover most of the use cases 
- the default country is pre-selected to 'pl'; I don't have exact stats data for przemienniki.net, but my guess would be that 8 out of 10 users are from Poland, so this seemed for me as the better option that choosing first country on the list with some minimal user base

This query source is not perfectly tested, but I think I have covered most common use test cases.

### DIFF
--- a/chirp/locale/de.po
+++ b/chirp/locale/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-07-04 11:04-0700\n"
+"POT-Creation-Date: 2023-07-09 17:54-0700\n"
 "PO-Revision-Date: 2012-10-02 22:11+0100\n"
 "Last-Translator: Benjamin, HB9EUK <hb9euk@hb9d.org>\n"
 "Language-Team: German\n"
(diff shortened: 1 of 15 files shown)


szporwolik:
@kk7ds seems that translation  files need update, could you please help with that ?

kk7ds:
I'm off grid this weekend but I will definitely look and advise when I get back. Thanks very much for working on this!


________________________________

kk7ds:
For reference, you need `gettext` installed and then just:
```
make -C chirp/locale clean all
```
to refresh the translation files any time you add a new marked-for-translation string.

Excellent job, thanks!

szporwolik:
Thanks for help! 👍🏻

PR #711, opened 2023-07-07
Add Maverick RA-100 & RA-425 radios

cetinajero:
Hello Dan!

I work for Grupo PV and we are distributing 2 radio models now compatible with Chirp (added in #603 & #700).

Can you help us by including our own OEM brand as a supported one by Chirp?

Please find attached some example resources where our products are published:

- https://www.grupopv.mx/maverick/analogos/portatiles/ra-100u/
- https://www.solucionestse.com/search/?q=maverick
- https://grupopv-public.s3.amazonaws.com/assets/pdf/documents/brochures/sytba/radios/maverick/rd-500.pdf
- https://www.studocu.com/es-ar/document/universidad-autonoma-de-chihuahua/sistemas-operativos-i/radio-portatil-analogo-maverick-ra-425-16-ch-4-watts-uhf-400-520-mhz/64692573
- https://www.google.com/search?q=Maverick+RA-100+RA-425

Thanks in advance!

✌🏼


kk7ds:
For less-popular (or just widely-applicable) direct aliases like this, we usually put them in `model_alias_map.yaml` so the website reflects the support but we don't actually list it in the download box. For popular models it's worth having it in the drop-down of course. I hope this won't turn into a long pattern of adding lots of empty aliases, but we definitely already have some, so we can go ahead with these.

PR #714, opened 2023-07-10
General translation fixes

szporwolik:
While checking Polish(PL) translation I've identified a few strings that are not properly translated.
Proposed fixes shall improve quality of all translations.

### DIFF
--- a/chirp/wxui/main.py
+++ b/chirp/wxui/main.py
@@ -757,6 +757,7 @@ class ChirpMain(wx.Frame):
 
         self._last_search_text = ''
         find_item = edit_menu.Append(wx.ID_FIND)
+        edit_menu.SetLabel(wx.ID_FIND, _('Find'))
         self.Bind(wx.EVT_MENU, self._menu_find, find_item)
 
         self._find_next_item = wx.NewId()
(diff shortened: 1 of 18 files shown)


chirp/wxui/main.py
```diff
@@ -942,7 +943,7 @@ def make_menubar(self):
         menu_bar.Append(edit_menu, wx.GetStockLabel(wx.ID_EDIT))
         menu_bar.Append(view_menu, '&' + _('View'))
         menu_bar.Append(radio_menu, '&' + _('Radio'))
-        menu_bar.Append(help_menu, wx.GetStockLabel(wx.ID_HELP))
+        menu_bar.Append(help_menu, _('Help'))
```
kk7ds, line 946: Hmm, we're really not supposed to need to do this for these special items, as the `GetStockLabel` helper accounts for platform differences, like on macos it's common for the "find" issue to be "Find..." but not on others. I think this means it's not translated for your platform, but should be. Which platform are you on? Any chance the formal builds work properly and running from a dev environment does not?

kk7ds:
I rebased this before I looked at the changes, which is why it's now failing that check. After we resolve the stock item thing you can just refresh it or I can. Thanks for doing this, definitely found a few missed ones there!

chirp/wxui/main.py
szporwolik: Yes, that is strange. I found this on formal build at two Win10 machines :
image

I'm not sure if this is PL related or Windows related, but definitely something is not working with that configuration. What is interesting rest of the labels (Copy & Paste etc.) were properly pulled via GetStockLabel. 

This is how it looks for me: 
image

kk7ds: Hmm, yeah, weird. On Linux (where I can easily flip the locale for one program) I get Help translated by stock item, but not Find, when using the `es` locale. On windows, `es` has both Find and Help translated. I can replicate your experience on windows with `pl`. On macos and `pl`, Help is translated but not find. Ugh!

Okay, I guess this is the best plan then. I'll rebase this and fix the conflicts.

szporwolik: Perfect, thanks! 👍🏻

PR #719, opened 2023-07-11
tdh8 add only CH mode

Sandmann34:
TDH8 add a new function: only CH mode

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -106,7 +106,9 @@ struct {
   u8 save;
   u8 ligcon;
   u8 voxdelay;
-  u8 alarm;
+  u8 onlychmode:1,
+     unused:6,
+     alarm:1;
 } settings;
 
 #seekto 0x0CB8;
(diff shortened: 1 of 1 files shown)


kk7ds:
This looks fine, except for two things:

First, you should not need to change the images in the tree for this, and it's better if you don't to make sure you're not depending on something in them to make the tests work. Can you reset them so only the code is changed?

Second, your commit is authored by "unknown". Can you please set your identity and amend the commit?

```
git config --global user.name "John Doe"
git config --global user.email johndoe@example.com
git commit --amend --reset-author
```
Thanks. If you need help with the first one, I can fix it for you, but do the second thing for sure.

kk7ds:
After you amend, you will need to push with `-f` to force the update into this PR.

kk7ds:
Ah, this is why I wanted you to not change the images. It looks like your get_settings() and set_settings() are not identical so that the language isn't getting changed perhaps?

kk7ds:
I think this is what you want. Can you confirm this works on the actual radio for both English and Chinese? If so, we can merge this.

Sandmann34:
> 我想这就是你想要的。您能确认这在英语和中文的实际广播中都有效吗？如果是这样，我们可以合并它。

yes i tested it works.

kk7ds:
Cool, thanks.

PR #721, opened 2023-07-12
FIX: Retevis-RB627B [Enabled Power Mode and Set Max Power to 2W (2.00)]

dguimaraes88:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/bf_t8.py
+++ b/chirp/drivers/bf_t8.py
@@ -902,12 +902,14 @@ class RetevisRB27V(RetevisRB27B):
 class RetevisRB627B(RetevisRB27B):
     VENDOR = "Retevis"
     MODEL = "RB627B"
-    POWER_LEVELS = [chirp_common.PowerLevel("High", watts=0.50),
+    DUPLEXES = ['', '-', '+', 'off']
+    POWER_LEVELS = [chirp_common.PowerLevel("High", watts=2.00),
                     chirp_common.PowerLevel("Low", watts=0.50)]
+    ODD_SPLIT = False
 
     _upper = 16
-    _pmr = True
-    _frs = _gmrs = _murs = False
+    _pmr = True    
+    _frs = _murs = _pmr = False
 
 
 @directory.register


chirp/drivers/bf_t8.py
```diff
@@ -902,12 +902,14 @@ class RetevisRB27V(RetevisRB27B):
...
+    DUPLEXES = ['', '-', '+', 'off']
+    POWER_LEVELS = [chirp_common.PowerLevel("High", watts=2.00),
                     chirp_common.PowerLevel("Low", watts=0.50)]
+    ODD_SPLIT = False
 
     _upper = 16
-    _pmr = True
-    _frs = _gmrs = _murs = False
+    _pmr = True    
+    _frs = _murs = _pmr = False
```
kk7ds, line 912: @KC9HI Does this look right to you?

KC9HI: No. It look like they are trying to remove the PMR restrictions from the RB627B like I have been doing for other Retevis models (I got the green light from them to drop PMR, FRS, GMRS, etc restrictions allow CHIRP to fully support the radios actual capability so I have been doing it when requested).

What I simply would have done here to drop the restrictions with the least amount of change would have been to simply change line 909 _pmr = True to _pmr = False and left line 910 alone. Removing _gmrs from line 910 could inadvertently turn on GMRS restrictions on for this PMR radio.

KC9HI: I have also been putting a comment on line 909 explaining:
_pmr = False  # sold as PMR radio but supports full band TX/RX

KC9HI: And I just remembered that the reason I didn't change this one is because the user can just use the RB27 selection and get the whole shooting match (99 channels, bull band, wide/narrow bandwidty, high/low power, etc). There was no need to change RB27V or RB627B.

dguimaraes88: Ok...I understand ;) no problem, Good joob

KC9HI:
The Model Alias Map change that I just submitted was related to this PR so that owners of this radio (RB627B) and its close cousins (RB27B and RB27V) know to use the RB27 to get full support. That makes this PR no long necessary.

PR #724, opened 2023-07-13
uv5r: Warn about other settings not being honored

kk7ds:


kk7ds:
@KC9HI Let me know what you think. We can also just drop this and proceed with your proposed fix if you think it'd be better.

KC9HI:
OK. I have done some testing. So this triggers a prompt when each affected setting is edited. Not what I was expecting but I see how this could be a useful thing to add to the Range Override Parameter setting on the Advanced Settings tab.

This needs to also be implemented for the settings listed under the Service Settings tab since they to are blocked from being uploaded.

kk7ds:
Ah I see I missed that change for the service ones. I'll add those.

And yes, it warns you for every single one - hopefully it'll be quite clear. It also only does it if your image has that version, so not everyone will get warned.

KC9HI:
Correct. So if you make that change and merge it, I can make my change and fewer radios will get the warning. Hopefully the 2 things together will eliminated (or at least reduce) the daily tickets regarding the missing tabs. :)

KC9HI:
OK. Looks good to me. I think I understand what you did good enough to be able to integrate what I've done with what you've done.

kk7ds:
Cool, sounds great, thanks!

PR #726, opened 2023-07-14
UV-5R: fine tune Aux block upload restrictions - related to #10505

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/uv5r.py
+++ b/chirp/drivers/uv5r.py
@@ -193,8 +193,10 @@ struct {
      txpower3:2;
 } vfob;
 
-#seekto 0x0F56;
-u16 fm_presets;
+#seekto 0x0F48;     // some new radios do not support this feature
+char bcstfmlo[14];  // broadcast FM 65-75 MHz band
+u16 fm_presets;     // broadcast FM frequency
+char bcstfmhi[14];  // broadcast FM 75-108 MHz band
 
 #seekto 0x1008;
 struct {
@@ -473,7 +475,10 @@ def _get_radio_firmware_version(radio):
     else:
         block1 = _read_block(radio, 0x1E80, 0x40, True)
         block2 = _read_block(radio, 0x1EC0, 0x40, False)
+        block3 = _read_block(radio, 0x0F40, 0x40, False)
         version = block2[48:62]
+        brdcstfm = block3[0:30]
+        return version, brdcstfm
     return version
 
 
@@ -517,7 +522,10 @@ def _ident_radio(radio):
 def _do_download(radio):
     data = _ident_radio(radio)
 
-    radio_version = _get_radio_firmware_version(radio)
+    if radio.MODEL == "BJ-UV55":
+        radio_version = _get_radio_firmware_version(radio)
+    else:
+        radio_version, radio_fm = _get_radio_firmware_version(radio)
     LOG.info("Radio Version is %s" % repr(radio_version))
 
     if b"HN5RV" in radio_version:
@@ -591,9 +599,15 @@ def _do_upload(radio):
             raise errors.RadioError("Image not supported by radio")
 
     image_version = _firmware_version_from_image(radio)
-    radio_version = _get_radio_firmware_version(radio)
+    if radio.MODEL == "BJ-UV55":
+        radio_version = _get_radio_firmware_version(radio)
+    else:
+        radio_version, radio_fm = _get_radio_firmware_version(radio)
     LOG.info("Image Version is %s" % repr(image_version))
     LOG.info("Radio Version is %s" % repr(radio_version))
+
+    imagev_matched_radiov = image_version == radio_version
+    _skip_aux_block = False
 
     # default ranges
     _ranges_main_default = [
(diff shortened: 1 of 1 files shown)


KC9HI:
If this is acceptable, you can merge if you want. Or I can add it to #10505 tomorrow for testing.

chirp/drivers/uv5r.py
```diff
@@ -663,20 +687,21 @@ def _do_upload(radio):
...
             "The upload has finished successfully.\n"
             "Please restart CHIRP.")
+        return
+
+    if radio._aux_block and not imagev_matched_radiov:
+        msg = ("This is NOT an error. The upload finished successfully.\n"
+               "The 'Other Settings' and 'Service Settings' were skipped "
+               "because the firmware version of the image (%s) does not "
+               "match that of the radio (%s).")
+        raise errors.RadioError(msg % (image_version, radio_version))
```
kk7ds, line 704: We need to figure out a better way to do this. I need to do some thinking on it, so don't let me forget.

chirp/drivers/uv5r.py
```diff
@@ -1104,14 +1129,17 @@ def _my_upper_band(self):
...
         _ani = self._memobj.ani
         _fm_presets = self._memobj.fm_presets
         _settings = self._memobj.settings
         _squelch = self._memobj.squelch_new
         _vfoa = self._memobj.vfoa
         _vfob = self._memobj.vfob
         _wmchannel = self._memobj.wmchannel
-        HN5RV = "HN5RV" in str(self._memobj.firmware_msg.line1)
+        _has_fmradio = (str(_mem.bcstfmlo) == "   FM   65-75M" and
+                        str(_mem.bcstfmhi) == "   FM  76-108M")
```
kk7ds, line 1141: Hmm, these strings are in the memory image and not changeable by the user?

KC9HI: They not changeable by the typical CHIRP user because they are not exposed to the user. But they are not immutable like the firmware version is in Aux memory block, if that's what you mean. They could be changed by someone that really wanted to change them.

It is for this reason that I block the upload of this region of memory if either or both the image or radio does not have these strings so they can't get overwritten by normal CHIRP use.

chirp/drivers/uv5r.py
```diff
@@ -591,55 +599,71 @@ def _do_upload(radio):
...
+        # 0x0F48-0x0F65 contains invalid data
+        # This radio is likely to 'break' the Aux block if it is changed.
+        # The safest thing to do is to skip uploading of the Aux block.
+        ranges_main = _ranges_main_no_fm
         _skip_aux_block = True
-    else:
-        _skip_aux_block = False
+    # check source image for invalid bcst FM data
+    elif "\xFF" in str(radio._memobj.bcstfmlo) or \
+            "\xFF" in str(radio._memobj.bcstfmhi):
```
kk7ds, line 650: Okay and this is the check for upload.

chirp/drivers/uv5r.py
```diff
@@ -193,8 +193,10 @@
      txpower3:2;
 } vfob;
 
-#seekto 0x0F56;
-u16 fm_presets;
+#seekto 0x0F48;     // some new radios do not support this feature
+char bcstfmlo[14];  // broadcast FM 65-75 MHz band
+u16 fm_presets;     // broadcast FM frequency
+char bcstfmhi[14];  // broadcast FM 75-108 MHz band
```
kk7ds, line 199: Oh I see you added these elements here just now, I got it.

I guess maybe these could be different in some radios for other geographies maybe? I guess

kk7ds:
I won't be around this weekend to merge stuff, so let's just go with this and we can revert later if something isn't right. Thanks!

chirp/drivers/uv5r.py
KC9HI: Yeah. The problem here is that Baofeng pretty much stopped separating the various models by using different magics. So it is entirely possible to upload an image from a dual-band radio into a tri-band radio hosing the band limits. It is a similar situation between radios with 2 TX power levels and 3 TX power levels.

kk7ds: What I mean is a better way to raise a message to the user after cloning that isn't an error. So that you don't have to say "ERROR: This is not an error" :)

chirp/drivers/uv5r.py
KC9HI: Yes. I only have 1 radio here with the issue. But the 2 or 3 images that I have gotten from others all have bcstfmlo[14] and bcstfmhi[14] fully populated with 0xFF. So far I have never seen an 0xFF in an older "good" radio so I just check for a single 0xFF when uploading. I would rather err on the side of not uploading to a "good" radio than accidentally upload to a "bad" one.

chirp/drivers/uv5r.py
KC9HI: Yeah. I've never liked this. People see the error symbol and never read what it says and always thing something bad happened. You are the one that suggested that I add the "This is not an error" part.

kk7ds: Yeah, I remember. I'll do some thinking.

PR #729, opened 2023-07-16
Update ft2800.py - Add support for extended TX version of radio

aaknitt:
Add new class for radios with extended transmit modification (MARS/CAP) that uses a different ID_BLOCK value.

https://chirp.danplanet.com/issues/3919

Succesfully bench tested upload and download with an FT-2800M with TX mod.

### DIFF
--- a/chirp/drivers/ft2800.py
+++ b/chirp/drivers/ft2800.py
@@ -33,7 +33,12 @@ def _send(s, data):
             raise Exception("Failed to read echo chunk")
 
 
-IDBLOCK = b"\x0c\x01\x41\x33\x35\x02\x00\xb8"
+# The IDBLOCK is the first thing sent during an upload or download
+# and indicates the radio subtype:
+#   USA Unmodified:            b"\x0c\x01\x41\x33\x35\x02\x00\xb8"
+#   USA With extended TX mod:  b"\x0c\x01\x41\x33\x35\x03\x00\xb9"
+SUPPORTED_IDBLOCKS = [b"\x0c\x01\x41\x33\x35\x02\x00\xb8",
+                      b"\x0c\x01\x41\x33\x35\x03\x00\xb9"]
 TRAILER = b"\x0c\x02\x41\x33\x35\x00\x00\xb7"
 ACK = b"\x0C\x06\x00"
 
@@ -43,7 +48,8 @@ def _download(radio):
     attempts = 30
     for _i in range(0, attempts):
         data = radio.pipe.read(8)
-        if data == IDBLOCK:
+        if data in SUPPORTED_IDBLOCKS:
+            radio.subtype = data
             break
         LOG.debug('Download attempt %i received %i: %s',
                   _i, len(data), util.hexprint(data))
@@ -103,7 +109,7 @@ def _upload(radio):
             break
         LOG.debug("What is this garbage?\n%s" % util.hexprint(data))
 
-    _send(radio.pipe, IDBLOCK)
+    _send(radio.pipe, radio.subtype)
     time.sleep(1)
     ack = radio.pipe.read(300)
     LOG.debug("Ack was (%i):\n%s" % (len(ack), util.hexprint(ack)))
@@ -181,6 +187,14 @@ class FT2800Radio(YaesuCloneModeRadio):
 
     _block_sizes = [8, 7680]
     _memsize = 7680
+
+    @property
+    def subtype(self):
+        return bytes(self.metadata['subtype_idblock'])
+
+    @subtype.setter
+    def subtype(self, value):
+        self.metadata = {'subtype_idblock': [x for x in value]}
 
     @classmethod
     def get_prompts(cls):


kk7ds:
Thanks for this, and thanks for mirroring the other 1900 example directly. However, I hate that we ever did it that way and have regretted it ever since :)

I wonder if you would be willing to try to do this differently, or at least entertain an attempt? I'm thinking that we could, on download, wait for any IDBLOCK-looking thing and when we get one make sure it's either the normal or modified one, and then record that. Then we can either tack that onto the end of the image, or store it in the image metadata chunk. The driver can then look at that to know which limits to expose, and when you go to upload, use that to identify itself to the radio. That would allow us to have one driver entry for this radio (which is more appropriate) but still keep the radio happy. Should work right? The Icom ID-5100 driver supports three firmware revs of the radio this way, and some Kenwood and Yaesu drivers support persisting file format details in the metadata blob. What do you think?

Also, please put a bug reference in the commit (i.e. "Fixes: #3919") so the ticketing system will correlate the fix to the bug.

aaknitt:
> I wonder if you would be willing to try to do this differently, or at least entertain an attempt? I'm thinking that we could, on download, wait for any IDBLOCK-looking thing and when we get one make sure it's either the normal or modified one, and then record that. Then we can either tack that onto the end of the image, or store it in the image metadata chunk. The driver can then look at that to know which limits to expose, and when you go to upload, use that to identify itself to the radio. That would allow us to have one driver entry for this radio (which is more appropriate) but still keep the radio happy. Should work right? The Icom ID-5100 driver supports three firmware revs of the radio this way, and some Kenwood and Yaesu drivers support persisting file format details in the metadata blob. What do you think?

Sure, I'll need to educate myself on the image format and metadata chunk and look at some of the other examples you referenced.  I'm still early on the learning curve of the overall software architecture and functionality.  Would storing this data in the image then prevent using an image from one variant to write to a different variant?  For example, if I have an unmodified radio that I read and save to an image...if it's pulling the IDBLOCK from that image, then it won't be able to write that same data to a modified radio, I think?  Is that a problem we need to solve for, or just use Export/Import to manage that?  I guess that same issue already exists with FT-1900 method, so maybe it's ok, but with today's approach you at least know you've selected a variant vs having it kind of occur under the hood in the image with no user visibility to it.  Wondering if it could lead to confusion/complaints.  Let me know what you think.  


> Also, please put a bug reference in the commit (i.e. "Fixes: #3919") so the ticketing system will correlate the fix to the bug.

Will do

kk7ds:
> Sure, I'll need to educate myself on the image format and metadata chunk and look at some of the other examples you referenced. I'm still early on the learning curve of the overall software architecture and functionality.

Here's an example of a driver that uses it in a simple way:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/ft4.py#L689

You don't have to implement the getter/setter in the same way there, but you can. That driver basically has to do the same thing, but for geo reasons instead of modification reasons. Hopefully you can follow the tentacles of that in the file there and figure it out, but feel free to ask if you need help.

> Would storing this data in the image then prevent using an image from one variant to write to a different variant? For example, if I have an unmodified radio that I read and save to an image...if it's pulling the IDBLOCK from that image, then it won't be able to write that same data to a modified radio, I think? Is that a problem we need to solve for, or just use Export/Import to manage that? I guess that same issue already exists with FT-1900 method, so maybe it's ok, but with today's approach you at least know you've selected a variant vs having it kind of occur under the hood in the image with no user visibility to it. Wondering if it could lead to confusion/complaints. Let me know what you think.

Yeah, it would prevent using a modified image on an unmodified radio and vice versa, but I think that's what Yaesu's intent was by making the radio identify as a different model once modified. That prevents you from doing a straight cable clone between the two, and is also why (IIRC) the radio must be reset post modification. We have cheap chinese radios supported in chirp where we *can't* tell the difference between one and another radio, and so we recommend people *always* keep the image tied to the radio and copy/paste memories between them. So I think there's prior art there, and I certainly have no problem fielding questions from users that complain about that. However, as you note, that's *already* a restriction of the multiple-class model that the 1900 started.

Thanks!

aaknitt:
@kk7ds Take a look at this now and see if it's matching what you were thinking.  It seems to be working well with a modified radio.  I also manually edited  the saved .img file to change the stored IDBLOCK value and after doing so it would not upload, which is what I would expect.

chirp/drivers/ft2800.py
```diff
@@ -182,6 +189,17 @@ class FT2800Radio(YaesuCloneModeRadio):
...
+    @property
+    def subtype(self):
+        if 'subtype_idblock' in self.metadata:
+            return bytes(self.metadata['subtype_idblock'], 'latin1')
+        else:
+            return SUPPORTED_IDBLOCKS[0]
+
+    @subtype.setter
+    def subtype(self, value):
+        self.metadata = {'subtype_idblock': value.decode('latin1')}
```
kk7ds, line 201: This encode and decode from latin1 is not quite right. I'm guessing you're doing this because the load and save doesn't restore the bytes because we're using json for the payload?

In that case, how about we store this as an integer array of byte values that will store properly in JSON? That will make the getter/setter you have here even more useful. So something like this for the setter:
```
self.metadata = {'subtype_idblock': [x for x in value]}
```
and this for the getter:
```
return bytes(self.metadata['subtype_idblock'])
```

kk7ds (commented):
Yep, this looks pretty close, just a couple things. Thanks!

kk7ds:
Oh one other thing. This is a nit, but could you make the first line of the commit message a short description (like the title of this PR) and then a blank line, and any additional stuff like the bug reference? The first line goes out in the build email summary each night and that kinda needs to be a short single line that makes sense to match everything else.

chirp/drivers/ft2800.py
```diff
@@ -43,7 +42,8 @@ def _download(radio):
     attempts = 30
     for _i in range(0, attempts):
         data = radio.pipe.read(8)
-        if data == IDBLOCK:
+        print(data)
```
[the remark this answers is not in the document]
aaknitt: yeah that print statement should have been deleted altogether prior to the PR, I just missed it.  There are already LOG.debug statements that capture that same data.  Fixed in the latest push.

chirp/drivers/ft2800.py
aaknitt: Agree, the latin1 thing was stupid.

kk7ds:
Cool, looks great now, thanks!

PR #746, opened 2023-07-30
Fix spelling

dforsi:
Fixes capitalization of units of measure; adds space before units of measure; other fixes.

I did check manually that the changes s/Compandor/Compander/ and s/Venezuala/Venezuela/ work as intended.


chirp/sources/repeaterbook.py
```diff
@@ -387,7 +387,7 @@ def included_band(item):
     "United Arab Emirates",
     "United Kingdom",
     "Uruguay",
-    "Venezuala",
+    "Venezuela",
```
kk7ds, line 390: Oof, this one was copied directly from the repeaterbook.org site:

https://www.repeaterbook.com/row_repeaters/index.php?state_id=none

However, the misspelled version doesn't work in the API but the correct one does. So this fixes a bug *and* the display string :)

PR #747, opened 2023-07-31
ga510.py: add 'reset' v2

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/ga510.py
+++ b/chirp/drivers/ga510.py
@@ -45,7 +45,10 @@ DTMFCHARS = '0123456789ABCD*#'
 
 
 def reset(radio):
-    radio.pipe.write(b'E')
+    try:
+        radio.pipe.write(b'E')
+    except Exception as e:
+        LOG.error('Failed to reset radio: %s' % e)
 
 
 def start_program(radio):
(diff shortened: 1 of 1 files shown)


KC9HI:
I gave up on 'finally:'. I couldn't get it to pass the 'driver' test.

kk7ds:
I'm guessing one of the clone tests? I imagine you need to swallow any exception you get in the `reset()` function which will re-raise something other than `RadioError` on the error test. If you don't mind, let me take a shot tomorrow and I'll see if I can get it to work.

kk7ds:
This is what I had in mind. Works for me with a ga510 and passes tests. @KC9HI let me know if you're okay with this. Your way definitely works, but it's easier to miss a case (or add a new one that doesn't reset).

KC9HI:
I did look at the tk3140.py driver last night. It looked like it utilized 'finally:' in the manner that I was looking for. I suspected that something might need to be done to reset() but it was too close to midnight, so I gave up and decided to take the 'comfortable' path to a solution for the problem.

I have torture tested with a Retevis RA85 and RA685. It is nice to not have to power cycle the radio after an error encountered or upload completed.

Thanks for helping me get this working, @kk7ds. This will definitely be my preferred method to use whenever a radio has an 'exit' command to end cloning.

PR #756, opened 2023-08-17
Remove TONE_EXTRA from chirp_common.py - fixes #10795

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -42,9 +42,6 @@ TONES = (
     225.7, 229.1, 233.6, 241.8, 250.3, 254.1,
 )
 
-TONES_EXTRA = (56.0, 57.0, 58.0, 59.0, 60.0, 61.0, 62.0,
-               62.5, 63.0, 64.0)
-
 OLD_TONES = list(TONES)
 [OLD_TONES.remove(x) for x in [159.8, 165.5, 171.3, 177.3, 183.5, 189.9,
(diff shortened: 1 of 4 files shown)


KC9HI:
@kk7ds This is what I have so far. The tones show up in the list but generates and error when the tone is selected ('xx.x' is not in valid list:). As expected, this then fails the driver tests.

I cannot find a single driver that is supposed to have added tones that works. Only the driver where tones are removed seem to be correct. I suspect that when I asked to add 'rf.valid_tones', something got overlooked in the implementation. Adding DTCS codes works fine.

Is this something that you need to look at, or is that still something that is left out of all drivers attempting to add additional CTCSS tones?

kk7ds:
Yep, a little more work required to make the chirp_common stuff agree to let you set custom lists. I'll push up a change in a sec, can you give it a once-over for me?

KC9HI:
I did a quick test. The radios in the drivers that I am working on, RB15/RB615, AT-5888UV and VV898 all display the added tones and the tones can be selected without causing any errors. Yay!

There are about a half dozen drivers with added tones that do not show up. This is because they do not have an rf.valid_tones entry. As a quick test I added this to one of the drivers and it works as expected.

Once these are merged and the dust settles, I will make a point to get the other drivers with extra tones updated so that they will be available for use.

So it is two thumbs up from me.

kk7ds:
Cool, thanks!

PR #761, opened 2023-08-28
radioddity_r2.py: Add Retevis RT24V - fixes #10816

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/radioddity_r2.py
```diff
@@ -319,6 +319,8 @@ class RadioddityR2(chirp_common.CloneModeRadio):
     _range = [400000000, 470000000]
     # maximum 16 channels
     _upper = 16
+    _mem_params = (_upper  # number of channels
```
kk7ds, line 322: If you want this to be a tuple, it needs a comma (and the paren should be on the same line). If you don't want it to be a tuple, no need for the parens at all. One thing I suggested to someone else recently is to make these a dict so you can name them. So:
```
_mem_parms = {'upper': _upper}
```
Then in MEM_FORMAT you can define it like this:
```
MEM_FORMAT = """
  ...
} memory[%(upper)i];
"""
```
Then when you do your `%` in the parse below, it will use the right parameter by name. Makes it easier to define a bunch of params and then you can use them in multiple places in your string *and* reorder the usage of them without having to fix the format. I think we should probably try to standardize on that mechanism so that we don't end up with bugs that creep in because someone reorders the format.

Is that okay? I can also fix this for you if you don't have time.

KC9HI: I want to fix it. That way I will remember better. I'll probably have to study it for a little bit to understand. I will let you know if I need help. Thanks!

kk7ds:
I remembered it was in the ft1d driver. This is an example of one of the named parameters, which he uses in multiple places in the MEM_FORMAT:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/ft1d.py#L208

and this is his list of params:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/ft1d.py#L723

It's used the same way though:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/ft1d.py#L915

It just matches up the named parameters with the items in the dict to use the right one in the right place.

kk7ds:
Looks good thanks!

PR #766, opened 2023-09-07
uv5r.py: work-in-progress - proposed solution for #10505

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/uv5r.py
+++ b/chirp/drivers/uv5r.py
@@ -359,6 +359,19 @@ def _upper_band_from_image(radio):
     return _upper_band_from_data(radio.get_mmap())
 
 
+def _read_from_data(data, data_start, data_stop):
+    data = data[data_start:data_stop]
+    return data
+
+
+def _get_data_from_image(radio, _data_start, _data_stop):
+    image_data = _read_from_data(
+        radio.get_mmap().get_byte_compatible(),
+        _data_start,
+        _data_stop)
+    return image_data
+
+
 def _firmware_version_from_data(data, version_start, version_stop):
     version_tag = data[version_start:version_stop]
     return version_tag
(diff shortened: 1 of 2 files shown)


KC9HI:
@kk7ds I need another set of eyes to look at this. I am trying to get 'Settings -> Other Settings -> Power-On Message' to not pop-up the set_warning() message. It is not stored in Aux memory so it doesn't need the message.

I have been trying to solve this all day. If the solution is not simple, I can move the setting to a different tab or just live with the pop-up message.

Thanks

KC9HI:
You can probably use the Baofeng_UV-5R.img for testing.

kk7ds:
It was an obscure bug in the settings editor itself actually! To be clear, I assume you want to remove the two set_warning() lines either always or some of the time right? AFAICT, in this patch you're always setting it anyway. I commented out the two poweron `set_warning` lines and saw that it *still* warned. That's the problem you were hitting right? Try with the commit I put in front, as I think it should be good now.

KC9HI:
The Power-On Message Line 1/Line 2 are to still get warned. It is only the Power-On Message setting right under them (chooses between Full and Message) that is not supposed to get a warning. But that's OK. I can fix it back after I test. Thanks.

kk7ds:
Ah, okay makes sense. Well, it's the same issue so should be good now.

KC9HI:
Look good to me. I took a lot of radios and a lot of time staring at images to identify the specific radios that have the issue and a workaround that makes them behave.  I found at least 2 in the nearly dozen radios that I acquired. I wanted to iron out this small issue before I attached a copy to the issue for testing. Now I should be able to do that tonight (or possibly tomorrow).

PR #767, opened 2023-09-07
ft1d: fix room number length check

goldstar611:
The ft1d driver would throw an error when editing the Wires-X room number if the room number was less than 5 digits because there was no zero padding.

https://chirp.danplanet.com/issues/10831

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/ft1d.py
+++ b/chirp/drivers/ft1d.py
@@ -1966,9 +1966,9 @@ class FT1Radio(yaesu_clone.YaesuCloneModeRadio):
         val = setting.value.get_value()
         setattr(obj, "name", val)
 
-    def apply_WiresX_roomid(cls, setting, obj):
+    def apply_WiresX_roomid(self, setting, obj):
         val = setting.value.get_value()
-        obj.ID = str(val)
+        obj.ID = self.zero_pad(val, 5)
 
     def apply_WiresX_roomname(cls, setting, obj):
         val = setting.value.get_value()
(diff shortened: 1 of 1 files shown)


chirp/drivers/ft1d.py
```diff
@@ -1966,9 +1966,9 @@ def apply_WiresX_category(cls, setting, obj):
         val = setting.value.get_value()
         setattr(obj, "name", val)
 
-    def apply_WiresX_roomid(cls, setting, obj):
+    def apply_WiresX_roomid(self, setting, obj):
```
kk7ds, line 1969: Yeah, this (and others) should be `self` not `cls`.. Weird.

kk7ds:
IDK about all the things you're zero-padding, but otherwise looks okay to me. I'll hold off on this for a bit as there are a couple other things pending and I'll try to merge all together to avoid a single-small-fix build. Thanks!

chirp/drivers/ft1d.py
goldstar611: I started down the path of fixing other instances but I realized some of the functions were modifying class-scoped variables, so in that aspect, it made sense logically. Plus it had nothing to do with the issue at hand so I only changed this one.

goldstar611:
> IDK about all the things you're zero-padding, but otherwise looks okay to me. I'll hold off on this for a bit as there are a couple other things pending and I'll try to merge all together to avoid a single-small-fix build. Thanks!

I started on a one line fix for the Wires-X zero padding issue but figured there must be an existing zero_pad function to support the other fields. It was `backtrack_zero_pad` so I renamed it to simply `zero_pad` and updated the logic while I was there. This is just function renaming fallout.

Thanks for reviewing and the feedback!

chirp/drivers/ft1d.py
kk7ds: Ack, there's almost no good reason for keeping per-class state so those are almost definitely bugs waiting to be found, but fair enough.

kk7ds:
Oh right, I see, duh. I was just focused on the room change, but glad it's even less change than I was thinking.

goldstar611:
> Oh right, I see, duh. I was just focused on the room change, but glad it's even less change than I was thinking.

It's crazy how a one line change can end up changing 10 more.

PR #768, opened 2023-09-10
rt98.py: fix check_ver() error - fixes #10833

KC9HI:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/retevis_rt98.py
+++ b/chirp/drivers/retevis_rt98.py
@@ -586,7 +586,7 @@ def _ident(radio):
     ver_response = radio.pipe.read(16)
     LOG.debug(util.hexprint(ver_response))
 
-    verok, model, bandlimit = check_ver(radio, ver_response,
+    verok, model, bandlimit = check_ver(ver_response,
                                         radio.ALLOWED_RADIO_TYPES)
     if not verok:
         _finish(radio)


chirp/drivers/retevis_rt98.py
```diff
@@ -586,7 +586,7 @@ def _ident(radio):
     ver_response = radio.pipe.read(16)
     LOG.debug(util.hexprint(ver_response))
 
-    verok, model, bandlimit = check_ver(radio, ver_response,
+    verok, model, bandlimit = check_ver(ver_response,
```
kk7ds, line 589: Hmm, yeah @dforsi changed this recently in a big cleanup, but it seems clearly wrong. Maybe we can get a read on what the intent was? Definitely looks like that was wrong and this is right though.

KC9HI:
I didn't try to check what happened. I thought that I may have unintentionally changed it some how and didn't follow up the test with upload/download using a physical radio. I went back to the CHIRP legacy driver to see what it was matched what I thought it should and made the "correction". But if @dforsi wants to look into it and submit a fix, that is fine with me.

kk7ds:
No, I think this is the right fix, I just want to know why the change amidst a bunch of otherwise typo-like fixes.

KC9HI:
Oh yeah... now I remember the "big cleanup" that you are talking about. Mostly cosmetic things. This change wasn't cosmetic, though.

kk7ds:
Right, which is why I just want @dforsi to comment.

dforsi:
My commit was about undefined names (commit https://github.com/kk7ds/chirp/commit/90cdd61ba2666c05882683ca863934c702d1fe80) but was wrong because somehow I missed to commit a change of the method signature.
I was trying to fix line 591 at that time, which is now line 571 and still needs a fix, the line is `_finish(radio)` but `radio` is undefined inside `check_ver()`.

PR #783, opened 2023-10-04
Add color-picker setting support

kk7ds:


mrtjr1:
Dan - Isi it possible for you to put this into the next build? I am cleaning up the driver that uses it and would like to take out the other method that uses the decimal value for color selection. I would like to release a test driver with the color picker enabled for others to use to make sure a wider user group can check it out.
Thanks!

mrtjr1:
Hold that thought.   For some reason now, the color picker seems to not be working like it did previously.   
I see the correct data and color displayed.. but I can't get it to allow me to change a value.  I click the color displayed to change it, but the color palette doesn't pop up.
image

PR #787, opened 2023-10-16
tox: use distro's default python

yarda:
I think it should used the distro's default python version and not override it, because otherwise e.g. on Fedora Rawhide (f40) it fails:

> py312: failed with env name py312 conflicting with base python python3

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,6 @@ builtins =
     _,
 
 [testenv]
-basepython = python3
 sitepackages = True
 passenv =
   HOME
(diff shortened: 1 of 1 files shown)


yarda:
We are trying to get the package into Fedora repository and this PR originated from this effort:
https://bugzilla.redhat.com/show_bug.cgi?id=1835934

kk7ds:
Ack, this is left over from supporting py2 and py3 in the same place, I think. I'll merge this when some other functional changes come in to avoid generating a build for this user-invisible change.

PR #789, opened 2023-10-19
Tdh8 duplex adds off option

Sandmann34:
Tdh8 duplex adds off option

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -696,7 +696,7 @@ class TDH8(chirp_common.CloneModeRadio):
             "->Tone",
             "DTCS->DTCS"]
         rf.valid_power_levels = TX_POWER
-        rf.valid_duplexes = ["", "-", "+", "split"]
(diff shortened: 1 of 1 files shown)


chirp/drivers/tdh8.py
```diff
@@ -1018,6 +1021,8 @@ def set_memory(self, mem):
             _mem.txfreq = (mem.freq + mem.offset) / 10
         elif mem.duplex == "-":
             _mem.txfreq = (mem.freq - mem.offset) / 10
+        elif mem.duplex == 'off':
+            _mem.txfreq = 166666665
```
kk7ds, line 1025: This value is larger than can be stored in the field, so it gets truncated. It also looks like an overflow value. What does the radio really want here? 0? 0xFFFF?

kk7ds:
I rebased this for you and tagged the bug. If the right thing to do is use 0xFF for the sentinels, please submit another PR (or confirm and I'll do it).

PR #796, opened 2023-10-27
Q10h test

mrtjr1:
# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/kgq10h.py
```diff
@@ -0,0 +1,3203 @@
...
+            _mem.txfreq = int(mem.freq / 10) - int(mem.offset / 10)
+        else:
+            _mem.txfreq = int(mem.freq / 10)
+        _mem.scan_add = int(mem.skip != "S")
+
+        if ((mem.mode == "FM") or (mem.mode == "NFM")):
+            _mem.iswide = int((mem.mode == "FM"))
+            _mem.am_mode = 0
+        elif mem.mode == "AM":
+            _mem.am_mode = 1
```
kk7ds, line 1857: If I edit the memory in the UI and change mode to AM, the UI sends a memory like this to the driver:
```
mem.mode = "AM"
mem.extra["am_mode"] = "OFF"
```
So then on this L1857, you set `am_mode=1`. Then see below.

chirp/drivers/kgq10h.py
```diff
@@ -0,0 +1,3203 @@
...
+        _mem.power = _mem.power & 0x3
+        if mem.power:
+            if _mem.power > 3:
+                _mem.power = 3
+            _mem.power = self.POWER_LEVELS.index(mem.power)
+        else:
+            _mem.power = True
+
+        for setting in mem.extra:
+            setattr(_mem, setting.get_name(), setting.value)
```
kk7ds, line 1872: Right here, when we get to `am_mode`, you set it right back to 0 because the memory's `am_mode` hasn't yet been set. Since you set it to 0 here, then when the UI refreshes the memory after the set operation, your own `get_memory()` is returning `mode=FM`.

If you're doing the edits direct in the spreadsheet with "show extra", you can't set both fields to a consistent value in one operation, but you can in the properties dialog, which lets you make several changes that get sent to the driver all at once.

You can fix the driver to not be broken like this, but I think it's probably not ideal. I think you should probably not make am_mode be a tri-state thing including "Off" and instead just make it be `["RX", "RX+TX"]`. That way the user selects `mode=AM` and `am_mode` separately, which is how it should be anyway. Since `am_mode` can't be stored when FM is selected anyway, if they choose anything for `am_mode` when mode is not AM, that's an error or something to ignore.

You could also only populate am_mode with "off" if mode is set to FM/NFM and populate it with the two choices only when mode=AM. That way there's nothing to select for am_mode *until* mode=AM.

Make sense?

kk7ds (changes requested):
I have some other concerns about some of this driver, but I'll hold off until we resolve this am_mode conversation.

Note that it would be *much* easier on me if you would submit a basic driver with just the cloning and trivial memory stuff implemented and then iterate from there to expand the functionality over time. Asking me to review 3K lines of code in one sitting is a big task, and it also means you don't get any feedback about what you're doing until you've done all that work, which also makes me feel bad about asking for changes. IMHO, it'd be a lot better for you, me, and the users if the basics were already merged and usable and we just needed to discuss and review smaller chunks to add support for mem.extra, settings, etc :)

mrtjr1:
Thanks Dan, I will take a look at what you sent.

I apologize for not sending pieces at a time.    Most of the basics are the
same as other Wouxun drivers.   They mainly just tweaked the checksum
calculation and how the data is stored which was the biggest issue we
talked about before.    I had to read memory from the  radio and then
rearrange the data so all the data is in consecutive order like previous
drivers to enable easy use of structures.    Then I had to sent it back to
the radio in the proper address locations all broken up like the radio
stores it.

I really value your support and don't expect you to be able to approve or
comment completely immediately.   Although you always do seem to find a way
to do that.

At least it looks like I learned how to share a commit with you.  I hope
that helped.  Baby steps.

Mel

kk7ds:
Yep, and that rearranging is what has me (very) concerned that something else is going on. I know we briefly discussed it before, but I think this is the first time I'm seeing the actual implementation of that. It surely looks to me like there *has* to be some endianess confusion going on or something and I'm not sure I think we should put something like this into the tree. But, get your other stuff figured out and then maybe we can deep dive on the massive rearrangement table thing to see if we can figure out what's going on.

mrtjr1:
Ok,

I am having no problems with the data manipulation.   All the rearranging
is working perfectly and is accepted by the radio and the factory CPS
without issue.  No Endianess issues or any other settings/memories issues.
I can compare straight from radio memory and my rearranged memory and
everything looks like it should other than being picked up and moved.
 Eg.  Radio 0x0300 moves to Chirp 0x0000, Radio 0x0200 moves to Chirp
0x0100, 0x0100 to 0x0200 and 0x0000 to 0x0300.  Then they repeat the
pattern starting at radio 0x0700 , 0x0600, 0x0500, 0x0400 etc all the way
to 0x8000.   This was the pattern that I had to decipher to get all the
memory contents to be continuous.  Otherwise part of a channel name could
be at say 0x700-0x704 with the rest at 0x1200-0x1208 as an example.   This
actually happens alot in the data.


There may be better ways to do the manipulation, and I am open to
suggestions, but I am pretty certain the algorithm is working well as long
as the mapping tables are configured correctly.

Anyways, let me get at the AM mode stuff and then we can get all into the
manipulation stuff.

chirp/drivers/kgq10h.py
mrtjr1: So the UI is forcing am_mode to "off"?   I think that is my problem.   am_mode is the only setting available in the radio to turn AM on.   So it seems like the UI is workng against me here.

chirp/drivers/kgq10h.py
mrtjr1: One concern with this approach.   There is only a single setting in memory to determine AM/FM.   It is the am_mode setting.
it has 3 values, Off, AM Rx, AM Rx+Tx.   If am_mode is OFF, then FM mode is used.  So when you say don't make am_mode a tri-state, it is exactly a tri-state setting.   What I was trying to do is when FM is selected in the spreadsheet view -  force am_mode to OFF like the radio expects.    When AM is selected, I wanted to force am_mode to AM Rx.  Then if the user wanted AM Rx+Tx they would choose that in the properties box.   I don't know how to not use am_mode as a tri-state.   What am I missing?

kk7ds: What you're describing is basically my primary suggestion above: if `mode==FM`, then am_mode is off regardless. No matter what they choose for am_mode in the properties or spreadsheet view, you ignore that if `mode==FM`. Then if `mode==AM`, you default to whichever am_mode makes sense, and then let them choose between those two sub-modes. It's the same behavior as was discussed on the list for the FT450, fwiw.

chirp/drivers/kgq10h.py
kk7ds: The UI is not forcing anything. In spreadsheet view only one thing can change at a time. Meaning the user can only change mode, or am_mode, in a given edit and you will receive a `set_memory()` each time something changes. Your driver is the only thing that can (and should) enforce that things are consistent in a way that the radio requires. The UI has no idea that those two "columns" are related or dependent. In this way, making sure it works both in the spreadsheet and the properties view helps to ensure that you're doing the right thing in your driver, and not requiring the human user to *know* which combinations work and don't work when they're using the properties dialog.

chirp/drivers/kgq10h.py
mrtjr1: I am confused.. The code does exactly what we are talking about.  But then after executing the IF / ELIF statement on MODE, and setting  am_mode correctly,

        if ((mem.mode == "FM") or (mem.mode == "NFM")):
            _mem.iswide = int((mem.mode == "FM"))
            _mem.am_mode = 0
        elif mem.mode == "AM":
            _mem.am_mode = 1

the following code is run 
        for setting in mem.extra:
  
which then sets am_mode back to what it was previously.   

If I remove the code
        for setting in mem.extra:

then the mem.mode value is updated as I want it to be through the spreadsheet UI, but then I can't change any of the Extra settings and get them to take effect.

What am I missing that gets the 2 pieces of code to play nice together?

mrtjr1: I added this to the code and it seems to be doing what I want.
When mode != AM -  the mem.extra am_mode is not shown in the properties and it is forced to 0.
When mode = AM, the mem.extra allows selection of AM Rx or AM Rx+Tx
Switching from FM/NFM to AM or AM to FM/NFM must be done via the spreadsheet UI.
The properties diaglog box can only change AM modes when AM is selected via the spreadsheet UI mem.mode setting.

        for setting in mem.extra:
            if setting.get_name() != "am_mode":
            else:
                if mem.mode != "AM":
                    setattr(_mem, setting.get_name(), 0)
                elif int(setting.value) == 2:
                    setattr(_mem, setting.get_name(), 2)

Although I think I need to add another elif to handle switching from value 2 back to default value 1.

                elif int(setting.value) == 1:
                    setattr(_mem, setting.get_name(), 1)

mrtjr1: with the above changes - all now pass Tox which was previously failing form FM/NFM/AM mode switching.

  style: OK (64.62=setup[27.91]+cmd[36.72] seconds)
  unit: OK (41.86=setup[12.97]+cmd[28.89] seconds)
  driver: OK (557.69=setup[13.22]+cmd[544.47] seconds)
  congratulations :) (664.42 seconds)

mrtjr1:
> Yep, and that rearranging is what has me (very) concerned that something else is going on. I know we briefly discussed it before, but I think this is the first time I'm seeing the actual implementation of that. It surely looks to me like there _has_ to be some endianess confusion going on or something and I'm not sure I think we should put something like this into the tree. But, get your other stuff figured out and then maybe we can deep dive on the massive rearrangement table thing to see if we can figure out what's going on.

Here is what the data on the radio looks like compared to what it looks like once rearranged in chirp.
File on left is a straight 0x0000 to 0x8000 read from radio.   On the right is what it looks like after rearrangment.
you can see in the original, data jumps around every 256 bytes.  Sometimes in the middle of a piece of data (like Ch Name).  
image

kk7ds:
> Here is what the data on the radio looks like compared to what it looks like once rearranged in chirp. File on left is a straight 0x0000 to 0x8000 read from radio. On the right is what it looks like after rearrangment. you can see in the original, data jumps around every 256 bytes. Sometimes in the middle of a piece of data (like Ch Name). 

Right, so "every 256 bytes" is exactly what I'd expect if you're using the wrong byte order in your address. If you switch your addressing to little-endian does it all line up without rearranging?

mrtjr1:
> > Here is what the data on the radio looks like compared to what it looks like once rearranged in chirp. File on left is a straight 0x0000 to 0x8000 read from radio. On the right is what it looks like after rearrangment. you can see in the original, data jumps around every 256 bytes. Sometimes in the middle of a piece of data (like Ch Name).
> 
> Right, so "every 256 bytes" is exactly what I'd expect if you're using the wrong byte order in your address. If you switch your addressing to little-endian does it all line up without rearranging?

Where would I do that?   My commands directly copy what the CPS does.   Using the same addressing format, I get the same data.   I would be willing to try, but would your thoughts explain why the go from  300 to 400 and then 200 to 300, then 100 to 200 and then 0 to 100 followed by 700 to 800, then 600 to 700, then 500 to 600 then 400 to 500  etc?

kk7ds:
> Where would I do that?

In your struct.pack, change `>` to `<`

> My commands directly copy what the CPS does. Using the same addressing format, I get the same data. I would be willing to try, but would your thoughts explain why the order goes from 300 to 400 and then 200 to 300, then 100 to 200 and then 0 to 100 followed by 700 to 800, then 600 to 700, then 500 to 600 then 400 to 500 etc?

Yes. If the OEM software is pulling them in non-sequential order, they're likely byte-swapping before inserting into their memory map.

mrtjr1:
> > Where would I do that?
> 
> In your struct.pack, change `>` to `<`
> 
> > My commands directly copy what the CPS does. Using the same addressing format, I get the same data. I would be willing to try, but would your thoughts explain why the order goes from 300 to 400 and then 200 to 300, then 100 to 200 and then 0 to 100 followed by 700 to 800, then 600 to 700, then 500 to 600 then 400 to 500 etc?
> 
> Yes. If the OEM software is pulling them in non-sequential order, they're likely byte-swapping before inserting into their memory map.

Simply swapping the address endianness and not rearranging did not return a complete usable image.  It did communicate with the radio, but it was pretty much the same data repeating over and over and just shifting in some bytes and shifting out others.  It never got to any channel data and all 999 channels were populated with data and names in the radio.

            for i in range(start, end, blocksize):
                time.sleep(0.005)
                req = struct.pack('<HB', i, blocksize)
                self._write_record(CMD_RD, req)

I have seen this repeating pattern before when I was trying to see how much more memory previous wouxun radios had above 7FFF.   Once above 7FFF, they just started repeating the data at 0 again.
[Wouxun_KG-Q10H_20231101 little endianess download test.zip](https://github.com/kk7ds/chirp/files/13232640/Wouxun_KG-Q10H_20231101.little.endianess.download.test.zip)

downloading with big endianess without rearranging yielded the following:

            for i in range(start, end, blocksize):
                time.sleep(0.005)
                req = struct.pack('>HB', i, blocksize)
                self._write_record(CMD_RD, req)

[Wouxun_KG-Q10H_20231101 big endianess download test.zip](https://github.com/kk7ds/chirp/files/13232642/Wouxun_KG-Q10H_20231101.big.endianess.download.test.zip)

You can see with the big endian download, all the data is there just interspersed on 256byte boundaries.
Did I do something wrong in the addressing when switching to little endian?  
I know you think that may be a simpler explanation, but I am pretty confident that what I am doing is what they are doing and it is an attempt to cause issues with 3rd party programmers.   My address commands are their address commands and the data sent and received is the same.  
When I emulated the CPS command order (with big endianess addressing) and just filled the map with the data as received (no rearranging),  It also generated a usable image.  The problem with that was that there were gaps as the CPS does not read all 32K bytes.  It only reads the settings, then jumps to the channel mem, then the ch names, then the valid bytes, etc.   So if new settings ( like freq limits ) were discovered it would mess up the map and older images would be incompatible with the new memmap.   By doing what I am doing, I have all 32768 bytes in every image and all I need to do is add to the memmap as things are discovered.

mrtjr1:
I looked a bit deeper at the radio responses when sending the Little Endian Addresses -   It does not appear that the radio is swapping the bytes before replying.  Any address that is >= 0x8000 gets all 0xDD responses from the radio.  All the other data is consistent with the data that is stored in Big Endian address format and it explains the repeating pattern shifted by 1 byte as seen in the image.

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/melvi/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/melvi/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



Address   Dec | Address Hex     Big End | Address Hex     Little End
-- | -- | --
0 | 0000 | 0000
64 | 0040 | 4000
128 | 0080 | 8000
192 | 00C0 | C000
256 | 0100 | 0001
320 | 0140 | 4001
384 | 0180 | 8001
448 | 01C0 | C001
512 | 0200 | 0002
576 | 0240 | 4002
640 | 0280 | 8002
704 | 02C0 | C002
768 | 0300 | 0003
832 | 0340 | 4003
896 | 0380 | 8003
960 | 03C0 | C003



</body>

</html>





























































image

PR #804, opened 2023-11-05
ft450: Fix IF filter setting breaking UI

kk7ds:
This driver exposed a different setting for each memory in extra
depending on the mode of that memory. That's not compatible with the
UI's exploding of extra fields into columns, and in general breaks
the ability to edit a whole memory in the properties dialog.

This makes it expose one setting with all the available options, but
ensures that only filters valid for a given mode are selected by
rounding to the nearest available filter width per memory.

Fixes #10929

### DIFF
--- a/chirp/drivers/ft450d.py
+++ b/chirp/drivers/ft450d.py
@@ -26,6 +26,7 @@ from chirp.settings import RadioSetting, RadioSettingGroup, \
 import time
 import struct
 import logging
+import itertools
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


kk7ds:
cc @WCurlew and @Old-Phart ... This is my proposed fix for bug 10929, let me know what you think. It'd be nice if at least one of you can test this too, of course. Bill, I will hold off on merging this until we get your PR #803 done and then I'll fix the conflicts.

WCurlew:
I did give this a quick run through with the FT450D image and it seemed OK, can't test FT450 as PY not updated for that, but don't see why it wouldn't work, common code.

kk7ds:
This has the ft450 changes in it for testing, but will have them rebased out when that merges (first).

PR #805, opened 2023-11-06
Bitwise strict checks and byte-compat deprecations

kk7ds:
- Add a way to test for compat bytes usage
- Default bitwise's get_raw() to byte-native
- Fix deprecated __trunc__ instead of __int__
- Add warnings for non-byte-clean raw methods
- Add test for strict bitwise seeks


kk7ds:
Unit failure is because repeaterbook is down right now.

PR #807, opened 2023-11-08
Fix some non-byteclean drivers

kk7ds:
- bitwise: Support string-to-char[] in some cases
- boblov_x3plus: Fix raw bytestring handling
- leixen: Fix raw bytestring handling
- puxing: Fix raw bytestring handling
- tk8102: Fix raw bytestring handling
- vx2,3,7: Fix raw bytestring handling
- vx8: Fix raw bytestring handling (et al)


goldstar611:
Now that CHIRP is python3 only, would you be opposed on adding typing information to the base/core functions where possible?

something like going from
`    def set_raw(self, data):`
to
`    def set_raw(self, data: bytes):`
or even
`    def set_raw(self, data: Union[bytes, CustomBytesClass]):`

When I first started looking at python3 code (coming from python2) I hated it, but having the hints make using an IDE much more productive.

kk7ds:
Yeah I dunno. I appreciate the type hints in my editor as well, but I hate them in my code :( Python was supposed to be duck typed from the beginning, and so type annotations especially ("either this or that completely different type") is pretty silly.

But, lots of work to do on all the drivers before we get squared away anyway.

PR #808, opened 2023-11-11
Issue #10943

Old-Phart:
Removed unnecessay seekto in MEM_FORMAT
Editted Info, pre-download and pre-upload prompts.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/th_uv8000.py
+++ b/chirp/drivers/th_uv8000.py
@@ -94,13 +94,10 @@ struct bitmap {
 #seekto 0x0010;
 struct chns chan_mem[128];
 
-#seekto 0x1010;
 struct frqx frq[2];
 
-#seekto 0x1050;
 struct fm_chn fm_stations[25];
 
-#seekto 0x1080;
 struct {
   u8  fmunk01[14];
   ul16 fmcur;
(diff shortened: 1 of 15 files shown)


chirp/drivers/th_uv8000.py
```diff
@@ -518,21 +512,20 @@ class THUV8000Radio(chirp_common.CloneModeRadio):
...
-             'editor to see/set the upper and lower frequency-mode values.\n')
+            ('Remember the # T-R key is the VFO/MEM toggle.\n')
 
         rp.pre_download = _(
             "Follow these instructions to download the radio memory:\n"
             "1 - Turn off your radio\n"
             "2 - Connect your interface cable\n"
             "3 - Turn on your radio, volume @ 50%\n"
-            "4 - Radio > Download from radio\n")
+            "4 - CHIRP Menu - Radio - Download from radio\n")
```
kk7ds, line 522: This doesn't really make sense I think because this is the message shown to the user *after* they've already clicked on the thing described in item 4 above. But, by the time this message is displayed, we won't have opened the serial port or anything, so I think it's fine to just say "4 - Click OK" right?

kk7ds:
You can ignore the "PR checks" complaint about the locale stuff, I can update that for you.

kk7ds:
Thanks, I'll merge this soon when we've got a few other things to apply.

chirp/drivers/th_uv8000.py
Old-Phart: Oops, step 4 should just be "OK".
You can edit those if you like, but I just pushed a new commit for it.

kk7ds:
Sorry, I didn't realize there were two things fixed in here when I squashed your patches and regenerated the locale stuff. It would have been better to keep them separate like you had it, I just didn't notice until it was too late. Thanks!

PR #811, opened 2023-11-16
Added baofeng_uv17Pro driver

vdwel:
This driver adds support for the UV 17Pro/R/L/GPS models with the black screen.

# CHIRP PR Checklist

* All tests must be passing.


vdwel:
I am wondering if the failures in the PR Checks are a problem?

kk7ds:
I just noticed the recently-introduced bug in that script which is causing it to think you don't have an image. I'll fix that so you can ignore that message. However, the "shares at least 87%..." error is real :)

vdwel:
Is the "shares at least 87%..." error, a problem for the pull request? Why completely write something new if there is already some good code available.. ;-)

kk7ds:
Yes, it's a problem. It's complaining because you copied that file and changed some things, which duplicates code in the tree, increases the maintenance overhead long-term, and means bugs have to be fixed in multiple places. Please import and inherit from that base class like all the other similar drivers:
```
$ grep -r '(baofeng_common.' chirp/drivers
chirp/drivers/mursv1.py:class MURSV1(baofeng_common.BaofengCommonHT):
chirp/drivers/baofeng_wp970i.py:class WP970I(baofeng_common.BaofengCommonHT):
chirp/drivers/gmrsuv1.py:class GMRSV1(baofeng_common.BaofengCommonHT):
chirp/drivers/uv6r.py:class UV6R(baofeng_common.BaofengCommonHT):
chirp/drivers/uv5x3.py:class UV5X3(baofeng_common.BaofengCommonHT):
chirp/drivers/gmrsv2.py:class GMRSV2base(baofeng_common.BaofengCommonHT):
```
The `baofeng_common.py` file is specifically not a driver on its own so it can be re-used by other drivers to avoid the duplication. I don't want you to write something new, I want you to use what's there to *avoid* writing something new and avoid copying what's there to effectively do the same thing. That check is there because people tend to come in, copy one driver to another file, make a bunch of tiny changes all over, and submit. That causes huge amounts of additional maintenance burden years from now when a bug has to be fixed or python4 comes out.

CHIRP is a 15-year-old project, and maintaining all that code long-term is done by basically two people. Please respect our (future) time and help us by not duplicating code we'll have to support and fix later.

Thanks!

vdwel:
I'll see what I can do.. It seems like a hell of a job to fix this.

vdwel:
I had a quick look at what I did. And I do completely understand your point. When I started this "project", I didn't know anything about Chirp and its structure yet... I just needed support for my radios :-). I will fix it.

vdwel:
@kk7ds 
I fixed the duplicate functions from baofeng_common, but now I get the message that there is 82% overlap with gmrsv2.py, which I really don't understand, since I never even looked at that file.

kk7ds:
Well, likely because that driver is very similar to yours and even more re-use could be achieved :) I'll look later today and see. Thanks for your efforts...

vdwel:
Interesting. I did a wdiff myself and this is the result:
baofeng_uv17Pro.py: 4079 words  906 22% common  583 14% deleted  2590 63% changed
gmrsv2.py: 3123 words  906 29% common  188 6% inserted  2029 65% changed

kk7ds:
Yeah same here the script shows it's 62% different. I've never had trouble with that before, but let me play around with it a bit.

kk7ds:
First off, wdiff behaves differently on macos than linux. Using `-i` makes it almost identical, so I'm guessing there's some platform-specific definition of "word" or "same". Pretty dumb, but locally on linux it behaves identical to the github action, and different on macos. :shrug:

Second, I think the false positive is because it sees the change between your driver and `gmrsv2` as a bunch of "deleted" things and then of what's left, a lot of "changed". So I changed it to look at the "how much is common" stat which should make more sense. You're the first to trip over that, I dunno why, but...

So, I shoved a few commits underneath yours to make those changes. Just keep them in your stack if you would and we'll merge them all together.

I still haven't reviewed the actual code and it's dinner time, so maybe tomorrow. We'll need to squash your commits down to just one per the guidelines. I also added another one about making them meaningful to the users, because "update $file again" going to the list is not helpful :)

Thanks!

vdwel:
Thanks a lot for your help!! I am not a GitHub wizard yet, nor am I a "real" programmer. So I am a littlebit nervous for the code review. :-)

vdwel:
Actually, I hope you can help with the squashing and renaming of the commits to something meaningful. I am afraid I would mess things up.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+            return chirp_common.format_freq(real_freq * 10)
+
+        def my_validate(value):
+            value = chirp_common.parse_freq(value)
+            freqOk = False
+            for band in self.VALID_BANDS:
+                if value > band[0] and value < band[1]:
+                    freqOk = True
+            if not freqOk:
+                raise InvalidValueError("Invalid frequency!")
```
kk7ds, line 908: Doing this will make it so that the user will not be able to see any settings at all and will just get this error when they click on that tab. It won't tell them which frequency is invalid and there will be no way for them to fix it. You should probably just display it anyway, log something, etc. Perhaps just change it to the lower bound of VALID_BANDS?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+        work.append(rs)
+
+        rs = RadioSetting("vfo.b.rxtone", "VFA B RX QT/DQT",
+                          RadioSettingValueList(self.RXTX_CODES,
+                                                getToneIndex(
+                                                    _mem.vfo.b.rxtone)))
+        rs.set_apply_callback(apply_rxtone, _mem.vfo.b)
+        work.append(rs)
+
+        rs = RadioSetting("vfo.b.txtone", "VFA B TX QT/DQT",
```
kk7ds, line 1088: Seeing all this - shouldn't you expose the VFO memories as special channels so they show up in the spreadsheet instead of as a setting? I know some radios do it this way (probably incorrectly) but you're reimplementing a *ton* of logic here which is sort of free if these are exposed as memories.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+                    mem.offset = offset
+                    mem.duplex = "+"
+            else:
+                mem.offset = 0
+
+        for char in _mem.name:
+            if str(char) == "\xFF":
+                char = " "  # The OEM software may have 0xFF mid-name
+            mem.name += str(char)
+        mem.name = mem.name.rstrip()
```
kk7ds, line 1241: You don't need to iterate for this. I think this whole loop is equivalent to:
```
mem.name = str(_mem.name).replace('\xFF', '').rstrip()
```
right?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+
+        if txmode == "Tone" and not rxmode:
+            mem.tmode = "Tone"
+        elif txmode == rxmode and txmode == "Tone" and mem.rtone == mem.ctone:
+            mem.tmode = "TSQL"
+        elif txmode == rxmode and txmode == "DTCS" and mem.dtcs == mem.rx_dtcs:
+            mem.tmode = "DTCS"
+        elif rxmode or txmode:
+            mem.tmode = "Cross"
+            mem.cross_mode = "%s->%s" % (txmode, rxmode)
```
kk7ds, line 1296: I think you want to use `chirp_common.split_tone_decode()` here to make sure you're consistent with everything else.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+
+        mem.dtcs_polarity = "".join(dtcs_pol)
+
+        mem.extra = RadioSettingGroup("Extra", "extra")
+
+        rs = RadioSetting("bcl", "BCL",
+                          RadioSettingValueBoolean(_mem.bcl))
+        mem.extra.append(rs)
+
+        if (_mem.pttid != 0xFF):
```
kk7ds, line 1306: Here it is *critical* that you not expose different `mem.extra` structures for different memories. You're probably seeing warnings in the logs because of this.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+                                                    self.SCODE_LIST[
+                                                        _mem.scode]))
+            mem.extra.append(rs)
+
+        return mem
+
+    def set_memory(self, mem):
+        _mem = self._memobj.memory[mem.number]
+
+        _mem.set_raw("\x00"*16 + "\xff" * 16)
```
kk7ds, line 1325: These need to be `bytes` now. I haven't enabled strict mode for this yet, but it's coming.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+            _mem.txfreq = (mem.freq - mem.offset) / 10
+        else:
+            _mem.txfreq = mem.freq / 10
+
+        _namelength = self.get_features().valid_name_length
+        for i in range(_namelength):
+            try:
+                _mem.name[i] = mem.name[i]
+            except IndexError:
+                _mem.name[i] = "\xFF"
```
kk7ds, line 1350: `name = name.ljust(_namelength, '\xFF')`

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+            else:
+                _mem.rxtone = 0
+        else:
+            _mem.rxtone = 0
+            _mem.txtone = 0
+
+        if txmode == "DTCS" and mem.dtcs_polarity[0] == "R":
+            _mem.txtone += 0x69
+        if rxmode == "DTCS" and mem.dtcs_polarity[1] == "R":
+            _mem.rxtone += 0x69
```
kk7ds, line 1384: Same here, `split_tone_decode()`

kk7ds (changes requested):

kk7ds:
I can help with some (maybe all) of this, but I won't be able to test the result of course. Need answers to some of the questions at least though.

vdwel:
I will have a look at your comments tomorrow. Thanks for your review. I'll try to fix them myself. I think it's a good exercise for me. :-)

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+    # Some weird encryption is used. From the table below, we only use "CO 7".
+    tblEncrySymbol = [b"BHT ", b"CO 7", b"A ES", b" EIY", b"M PQ",
+                      b"XN Y", b"RVB ", b" HQP", b"W RC", b"MS N",
+                      b" SAT", b"K DH", b"ZO R", b"C SL", b"6RB ",
+                      b" JCG", b"PN V", b"J PK", b"EK L", b"I LZ"]
+    tblEncrySymbols = tblEncrySymbol[symbolIndex]
+    decBuffer = b""
+    index1 = 0
+    for index2 in range(len(buffer)):
+        boolEncryptChar = ((tblEncrySymbols[index1] != 32) &
```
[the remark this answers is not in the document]
vdwel: Yes, it's a bitwise and. It's the baofeng encryption routine, :-S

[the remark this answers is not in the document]
vdwel: hmmm reading it.. I am actually not sure.. :S Probably "and" will also work.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+    if len(data) != 0:
+        frame += data
+    # return the data
+    return frame
+
+
+def _do_ident(radio):
+    """Put the radio in PROGRAM mode & identify it"""
+    radio.pipe.baudrate = radio.BAUDRATE
+    radio.pipe.parity = "N"
```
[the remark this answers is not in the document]
vdwel: I do need to set the baudrate.. What is a better way to set it to 115200?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+    radio.pipe.timeout = STIMEOUT
```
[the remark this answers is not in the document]
vdwel: Removed the timeout and parity from the code. Everything still works fine. Can't remove the BAUDRATE, because need to set it to 115200.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+
+    def process_mmap(self):
+        """Process the mem map into the mem object"""
+        # make lines shorter for style check.
+        check1 = (len(self._mmap) == self.MEM_TOTAL)
+        check2 = (len(self._mmap) == self.MEM_TOTAL + len(self.MODEL))
+        if (check1 or check2):
+            self._memobj = bitwise.parse(self.MEM_FORMAT, self._mmap)
+        else:
+            raise errors.ImageDetectFailed('Image length mismatch.\n'
```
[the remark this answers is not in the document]
vdwel: I think it was put here in an early stage of the driver development. I discovered that I needed to load more data. People that were already testing the driver, had image versions which were too short and incompatible. I think we we remove this check now.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+        def getKey2shortIndex(value):
+            if value == 0x07:
+                return 0
+            if value == 0x1C:
+                return 1
+            if value == 0x1D:
+                return 2
+            if value == 0x2D:
+                return 3
+            return 0
```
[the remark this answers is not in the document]
vdwel: Thanks.. I learned something. :-)

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+            if val == "FM":
+                obj.key2short = 0x07
+            if val == "Scan":
+                obj.key2short = 0x1C
+            if val == "Search":
+                obj.key2short = 0x1D
+            if val == "Vox":
+                obj.key2short = 0x2D
+
+        if self._skey2_short:
```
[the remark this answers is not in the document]
vdwel: That is because the other radio supported by this driver, does not have this function... I'll try to find another solution.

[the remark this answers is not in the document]
vdwel: I can't think of a way to make this nice. I did change the variable name to _has_skey2_short. Hope that's sufficient. Other suggestions are welcome.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+            charset = chirp_common.CHARSET_ASCII
+            for char in name:
+                if ord(str(char)) == 255:
+                    break
+                if str(char) not in charset:
+                    char = "X"
+                fname += str(char)
+            if fname == "XXXXXX":
+                fname = "ZONE" + zone
+            return fname
```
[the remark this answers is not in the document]
vdwel: The problem here is, that the default bank names are in Chinese. I am changing the bank names to ZONE x when they are in Chinese.

Do you have a better suggestion?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+
+STIMEOUT = 1.5
+
+
+def model_match(cls, data):
+    """Match the opened image to the correct version"""
+    return data[cls.MEM_TOTAL:] == bytes(cls.MODEL, 'utf-8')
+
+
+def _crypt(symbolIndex, buffer):
```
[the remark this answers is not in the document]
vdwel: Fixed

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+                              RadioSettingValueString(
+                                0, 12, _filterName(_msg.name9, _zone)))
+            bank.append(rs)
+
+            _zone = "10"
+            _msg = _mem.bank_names
+            rs = RadioSetting("bank_names.name10", "Bank name 10",
+                              RadioSettingValueString(
+                                0, 12, _filterName(_msg.name10, _zone)))
+            bank.append(rs)
```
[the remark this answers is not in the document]
vdwel: Done.

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
kk7ds: You're and-ing booleans, so it really should never be anything other than one or zero. That makes it effectively a logical operation, so using `and` would be more correct to make it clear what's going on.

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
kk7ds: It's set by the UI for you according to `radio.BAUDRATE` before it hands the serial device to the driver. So this should already be done for you, no need to re-do it (which resets the stack on Windows).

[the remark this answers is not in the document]
vdwel: Somehow that doesn't work... hmmm..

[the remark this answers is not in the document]
kk7ds: It happens here: https://github.com/kk7ds/chirp/blob/master/chirp/wxui/clone.py#L135 .. Very few of the drivers manipulate the baudrate because it's already set for them. The exceptions are mostly drivers that probe for one of multiple allowed rates that the radio could be set to. So yeah, need to figure out what's going on there.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+        rs = RadioSetting("ani.dtmfoff", "DTMF Speed (off)",
+                          RadioSettingValueList(LIST_DTMFSPEED,
+                                                LIST_DTMFSPEED[val]))
+        dtmfe.append(rs)
+
+        _codeobj = self._memobj.ani.code
+        _code = "".join([DTMF_CHARS[x] for x in _codeobj if int(x) < 0x1F])
+        val = RadioSettingValueString(0, 5, _code, False)
+        val.set_charset(DTMF_CHARS)
+        rs = RadioSetting("ani.code", "ANI Code", val)
```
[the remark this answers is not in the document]
vdwel: True.. :-) Removed duplicate one.

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
vdwel: Resolved. I used BAUDRATE which had to be BAUD_RATE

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+                                                        _mem.ani.aniid]))
+            dtmfe.append(rs)
+
+        rs = RadioSetting("settings.hangup", "Hang-up time",
+                          RadioSettingValueList(LIST_HANGUPTIME,
+                                                LIST_HANGUPTIME[
+                                                    _mem.settings.hangup]))
+        dtmfe.append(rs)
+
+        def convert_bytes_to_freq(bytes):
```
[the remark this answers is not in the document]
vdwel: It looks like BCD, but I think it's different. Each decimal digit in the frequency is 1 byte here. With bcd, each two digits is one byte.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1427 @@
...
+      u8 unknown5[2];
+      u8 sqmode;
+      u8 unknown6[3];
+    };
+
+    #seekto 0x8000;
+    struct {
+      struct vfo a;
+      struct vfo b;
+    } vfo;
```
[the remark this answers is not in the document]
vdwel: renamed one of the structs.

chirp/drivers/baofeng_uv17Pro.py
vdwel: The way it is coded now, it just reverts to the frequency that was already in the vfo field. It does not affect the settings tab. If the error is raised, the field reverts to its previous value. It's done this way in many baofeng drivers.
I did make the error more verbose now.

chirp/drivers/baofeng_uv17Pro.py
vdwel: They implemented the frequency mode settings, slightly different in the internal memory of the radio. I don't understand why they did it this way. The frequency mode entries differ too much from the normal memory entries to treat them the same way.

chirp/drivers/baofeng_uv17Pro.py
vdwel: fixed. thnx

chirp/drivers/baofeng_uv17Pro.py
vdwel: Again, gained a lot of efficiency with code reuse. thnx

chirp/drivers/baofeng_uv17Pro.py
vdwel: No warnings, but I fixed it.

chirp/drivers/baofeng_uv17Pro.py
vdwel: Fixed

chirp/drivers/baofeng_uv17Pro.py
vdwel: fixed

chirp/drivers/baofeng_uv17Pro.py
vdwel: I assume you mean "split_tone_encode". Fixed it. :-)

vdwel:
I have implemented all the requested changes. There are still some open requests, where I have some questions. Could you have a look at the questions and the implementation of the changes? Thank you very much.

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
kk7ds: I didn't realize this was a feature flag for the two models when I wrote that, so it's okay.

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
kk7ds: How are they encoded? If it's UTF-8 we should decode that and show it properly (so it doesn't break for chinese people).

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
kk7ds: Okay, then can you move these up to module scope? Also, the freq and offset variants can be the same and you can just scale the result when you use it right? I think having two functions for the same thing other than the multiplier is a little heavy.

chirp/drivers/baofeng_uv17Pro.py
kk7ds: You're assuming you only hit that error if the user enters an incorrect frequency. However, if it's already bad in the radio, then you'll break everything. We see a ton of examples of the chinese radios coming from the factory with seemingly corrupt defaults and as you can see from the raft of bugs we have, it's very annoying for people to get no settings and be unable to fix them. So please just coerce the value here. Even if they put in something invalid and have it snap back to the nearest value, that should be okay. I again said these VFO things should probably be special channels so the UI will do all the normal valid_bands enforcement for you, like this:

image

Personally I see zero reason to go to the work of exposing the VFOs in the programming software. I've never once felt like it would be helpful to do that in my own programming. So if it were me, I'd nix all that stuff from here.

chirp/drivers/baofeng_uv17Pro.py
kk7ds: No, because you don't have to implement it with the same code. You can have different code for different-formatted memories. The FT450 driver, for example, has to do quite a bit of different stuff for the VFO channels:

See this for example:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/ft450d.py#L628

..or as above, just don't expose the VFOs at all because I don't think there's much point.

kk7ds:
How about you remove the VFOs from the driver for this PR so we can iterate on the less controversial stuff and then we can add them back in one form or another after?

chirp/drivers/baofeng_uv17Pro.py
vdwel: Fixed: removed VFO settings.

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
vdwel: Fixed: removed VFO settings. So this function becomes obsolete. :-)

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
vdwel: They are encoded in GB2312. I implemented the conversion en viewing for this encoding. One of the problems I have still left is that this encoding is 1 byte for regular non-chines characters and 2 bytes for Chinese characters. I solved this now by just limiting the encoded string at 16 bytes.

vdwel:
I think we are almost there. :-)

kk7ds:
Sorry, I still haven't looked at this, some life stuff came up, but I promise I haven't forgotten.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1039 @@
...
+LIST_ID_DELAY = ["%s ms" % x for x in range(100, 3100, 100)]
+LIST_QT_SAVEMODE = ["Both", "RX", "TX"]
+LIST_SKEY2_SHORT = ["FM", "Scan", "Search", "Vox"]
+LIST_RPT_TAIL_CLEAR = ["%s ms" % x for x in range(0, 1100, 100)]
+LIST_VOX_DELAY_TIME = ["%s ms" % x for x in range(500, 2100, 100)]
+LIST_VOX_LEVEL = ["Off"] + ["%s" % x for x in range(1, 10, 1)]
+LIST_VOX_LEVEL_ALT = ["%s" % x for x in range(1, 10, 1)]
+LIST_GPS_MODE = ["GPS", "Beidou", "GPS + Beidou"]
+LIST_GPS_TIMEZONE = ["%s" % x for x in range(-12, 13, 1)]
+CHARSET_GB2312 = chirp_common.CHARSET_ASCII
```
kk7ds, line 82: Are you just doing this to limit the charset for the setting string? If so, I think it would be better (i.e. more conventional) to just let it be a string (which is unicode) and then encode it to gb2312 when we store it. If they use a character that isn't supported, we can let the UnicodeEncodeError bubble up to the user to alert them.

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
kk7ds: I think `decode('gb2312') should handle this properly, but you need to collect all the bytes (as `bytes`) and then call decode on it.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1039 @@
...
+for x in range(0xB0, 0xD7):
+    for y in range(0xA1, 0xFF):
+        CHARSET_GB2312 += bytes([x, y]).decode('gb2312')
+
+
+TXP_CHOICES = ["High", "Low"]
+TXP_VALUES = [0x00, 0x02]
+
+STIMEOUT = 1.5
```
kk7ds, line 91: I think this is no longer used?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1039 @@
...
+    VENDOR = "Baofeng"
+    MODEL = "UV-17Pro"
+    NEEDS_COMPAT_SERIAL = False
+
+    MEM_STARTS = [0x0000, 0x9000, 0xA000, 0xD000]
+    MEM_SIZES = [0x8040, 0x0040, 0x02C0, 0x0040]
+
+    MEM_TOTAL = 0x8380
+    BLOCK_SIZE = 0x40
+    STIMEOUT = 2
```
kk7ds, line 261: Not used?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1039 @@
...
+    def process_mmap(self):
+        """Process the mem map into the mem object"""
+        # make lines shorter for style check.
+        self._memobj = bitwise.parse(self.MEM_FORMAT, self._mmap)
+
+    def get_settings(self):
+        """Translate the bit in the mem_struct into settings in the UI"""
+        _mem = self._memobj
+        basic = RadioSettingGroup("basic", "Basic Settings")
+        bank = RadioSettingGroup("bank", "Bank names")
```
kk7ds, line 474: So I wasn't really thinking about this earlier, but putting the bank names in the settings is not really right. You should expose `has_bank` and `has_bank_names` and implement a `BankModel`. That will let you do the actual bank assignments, but it also has provisions to allow changing the name of the banks in the bank editor. You can look at some of the other radios that support bank names for examples.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -0,0 +1,1039 @@
...
+
+    _tri_band = True
+    _fileid = []
+    _magic = MSTRING_UV17L
+    _magic_response_length = 1
+    _fingerprint = b"\x06"
+    _magics = [b"\x46", b"\x4d",
+               b"\x53\x45\x4E\x44\x21\x05\x0D\x01\x01\x01\x04\x11\x08\x05" +
+               b"\x0D\x0D\x01\x11\x0F\x09\x12\x09\x10\x04\x00"]
+    _magicResponseLengths = [16, 15, 1]
```
kk7ds, line 276: Missed a camelCase

kk7ds:
Thanks for your work on this. I think I'll go ahead and merge this, and hope you'll circle back with some of the other cleanups. I think we should try to get the bank names moved out to a bank model, and the bank name encode/decode from the full bytestring so it works properly.

chirp/drivers/baofeng_uv17Pro.py
vdwel: I am defining the GB2312 charset here and the following lines. Later I use it for displaying the Chinese characters in the bank names. It was a hell of a job to figure out how to make this work. I have no clue how to make this better.

chirp/drivers/baofeng_uv17Pro.py
vdwel: Removed

chirp/drivers/baofeng_uv17Pro.py
vdwel: Removed STIMEOUT. The other stuff is used

chirp/drivers/baofeng_uv17Pro.py
vdwel: Fixed

chirp/drivers/baofeng_uv17Pro.py
kk7ds: What's the actual width of the field of potential characters? Is it 16 or more? If it's 16, then use a `u8` in the memory format, then we can collect the bytes and decode them like this I think:
```
bank_raw = bytes([int(x) for x in _mem.banknames[0].name])
bank_decoded = charbytes.decode('gb2312')
```
I think that should be fully clean and the reverse will work for encoding.

chirp/drivers/baofeng_uv17Pro.py
vdwel: I am going to have a look at it.

chirp/drivers/baofeng_uv17Pro.py
vdwel: That's basically what I am doing in _filtername.

chirp/drivers/baofeng_uv17Pro.py
vdwel: The bank names are just a list of 10 bank names. Hardly uses by the radio..... Not sure if it's worth the effort.

kk7ds: Are the banks not editable? Meaning can you assign memories to banks?

chirp/drivers/baofeng_uv17Pro.py
kk7ds: I thought you said it doesn't decode properly when there are multi-byte characters? You're not passing the full list of bytes to the decode when you get to a `0xFF`, which may or may not be right. I'm not sure what the raw encoding of `gb2312` looks like. Either way, treating those as a `char` and doing the `chr(ord(c))` stuff is mixing the interpretations of the contents back and forth between raw bytes and unicode strings in python. Treating them as ints, then raw bytes, and then performing the decode to unicode as a whole is the proper way.

chirp/drivers/baofeng_uv17Pro.py
vdwel: No.. you cannot assign them. The radio contains 1000 channels. The first 100 channels are in the first bank. The second 100 channels are in the second bank etc.

PR #819, opened 2023-11-23
Added driver for Baofeng UV-17 fixes #10865

vdwel:
This driver adds support for the Baofeng UV-17 radio with the blue background screen. It fixes issue #10865

# CHIRP PR Checklist

* All tests must be passing.


vdwel:
I somehow can't manage to reduce the overlapping code with the UV-17 Pro driver. Any suggestions?

kk7ds:
Yeah, there's still a *lot* that is shared between them, so there's plenty of room to re-use code I think. Some of your closures like `apply_code()` can be brought out to module-level and used by both. Looks like your `sync_in()`, `sync_out()` and `get_features()` are basically identical between the two. There's no need to re-define those if they're the same as the parent of course. Also looks like a lot of `get_memory()` could be the same. You can refactor the pro driver to break out things like the `mem.extra` stuff into a separate helper and then just call that from both drivers to avoid replicating that.

chirp/drivers/baofeng_uv17.py
```diff
@@ -0,0 +1,747 @@
...
+    VALID_BANDS = [baofeng_uv17Pro.UV17Pro._vhf_range,
+                   baofeng_uv17Pro.UV17Pro._uhf_range]
+    SCODE_LIST = LIST_SCODE
+
+    MEM_FORMAT = """
+    #seekto 0x0030;
+    struct {
+      lbcd rxfreq[4];
+      lbcd txfreq[4];
+      u8 unused1:8;
```
kk7ds, line 239: I didn't notice that you did this in the previous driver, but `u8 foo:8;` is not only redundant, but it makes the bitwise structure more complex (and thus slower) than it needs to be. You should only use bitfields when you need to access individual bits of a larger type. A byte where all eight bits are treated as one, should just be `u8 foo;`. We can queue that cleanup for the other driver, but it'd be good if we can fix that before we merge this one.

vdwel:
Don't merge this one yet. It seems that some people are not able to connect to their radio, using this driver.. I am trying to figure out if the problem is the driver or that there is another UV-17 radio out there which I haven't heard of yet.

chirp/drivers/baofeng_uv17.py
vdwel: Good one!! I am going to fix that.

vdwel:
People were trying to use the Pro driver for this non-pro driver. :-) How do we fix the fact that it is all so confusing??

So don't put it on hold for merging..

I will try to fix the code reuse issue first... But I'm afraid that's going to be a challenge.. And I don't have infinite time..

vdwel:
I have modified this driver to use a lot of code from the UV-17Pro driver. In order to do so, I also had to modify the UV-17Pro driver. The new version of the Pro driver needs to be merged before this one.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -460,64 +356,186 @@ def process_mmap(self):
...
         self._memobj = bitwise.parse(self.MEM_FORMAT, self._mmap)
 
-    def get_settings(self):
-        """Translate the bit in the mem_struct into settings in the UI"""
-        _mem = self._memobj
-        basic = RadioSettingGroup("basic", "Basic Settings")
-        bank = RadioSettingGroup("bank", "Bank names")
-        dtmfe = RadioSettingGroup("dtmfe", "DTMF Encode Settings")
-        top = RadioSettings(basic, bank, dtmfe)
+    def _download(radio):
```
kk7ds, line 359: This should be `self` now that it's an instance method. It still works because the name is not actually special, but it's very unconventional to have this be anything else and will confuse people into thinking these are a `@staticmethod`.

chirp/drivers/baofeng_uv17.py
```diff
@@ -0,0 +1,423 @@
...
+    def _make_frame(self, cmd, addr, length, data=""):
+        """Pack the info in the header format"""
+        frame = struct.pack("<cH", cmd, addr) + struct.pack(">H", length)
+        # add the data if set
+        if len(data) != 0:
+            frame += data
+        # return the data
+        return frame
+
+    def _download(radio):
```
kk7ds, line 204: You moved all of this into the radio class, but then override it in the subclass anyway. Are the download routines re-usable or not? If not, there's no reason to move them into the class, and it's generally preferable to keep class sizes down, when possible. Things that need not be overridden can be module-scoped to keep the classes small.

Was there a reason for this or was it just to avoid the reuse heuristic? If the latter, I'm sorry you did that, but it wasn't necessary. That check is trying to prevent copying code (or copying with just a few changes), but we need not jump through hoops that don't otherwise make sense to make it pass.

chirp/drivers/baofeng_uv17.py
```diff
@@ -0,0 +1,423 @@
...
+        if mem.number >= 762:
+            offset += 1
+        _mem = self._memobj.memory[mem.number + offset]
+        _nam = self._memobj.names[mem.number]
+
+        if mem.empty:
+            _mem.set_raw("\xff" * 16)
+            return
+
+        _mem.set_raw("\x00" * 16)
```
kk7ds, line 418: This should be `b"\x00"`

chirp/drivers/baofeng_uv17.py
```diff
@@ -0,0 +1,423 @@
...
+            offset += 1
+        if mem.number >= 507:
+            offset += 1
+        if mem.number >= 762:
+            offset += 1
+        _mem = self._memobj.memory[mem.number + offset]
+        _nam = self._memobj.names[mem.number]
+
+        if mem.empty:
+            _mem.set_raw("\xff" * 16)
```
kk7ds, line 415: This should be `b"\xFF"`

chirp/drivers/baofeng_uv17.py
```diff
@@ -0,0 +1,423 @@
...
+                               int(xtone[3])))
+        else:
+            memtone.set_value(0)
+
+        if mode == "DTCS" and pol == "R":
+            memtone.set_value(memtone + 0x4000)
+
+    def set_memory(self, mem):
+        offset = 0
+        # skip 16 bytes at memory block boundary
```
kk7ds, line 404: Does this comment refer to the conditional offsets below?

You can use `#skip` in your memory format to move forward a specific amount, if it helps. When I've written drivers that don't store memories all linearly, I generally try to define the format declaratively instead of doing stuff like this. Something like:
```
struct mem lowmems[252];
#skip 0x0x10;
struct mem highmems[506];
// ... etc
```

Your comment in `get_memory()` mentions that there's a 16-byte "page number". Not really sure what that means but you can also do something like this:
```
struct {
  struct mem memories[252];
  u8 page_stuff[16];
} mem_page[3];
```

The reason to define it like this instead of just papering over it the way you are doing here, is that these "page number" structures will be represented in the browser as memories with non-sensical data in them, which is not great.

At the very least, you should probably make this logic a helper so you don't have to repeat it in `get_memory()` and `set_memory()`, especially if you ever have to tweak it.

Oh and this makes me realize - you didn't implement `get_raw_memory()` here or for the pro one. You should do that, because it makes it *much* easier to iterate and fix things.

chirp/drivers/baofeng_uv17.py
```diff
@@ -0,0 +1,423 @@
...
+        mem = chirp_common.Memory()
+        mem.number = number
+
+        self.get_memory_common(_mem, _nam.name, mem)
+
+        return mem
+
+    def encode_tone(self, memtone, mode, tone, pol):
+        if mode == "Tone":
+            xtone = str(int(tone * 10)).rjust(4, '0')
```
kk7ds, line 384: The more conventional way to do this (which is a numeric operation instead of a string one, and which avoids the pre-conversion) is:
```
xtone = '%04i' % (tone * 10)
```

kk7ds (commented):
Thanks, it's a holiday weekend here so I've been a bit tied up with other stuff.

vdwel:
Thanks for looking at the code!!

chirp/drivers/baofeng_uv17.py
vdwel: I did struggle a lot with the reuse heuristic. :-) This was to prevent the syn_in method twice. I managed to move it outside of the class now.

chirp/drivers/baofeng_uv17.py
vdwel: Fixed. Added get_raw_memory and removed the offset stuff..

chirp/drivers/baofeng_uv17.py
vdwel: Thnx. :-) Fixed.

kk7ds:
Ah, looks like you squashed the refactor and new driver commits together. I adjust brought the other commit in there. I know it's probably maddening to understand my preferences here but: commit history should be clean, methodical changes that build on each other and don't do and then undo things in one epoch of development. So, refactoring one driver in a commit so the next commit can make a change on its own is totally cool, and means that reviewing one commit alone is still easy.

So, if it's cool with you I'll split this back in two (commits) before I merge, since it's trivial to do so and then I promise to take my pill.

Also, I didn't notice before but you need to be adding these drivers to the test matrix as "python3 tested" which is why the existing two drivers show up on the front page with an asterisk. I'll do that for you while I'm splitting this.

PR #823, opened 2023-11-27
update chirp.desktop change MimeType from directory to app-specific, …

wyae:
…fix bug #2955

Original is "MimeType=inode/directory"

This indicates to the system that CHIRP is supposed to be used for opening a directory (e.g. when plugging in a USB stick).

That is NOT what a user wants as default action: opening Chirp instead of a file explorer.

So I changed the MimeType to an (invented) unobtrusive one.


kk7ds:
IDK, it's been that way since 2008 :)

Probably some copy/pasta, I've never known it to be a problem, but  makes sense to fix it of course. I'll merge this when there's some more meat to generate a build. Thanks!

PR #824, opened 2023-11-27
Add TYT TH8600 driver.  Issue 4851.

aaknitt:
[CHIRP issue](https://chirp.danplanet.com/issues/4851)

# CHIRP PR Checklist

* All tests must be passing.


kk7ds:
The support matrix needs re-generating by running `tox -e makesupported` and then commit the result. Looks like maybe you just hand-edited that file, but it's a generated one (from `tests/py3_driver_testers.txt' which is what you should actually modify).

The PR checks are failing because your file is too similar to `th_uv88.py`. Did you copy that file and make changes? If so, please just subclass that driver and change the necessary behaviors. If the parent driver needs some refactoring to make it so you can reimplement certain things, that's fine. We have way too many drivers in chirp that are created by copying a file, making a ton of tiny changes, and then putting it in tree. That means that bugs get copied around a log and fixing them (like the massive effort for python3 compatibility) is made much more difficult.

aaknitt:
> The support matrix needs re-generating by running `tox -e makesupported` and then commit the result. Looks like maybe you just hand-edited that file, but it's a generated one (from `tests/py3_driver_testers.txt' which is what you should actually modify).

Ok, I think I have that fixed now.  The tests don't seem to be running in Github now, not sure why that is. 

> The PR checks are failing because your file is too similar to `th_uv88.py`. Did you copy that file and make changes? 

There were elements from several different TYT radio drivers.  The 8600 seemed to have similarities with several models but there were none that I found were a match in either memory map or supported functions.  I may not have looked hard enough.

> If so, please just subclass that driver and change the necessary behaviors. If the parent driver needs some refactoring to make it so you can reimplement certain things, that's fine. 

I'm not going to have time to refactor this any time in the foreseeable future.  I had some free time to work on this driver over the summer, but that's no longer the case.  

> We have way too many drivers in chirp that are created by copying a file, making a ton of tiny changes, and then putting it in tree. That means that bugs get copied around a log and fixing them (like the massive effort for python3 compatibility) is made much more difficult.

It would be helpful to have this type of background info & preferences on [this page](https://chirp.danplanet.com/projects/chirp/wiki/DevelopersAdd_a_Radio) for new contributors to avoid rework.

kk7ds:
> It would be helpful to have this type of background info & preferences on [this page](https://chirp.danplanet.com/projects/chirp/wiki/DevelopersAdd_a_Radio) for new contributors to avoid rework.

Okay, seems like common good development practice to me, but I'll make it crystal clear on that page.

kk7ds:
I just rebased this and fixed some things that have shifted in the rest of the code base. I still have concerns over how much of this is identical to `th_uv88`, especially since there are lots of little changes which aren't significant which means it's actually more common than it appears. I'm not sure what to do about that at this point, but at the very least this is rebased and brought current.

PR #826, opened 2023-11-27
repeaterbook: Support digital-as-FM conversion

kk7ds:
This makes repeaterbook allow the query to convert multi-mode repeaters
that support FM to be represented as such in the result set, thus
making it easier to add those repeaters to an analog-only radio.

Fixes #10962


kk7ds:
@MELERIX This will add some strings. Can I merge this first and have you update your #825 for this as well?

PR #830, opened 2023-11-29
TIDRADIO H3

songjunbiao:
The new model TIDRADIO H3 is similar to TIDRADIO H8, but the address is different in the protocol.


songjunbiao:
I'm not sure what to make of these tests not passing, can you advise?

kk7ds:
The PR checks failure is telling you (among other things) that your driver is basically just a copy of the H8 (as you noted). You need to subclass the H8 and make only the changes required for the H3 instead of just copying the file. Doing the latter just duplicates code (and bugs) and means we have to maintain two almost identical drivers. The H8 already has some subclasses for Ham, GMRS, etc. You may need to refactor parts of the H8 driver to allow the H3 to change those behaviors, which is fine.

songjunbiao:
Since H8 is not maintained by me, reconstructing the H8 driver is more complicated and the protocol addresses are quite different. Is it feasible to directly submit the H3 driver?

kk7ds:
Not with so much duplication, no. I took a look at the diff between the two and they are indeed almost exactly the same. We have lots of drivers that share most of the implementation but with very different memory addresses, so this is the same situation.

songjunbiao:
We have many drivers that share most of the implementation but use very different memory addresses. Can you tell us which models you have? Do I have permission to modify the code of the H8 driver?

songjunbiao:
Create support matrix keeps failing. What do I need to do to succeed

songjunbiao:
Snipaste_2023-11-30_15-12-27
I keep getting this message. What do I need to do to solve this problem?

kk7ds:
Yes, you can modify the H8 driver (I said that earlier). The makesupported target both generates the matrix file and checks that it is not modified since the last commit. Since you need to modify it, run makesupported, commit the result, and then run it again and it should pass. Also note that currently you still need to add the new driver to `tests/py3_driver_testers.txt` so it will show up as tested.

Here's an excellent recent example from @vdwel doing a refactor of an existing driver before adding another one that subclassed it: https://github.com/kk7ds/chirp/commit/df11f1502808b9e673704bcb1e6c9b0e82d4e389

songjunbiao:
I have merged it. Can you help me find out what else needs to be modified?

chirpwx.py
```diff
@@ -1,5 +1,4 @@
-#!/usr/bin/env python3
-
+#!/usr/bin/env python3;9
```
kk7ds, line 1: I don't think you meant this change, but if you did, please do not :)

tests/py3_driver_testers.txt
```diff
@@ -387,3 +387,7 @@ Yaesu_VX-8DR,@kk7ds,15-Feb-2019
 Yaesu_VX-8GE,+Yaesu_VX-8DR,15-Feb-2019
 Yedro_YC-M04VUS,+Retevis_RT95,13-Nov-2022
 Zastone_ZT-X6,+Retevis_RT22,9-Dec-2022
+TIDRADIO_TD-H3,@songjunbiao,1-December-2023
+TIDRADIO_TD-H3-GMRS,@songjunbiao,1-December-2023
+TIDRADIO_TD-H3-HAM,@songjunbiao,1-December-2023
+TID_TD-H3,@songjunbiao,1-December-2023
```
kk7ds, line 393: These should go in alphabetical order, just like everything else in this file.

chirp/drivers/tdh8.py
```diff
@@ -710,7 +1006,10 @@ def get_features(self):
         return rf
 
     def process_mmap(self):
-        self._memobj = bitwise.parse(MEM_FORMAT, self._mmap)
+        if self.MODEL in H8_LIST:
+             self._memobj = bitwise.parse(MEM_FORMAT, self._mmap)
+        else:
+            self._memobj = bitwise.parse(MEM_FORMAT_H3, self._mmap)
```
kk7ds, line 1012: This would be better as a subclass override, much like `_get_nam()` and some others.

kk7ds (commented):
Thanks, this is much closer. The commits need to be squashed into small logical commits (likely just one).

Would you mind if I make some of the edits I describe here, and push a squashed version to your branch? I would want you to test it with your radio one more time before we merge, but I'm happy to do some of the work to get it into the form I prefer.

Thanks!

songjunbiao:
Excuse me. Is code formatting necessary? Is there any way to automatically format the code? Are these certain before merging the code?

kk7ds:
Yes, the code formatting is necessary.

kk7ds:
No, you don't (and should not) change the H8 image, only add the H3 one like you're doing. The tests are failing because you have a bug in your driver which is affecting the H3 and H8, which you need to fix.

> I keep getting this message. What do I need to do to solve this problem?

As I said earlier, run the target and commit the result. The test is failing because it is checking that there are no changes to the matrix when your commit is applied. Since you're not committing the matrix, it's telling you that you failed.
```
tox -emakesupported
git commit tests/Python3_Driver_Testing.md
```

songjunbiao:
Code format: Is there any tool you recommend? Missing whitespace around operator is often implemented and it is difficult for me to troubleshoot.

kk7ds:
No tool, just fix the problems. It's telling you where all of them are exactly and there aren't that many. It will take less time than asking about and/or installing a tool :)

songjunbiao:
F:\tid\python\chirp-tid>tox -emakesupported
.pkg: _optional_hooks> python d:\python38\lib\site-packages\pyproject_api\_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist> python d:\python38\lib\site-packages\pyproject_api\_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_wheel> python d:\python38\lib\site-packages\pyproject_api\_backend.py True setuptools.build_meta __legacy__
.pkg: prepare_metadata_for_build_wheel> python d:\python38\lib\site-packages\pyproject_api\_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist> python d:\python38\lib\site-packages\pyproject_api\_backend.py True setuptools.build_meta __legacy__
makesupported: install_package> python -I -m pip install --force-reinstall --no-deps F:\tid\python\chirp-tid\.tox\.tmp\package\4\chirp-0.tar.gz
makesupported: commands[0]> python chirp/share/make_supported.py model_support.html
  File "chirp/share/make_supported.py", line 181, in <module>
    output(supported_row(_radio))
  File "chirp/share/make_supported.py", line 108, in output
    dest.write(string + '\n')
UnicodeEncodeError: 'gbk' codec can't encode character '\u20ac' in position 1618: illegal multibyte sequence
makesupported: exit 1 (0.52 seconds) F:\tid\python\chirp-tid> python chirp/share/make_supported.py model_support.html pid=2508
.pkg: _exit> python d:\python38\lib\site-packages\pyproject_api\_backend.py True setuptools.build_meta __legacy__
This happened after I ran it. Did I do something wrong?

songjunbiao:
tox -emakesupported
git commit tests/Python3_Driver_Testing.md
I have successfully followed your instructions and submitted it, but the last test still failed. Can you tell me why? Will the last test fail and affect the merged code? Thank you for your reply.

chirpwx.py
kk7ds: Again, you need to remove this change from your PR.

tox.ini
```diff
@@ -67,6 +67,6 @@ allowlist_externals =
 deps =
     -rtest-requirements.txt
 commands =
-    python chirp/share/make_supported.py model_support.html
+    #python chirp/share/make_supported.py model_support.html
```
kk7ds, line 70: You may not make this change, of course. Please remove.

chirp/drivers/tdh8.py
```diff
@@ -2062,3 +2477,53 @@ def validate_memory(self, mem):
...
+            msgs.append(chirp_common.ValidationError(
+                "The frequency in channels 31-54 must be between"
+                "462.55000-462.72500 in 0.025 increments."))
+        return msgs
+
+
+@directory.register
+class TIDH3(TDH3):
+    VENDOR = "TID"
+    MODEL = "TD-H3"
```
kk7ds, line 2529: This adds a clone of the radio under a slightly-different vendor. Please do not do that and remove this last subclass.

songjunbiao:
Thank you very much for your support. I have done all. When I can see the merged code in the chirp official? Thank you for your reply

songjunbiao:
Required. I have passed them all. What else do I need to do?

tox.ini
```diff
@@ -67,6 +67,6 @@ allowlist_externals =
 deps =
     -rtest-requirements.txt
 commands =
-    python chirp/share/make_supported.py model_support.html
+    # python chirp/share/make_supported.py model_support.html
     python tools/py3_driver_progress.py -o tests/Python3_Driver_Testing.md tests/py3_driver_testers.txt tests/model_stats.tsv
-    git diff --exit-code tests/Python3_Driver_Testing.md
+    # git diff --exit-code tests/Python3_Driver_Testing.md
```
kk7ds, line 72: Obviously you cannot make these changes.

chirp/drivers/tdh8.py
```diff
@@ -638,8 +937,8 @@ class TDH8(chirp_common.CloneModeRadio):
     ident_mode = b'P31183\xff\xff'
     BAUD_RATE = 38400
     NEEDS_COMPAT_SERIAL = False
-    _memsize = 0x1eef
-    _ranges_main = [(0x0000, 0x1eef)]
+    _memsize = 0x1fef
+    _ranges_main = [(0x0000, 0x1fef)]
```
kk7ds, line 941: This changes the H8's memsize, which you should  not do. This probably breaks the H8 - why are you making this change?

tests/py3_driver_testers.txt
```diff
@@ -318,6 +318,9 @@ TIDRADIO_TD-H8-GMRS,@Sandmann34,7-Feb-2023
 TIDRADIO_TD-H8-HAM,@Sandmann34,7-Feb-2023
 TID_TD-M8,+Retevis_RT22,9-Dec-2022
 TID_TD-UV68,@Sandmann34,7-Feb-2023
+TIDRADIO_TD-H3,@songjunbiao,1-December-2023
+TIDRADIO_TD-H3-GMRS,@songjunbiao,1-December-2023
+TIDRADIO_TD-H3-HAM,@songjunbiao,1-December-2023
```
kk7ds, line 323: Please use the same date format as the other several hundred lines in this file (1-Dec-2023)

kk7ds:
In order to merge into CHIRP, you need to:

1. Remove the changes have made in `tox.ini` (which makes the `makesupported` test not run, does not count as a pass)
2. Remove the change you made to `chirpwx.py`
3. Fix your date formatting in the driver testers file
4. Remove the code that breaks the H8 memsize
5. Either squash your changes into one commit or tell me you want me to do that for you

chirp/drivers/tdh8.py
songjunbiao: It will not damage H8. I have updated the test file and it was successful. This is the new function added later.

kk7ds: It changes the communication with the radio. You cannot know that it works by running tests on an image. Also, you may not modify the image for the H8, only the code, and only in ways that do not alter its behavior. This changes its behavior and thus is not allowed. Set `_ranges_main` on your H3 driver if you want it to be different there.

songjunbiao:
All questions answered and fixed
I need you to help me with this thanks

songjunbiao:
I have a question for you.
You are very professional.
 When we have new radios to adapt, can we send them to you for adaptation?

chirp/drivers/tdh8.py
songjunbiao: Thanks for the heads up, we're the vendor of H8 and H3, and since we're adding new features, the address will change, and we'll keep testing it, and if there's a problem, we'll fix it. Do you have any good suggestions?

kk7ds:
A new build is automatically generated the day after something is merged. So there will be a new build with this in a few hours from now.

PR #831, opened 2023-11-29
Fix decoding module during SHA calculation

kk7ds:
When we load a module, we read it in to calculate the SHA for logging.
The current code reads it as text and decodes it, which defaults to
cp1252 on windows instead of UTF-8 as it should be (and as other
platforms do). However, we have to convert that back to bytes to do
the hash, so we should just read it as bytes in the first place to
avoid the issue altogether.


kk7ds:
@KC9HI Can you see if this fixes it for you?

KC9HI:
@kk7ds Yes. This fixed it for me.

PR #832, opened 2023-11-30
Add Wouxun KG-950P support

mrtjr1:
Add support for Wouxun KG-950P 
Fixes #10498

# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/kguv980p.py
```diff
@@ -646,6 +647,19 @@ def _oem_str_encode_1000GPLUS(in_str):
     return out_str
 
 
+def _str_decode_950(in_str):
+    out_str = ''
+    for c in in_str:
+        if c > 95:
+            out_str += ''
```
kk7ds, line 654: This is not free but has no effect. Just replace this line with `pass`

chirp/drivers/kguv980p.py
```diff
@@ -646,6 +647,19 @@ def _oem_str_encode_1000GPLUS(in_str):
...
+        elif c >= 80:
+            out_str += chr(c-48)
+        else:
+            out_str += chr(c+48)
```
kk7ds, line 658: Can you explain what you're doing here? It kinda looks like maybe you're just needing to mask out `0x80` from each of these characters, but I'm not sure what you're trying to accomplish here. Is the character set in the radio just starting at zero (for the character `0`)? Maybe `0x80` is just a flag for "capitals" or something? Either way, doing this with magic decimal numbers doesn't make this very understandable by other people with no comments.

chirp/drivers/kguv980p.py
```diff
@@ -673,6 +687,22 @@ def _str_encode(in_str):
...
+    for c in in_str:
+        if ord(c) < 32 or ord(c) > 126:
+            out_str += ''
+        elif ord(c) >= 32 and ord(c) <= 47:
+            out_str += chr(int(ord(c) + 48))
+        else:
+            out_str += chr(int(ord(c) - 48))
+    while len(out_str) < 8:
+        out_str += chr(0x50)
+    if out_str == "        " or out_str == "":
```
kk7ds, line 701: And this can be:
```
if not out_str.strip():
```
The next person that comes along will have to count how many spaces in that string trigger this special behavior, but the above makes it clear that it's "if it's all spaces, use this special thing".

chirp/drivers/kguv980p.py
```diff
@@ -1249,7 +1287,8 @@ def _get_settings(self):
         vfo_grp.append(vfob_grp)
         vfoa_grp.append(vfo150_grp)
         vfoa_grp.append(vfo450_grp)
-        if self.MODEL == "KG-UV980P":
+        if (self.MODEL == "KG-UV980P" or
+           self.MODEL == "KG-UV950P"):
```
kk7ds, line 1291: This is the sort of thing where I'm willing to go along with the use of the model thing, because it's in the middle of a larger block of stuff that's not easy to swap out. But for the str encode/decode you've already got those as separate methods, so you should use the class hierarchy to do the thing for you.

chirp/drivers/kguv980p.py
```diff
@@ -2352,7 +2455,10 @@ def _decode(lst):
 
         for i in range(1, 11):
             x = str(i)
-            _str = _decode(eval("_settings.scanname"+x))
+            if self.MODEL != "KG-UV950P":
+                _str = _decode(eval("_settings.scanname"+x))
+            else:
+                _str = _decode_950(eval("_settings.scanname"+x))
```
kk7ds, line 2461: Here again, this should be just `self._decode()` and override it on the subclasses that need it.

kk7ds (changes requested):

PR #839, opened 2023-12-06
Fix Windows icon missing other resolutions

kk7ds:
When I changed the icon recently I didn't generate a .ico file with
the various resolution options, which caused it to look terrible.


kk7ds:
@KC9HI I think this will fix it. I'll leave this until we have some more meat so we're not rolling out a build just for an icon change :)

PR #840, opened 2023-12-07
Added support for Baofeng 5RM, Fixes #10991

vdwel:
- Added the Baofeng 5RM radio to the driver with airband support. Fixes #10991
- Fixed channel numbering, now starting at 1
- Removed bank names setting when bank names are not supported by the radio
- Corrected dates in py3_driver_testers.txt

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/tests/Python3_Driver_Testing.md
+++ b/tests/Python3_Driver_Testing.md
@@ -59,9 +59,9 @@
 | <a name="Baofeng_F-11"></a> Baofeng_F-11 | [Implied by Baofeng_UV-5R](#user-content-Baofeng_UV-5R) | 18-Nov-2022 | Yes | 0.12% |
 | <a name="Baofeng_GT-3WP"></a> Baofeng_GT-3WP | [@KC9HI](https://github.com/KC9HI) | 3-Dec-2022 | Yes | 0.43% |
 | <a name="Baofeng_GT-5R"></a> Baofeng_GT-5R | [@KC9HI](https://github.com/KC9HI) | 18-Nov-2022 | Yes | 0.68% |
-| <a name="Baofeng_UV-17"></a> Baofeng_UV-17 | [@vdwel](https://github.com/vdwel) | 27-Nov-2022 | Yes |  |
-| <a name="Baofeng_UV-17Pro"></a> Baofeng_UV-17Pro | [@vdwel](https://github.com/vdwel) | 27-Nov-2022 | Yes |  |
-| <a name="Baofeng_UV-17ProGPS"></a> Baofeng_UV-17ProGPS | [@vdwel](https://github.com/vdwel) | 27-Nov-2022 | Yes |  |
+| <a name="Baofeng_UV-17"></a> Baofeng_UV-17 | [@vdwel](https://github.com/vdwel) | 27-Nov-2023 | Yes |  |
+| <a name="Baofeng_UV-17Pro"></a> Baofeng_UV-17Pro | [@vdwel](https://github.com/vdwel) | 27-Nov-2023 | Yes |  |
+| <a name="Baofeng_UV-17ProGPS"></a> Baofeng_UV-17ProGPS | [@vdwel](https://github.com/vdwel) | 27-Nov-2023 | Yes |  |
 | <a name="Baofeng_UV-3R"></a> Baofeng_UV-3R | [@yarda](https://github.com/yarda) | 26-Sep-2022 | Yes | 0.45% |
 | <a name="Baofeng_UV-5R"></a> Baofeng_UV-5R | [@lekv](https://github.com/lekv) | 23-Oct-2022 | Yes | **28.81%** |
 | <a name="Baofeng_UV-6"></a> Baofeng_UV-6 | [Implied by Baofeng_UV-5R](#user-content-Baofeng_UV-5R) | 18-Nov-2022 | Yes | 0.46% |
--- a/tests/py3_driver_testers.txt
+++ b/tests/py3_driver_testers.txt
@@ -74,9 +74,9 @@ Baofeng_UV-82HP,@KC9HI,18-Nov-2022
 Baofeng_UV-82WP,@KC9HI,3-Dec-2022
 Baofeng_UV-9G,@KC9HI,3-Dec-2022
 Baofeng_UV-9R,@KC9HI,3-Dec-2022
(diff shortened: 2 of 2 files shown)


chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -865,7 +868,7 @@ def get_features(self):
             "DTCS->DTCS"]
         rf.valid_skips = self.SKIP_VALUES
         rf.valid_dtcs_codes = self.DTCS_CODES
-        rf.memory_bounds = (0, self.CHANNELS - 1)
+        rf.memory_bounds = (1, self.CHANNELS)
```
kk7ds, line 871: This is a change for all the affected radios right? Was this an error in the initial implementation? It's obviously really unideal to ever change this on a whim...

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -1066,3 +1069,11 @@ class UV17ProGPS(UV17Pro):
...
     VALID_BANDS = [UV17Pro._airband, UV17Pro._vhf_range, UV17Pro._vhf2_range,
                    UV17Pro._uhf_range, UV17Pro._uhf2_range]
+
+
+@directory.register
+class BF5RM(UV17Pro):
+    VENDOR = "Baofeng"
+    MODEL = "5RM"
+    VALID_BANDS = [UV17Pro._airband, UV17Pro._vhf_range, UV17Pro._vhf2_range,
+                   UV17Pro._uhf_range, UV17Pro._uhf2_range]
```
kk7ds, line 1079: Is this actually different from the thing it inherits from? The `VALID_BANDS` look identical to L1070 above. If not, it might be better to just add this as a website alias (see all the "use..." notes on the front page)

chirp/drivers/baofeng_uv17Pro.py
vdwel: It affects all radios. People were complaining that the channel numbers in CHIRP started at 0, where in the radio it starts at 1. So now it shows channel 1 as 1 in CHIRP. Only affects the way things are displayed.

chirp/drivers/baofeng_uv17Pro.py
vdwel: Yes, airband is added for this radio.

chirp/drivers/baofeng_uv17Pro.py
kk7ds: Yeah, I know it's just the display, and it should be right, it's just unfortunate to change it after the driver has been merged.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -568,9 +568,12 @@ def get_settings(self):
...
         _mem = self._memobj
         basic = RadioSettingGroup("basic", "Basic Settings")
-        bank = RadioSettingGroup("bank", "Bank names")
         dtmfe = RadioSettingGroup("dtmfe", "DTMF Encode Settings")
-        top = RadioSettings(basic, bank, dtmfe)
+        if self._has_support_for_banknames:
+            bank = RadioSettingGroup("bank", "Bank names")
+            top = RadioSettings(basic, bank, dtmfe)
+        else:
+            top = RadioSettings(basic, dtmfe)
```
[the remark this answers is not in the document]
kk7ds: Yeah, I understand what it's doing and why, it's just not related to the adding of the new radio.

chirp/drivers/baofeng_uv17Pro.py
vdwel: I agree... I didn't realise it was possible. Does it introduce a problem?

vdwel:
> Are there bugs for the other issues other than the adding of the new radio? I'd like to split this into three commits for the three completely unrelated changes being made here. I can do it, but we should tag those issues in those commits.

No there no other issues other than the adding to the new radio. (not as I know of)

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
kk7ds: It's just best if they're separate commits, and separate issues of course. If we needed to revert, say, the new radio model commit, we wouldn't want to regress on the two other bug fixes just because they were all squashed together.

Anyway, I'll just create issues and split the commits.

chirp/drivers/baofeng_uv17Pro.py
vdwel: I agree.. Still learning the CHIRP ways. :-)

chirp/drivers/baofeng_uv17Pro.py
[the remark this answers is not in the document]
vdwel: 👍 I'll remember that for next time.

vdwel:
I think many radios are supported with this driver. Such as the UV-21, UV-18... Since I don't own these radios, it's difficult to verify. How do we add these as an alias?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -1066,3 +1069,11 @@ class UV17ProGPS(UV17Pro):
     _has_skey2_short = True
     VALID_BANDS = [UV17Pro._airband, UV17Pro._vhf_range, UV17Pro._vhf2_range,
                    UV17Pro._uhf_range, UV17Pro._uhf2_range]
+
+
+@directory.register
+class BF5RM(UV17Pro):
+    VENDOR = "Baofeng"
+    MODEL = "5RM"
```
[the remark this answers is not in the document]
vdwel: It's not the radio from your link... This one has a color screen. I know... very confusing Baofeng stuff.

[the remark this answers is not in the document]
vdwel: We are talking about this radio: https://www.miklor.com/COM/Review_5RM.php

kk7ds:
> I think many radios are supported with this driver. Such as the UV-21, UV-18... Since I don't own these radios, it's difficult to verify. How do we add these as an alias?

The support list on the front page is generated from the code, and augmented with just-a-clone aliases using `chirp/share/model_alias_map.yaml`. So if you add things in there, the next build will update the front page to match.

kk7ds:
> Great!! Before we do that, I want to be 100% sure they are the same.

Yep, for sure.

Okay, this should be neatly split now. Look okay?

PR #844, opened 2023-12-10
update ja_JP locale

weboo:
I have updated the Japanese translation and adding some translation targets.
I would be happy to reflect them if there is no problem.

### DIFF
--- a/chirp/drivers/uvk5.py
+++ b/chirp/drivers/uvk5.py
@@ -646,14 +646,14 @@ class UVK5Radio(chirp_common.CloneModeRadio):
 
     def get_prompts(x=None):
         rp = chirp_common.RadioPrompts()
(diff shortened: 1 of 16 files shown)


weboo:
It seems to be stuck in the PEP8 style check, should I fix this?

kk7ds:
I'd like to keep the changes to the K5 driver as a separate commit from updating the jp locale itself.

Mind if I split that stuff up for you and push it here?

weboo:
I split that commit in two. Please check it.

kk7ds:
Yep, perfect, thanks. I just tweaked the commit message on the first one to mention which driver was affected. Will merge this for tomorrow.

weboo:
Thank you very much for your support.

PR #848, opened 2023-12-15
Reflect settings modifications of RadioSettingValue._validate_callbac…

egzumer:
…k onto a GUI

When a validate callback is defined and it modifies a value, or rises an exception it is not reflected onto the user interface. The value in internal settings will be different from what a user is presented with. Moreover if the validation rises an exception the value in the interface is preserved, subsequent changes to any other settings will again read the value from GUI and rise the exception each and every time. This PR fixes it.

### DIFF
--- a/chirp/wxui/settingsedit.py
+++ b/chirp/wxui/settingsedit.py
@@ -125,7 +125,15 @@ class ChirpSettingsEdit(common.ChirpEditor):
                                                             val))
                         realname, index = name.split(common.INDEX_CHAR)
                         if int(index) == j:
-                            setting[j] = val
+                            try:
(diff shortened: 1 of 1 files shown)


kk7ds:
I'm not sure I understand when this can happen exactly. Can you give me a model/image/setting to tweak to reproduce this?

egzumer:
Sorry. I thought this is straight forward.

[fake.zip](https://github.com/kk7ds/chirp/files/13692203/fake.zip)

1. Choose item4, then every time you change `A knob` you get an error
2. Choose item2, GUI shows iteam2, where in reality there is iteam1 in the settings

PR #852, opened 2023-12-18
Fix bug in getting grid property in ChirpSettingsEdit._apply_settings

egzumer:
PR #848 contained a bug. I was using `_propgrid` which is always equal to the last page that was created. Instead active page has to be used.

### DIFF
--- a/chirp/wxui/settingsedit.py
+++ b/chirp/wxui/settingsedit.py
@@ -124,6 +124,11 @@ class ChirpSettingsEdit(common.ChirpEditor):
                                                             j,
                                                             val))
                         realname, index = name.split(common.INDEX_CHAR)
+                        prop = page.propgrid.GetProperty(name)
+                        if not prop:
+                            LOG.warning('Unable to find property %s in page',
+                                        name)
+                            continue
                         if int(index) == j:
                             try:
                                 setting[j] = val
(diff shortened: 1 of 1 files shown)


chirp/wxui/settingsedit.py
```diff
@@ -131,9 +131,11 @@ def _apply_settings(self):
                                 if setting[j].get_value() != val:
                                     # setting value modified or not accepted in
                                     # validate callback, propagate back to GUI
-                                    prop = self._propgrid.propgrid.GetProperty(
-                                        name)
-                                    prop.SetValue(setting[j].get_value())
+                                    prop = self._group_control \
+                                        .GetCurrentPage() \
```
kk7ds, line 135: This should just be `page` above right? We're applying all the settings at once, which means current page is probably not right either.

chirp/wxui/settingsedit.py
```diff
@@ -131,9 +131,11 @@ def _apply_settings(self):
...
+                                        .propgrid.GetProperty(name)
+                                    if prop:
+                                        prop.SetValue(setting[j].get_value())
```
kk7ds, line 138: This is getting way too nested, and also using backslash continuations is not very conventional in chirp's style. However, I think using `page` per above will help.

kk7ds (commented):
I'm going to make the change described here because we've got multiple bug reports about this. I'll also tag those so they get closed when this merges.

kk7ds:
@kc9hi can you re-test this fix with my changes? It was correct-er, but not fully correct I think. It works for me, but I wasn't able to reproduce the core issue, so I'd appreciate a quick re-validation if you can.

KC9HI:
@kk7ds it is still working for me.

PR #862, opened 2024-01-02
Simplify drivers loading

egzumer:
For your consideration, I was trying to make a binary and had problems with drivers loading. I guess you replace drivers/__init__.py when you do a release because it isn't working when chirp is freezed. This version seems to work for both cases. I took that from [nuitka note](https://nuitka.net/doc/user-manual.html#manual-python-file-loading)

BTW. can you say what tool you're using to do your releases?


kk7ds:
I'll have to test this in several situations (and platforms) before I can consider making the change. I'm no opposed to it in principle, but I need to make sure it doesn't break anything first. Not sure when I'll get to that, but I'll leave this open until then.

PR #869, opened 2024-01-13
Fix/vulture

dforsi:
This PR fixes some issues found with `vulture --min-confidence 100 chirp`; other issues with confidence 100 and below remain.

The first commit modifies what looks like a constant added to debug code.

The second commit removes what looks like a copy and paste error from `huv1f.py` because that's the only driver that handles the `"disnm"` setting.

The third commit does some refactoring so in both cases and object with only the mandatory arguments gets created.

The fourth commit fixes a check for an integer number.

### DIFF
--- a/chirp/drivers/ic9x_ll.py
+++ b/chirp/drivers/ic9x_ll.py
@@ -145,7 +145,7 @@ class IC92Frame:
 
     def send(self, pipe, verbose=False):
         """Send the frame to the radio via @pipe"""
-        if verbose or True:
+        if verbose:
             LOG.debug("Sending:\n%s" % util.hexprint(self.get_raw()))
 
         response = ic9x_send(pipe, self.get_raw())
--- a/chirp/drivers/th_uvf8d.py
+++ b/chirp/drivers/th_uvf8d.py
@@ -621,12 +621,6 @@ class TYTUVF8DRadio(chirp_common.CloneModeRadio):
 
         return top
 
-        group.append(RadioSetting(
-                "disnm", "Display Name",
-                RadioSettingValueBoolean(_settings.disnm)))
-
-        return group
-
     def set_settings(self, settings):
         _settings = self._memobj.settings
 
--- a/chirp/drivers/uv6r.py
+++ b/chirp/drivers/uv6r.py
@@ -853,7 +853,7 @@ class UV6R(baofeng_common.BaofengCommonHT):
         match_model = False
 
         # testing the file data size
-        if len(filedata) == 0x2008 or 0x2010:
+        if len(filedata) in (0x2008, 0x2010):
             match_size = True
 
         # testing the firmware model fingerprint
--- a/chirp/wxui/memedit.py
+++ b/chirp/wxui/memedit.py
@@ -266,9 +266,6 @@ class ChirpFrequencyColumn(ChirpMemoryColumn):
 
 
 class ChirpVariablePowerColumn(ChirpMemoryColumn):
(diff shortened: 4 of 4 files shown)


chirp/drivers/ic9x_ll.py
```diff
@@ -145,7 +145,7 @@ def __str__(self):
 
     def send(self, pipe, verbose=False):
         """Send the frame to the radio via @pipe"""
-        if verbose or True:
+        if verbose:
```
kk7ds, line 148: This is leftover debug. I often do this to force a condition, intending to remove it later :)

chirp/drivers/uv6r.py
```diff
@@ -853,7 +853,7 @@ def match_model(cls, filedata, filename):
         match_model = False
 
         # testing the file data size
-        if len(filedata) == 0x2008 or 0x2010:
+        if len(filedata) in (0x2008, 0x2010):
```
kk7ds, line 856: I obviously prefer the latter, but these sorts of machine-generated changes just create noise in the git history for very little benefit, IMHO.

chirp/wxui/memedit.py
```diff
@@ -872,12 +869,13 @@ def filter_unknowns(items):
...
-        valid_power_levels = self._features.valid_power_levels
         valid_tuning_steps = self._features.valid_tuning_steps
         if self._features.has_variable_power:
-            power_col_cls = ChirpVariablePowerColumn
+            power_column = ChirpVariablePowerColumn('power', self._radio)
         else:
-            power_col_cls = ChirpChoiceColumn
+            valid_power_levels = self._features.valid_power_levels
+            power_column = ChirpChoiceColumn('power', self._radio,
+                                             valid_power_levels)
```
kk7ds, line 878: This too. I might should have done it this way from the beginning, but I don't think these sorts of changes are worth the churn, personally.

chirp/drivers/th_uvf8d.py
```diff
@@ -621,12 +621,6 @@ def _filter(name):
 
         return top
 
-        group.append(RadioSetting(
-                "disnm", "Display Name",
-                RadioSettingValueBoolean(_settings.disnm)))
-
-        return group
-
```
kk7ds, line 629: Looks like this goes with the TODO above for implementing this, so I guess it's okay to just remove.

kk7ds:
There's a legit fix and real cleanup here so I'll merge this, but in general the "the machine likes this style better" is not something I'd like to spend more time and churn on.

chirp/drivers/uv6r.py
dforsi: The tool doesn't edit the code, I did it by hand.

If the purpose of that line was to accept as valid the length of 0x2008 and the length of 0x2010, then that code was broken because it accepted as valid also a length of zero:
```
>>> 0x2008 == 0x2008 or 0x2010
True
>>> 0x2010 == 0x2008 or 0x2010
8208
>>> 0x0000 == 0x2008 or 0x2010
8208
```

kk7ds: > The tool doesn't edit the code, I did it by hand.

I know, I meant machine-generated complaints/suggestions.

> If the purpose of that line was to accept as valid the length of 0x2008 and the length of 0x2010, then that code was broken because it accepted as valid also a length of zero:

Ah, you're right. This looks very similar to some changes that "code hygiene" tools suggest that try to clean up "condition or condition or condition" to "value in [valid1, valid2, ...]". Apologies for jumping to that conclusion :)

--Dan

PR #870, opened 2024-01-16
Added ability to open/import .csv files with # comments in them.

NGD-Skier:
(i.e. lines that start with "#" are full line comments)

Feature#:  11081
The comments are stripped out and written into a <filename.csv>.chirptmp.csv file before calls to portions of code that won't allow comments (e.g. the library used to process .csv files doesn't support comments, so it is passed the temp file that had comments stripped, so it doesn't fail).  After that is done, the .chirptmp.csv file is deleted, and the original file is never touched in the process.  This has been tested and works fine for csv file opens and exports.  With regards to imports, I don't see any behavior that would be unexpected by the user.  With regards to opens, all works fine, but if the user makes a change to the csv file they opened in chirp and saves it to the original name, chirp will write it in the standard format it always has, which does NOT include any comments, so they'll lose any comments in the original file because this new features only enables chirp to not error out on files with comments, but it does not preserve any full line comments if they overwrite the file.  I suppose a warning dialog could be added when saving a csv file, stating something like "If this file had any comments in it when opened, saving this file will remove those comments from this file.  If you want to save these changes and leave the original source file with comments intact, it is suggested that you cancel and use File, Save As to save this file with a different name.

# CHIRP PR Checklist

* All tests must be passing.


NGD-Skier:
P.S.  I wasn't sure exactly what was meant by these two lines:
-Commits should be squashed into logical units.
-Commits should be rebased (or simply rebase-able in the web UI) on current master. Do not put merge commits in a PR.
, so let me know if I need to change something.  I started with the master branch, made a branch with my changes in it (changes to just one file), and pushed my new branch to my fork, and then did a pull request.  If I need to do the changes in master directly or other squashing/rebasing, please give me a hint at what needs to change.  Thanks.

chirp/directory.py
```diff
@@ -120,6 +120,16 @@ def get_driver(rclass):
...
+    # Creating temp file for csv comment support
+    if image_file.endswith(".csv"):
+        with open(image_file, 'r') as origfile, \
+                open(image_file + '.chirptmp.csv', 'w') as tempfile:
+            for line in origfile:
+                # exclude lines starting with comment character "#"
+                if not line.startswith('#'):
+                    tempfile.write(line)
+        image_file = image_file + '.chirptmp.csv'
+
```
kk7ds, line 132: This is really not a good way to do this. For one, it doesn't catch files that end in `.CSV` but it also may try to convert files that claim to be CSV in name but don't otherwise contain that kind of content. We really just need to filter the content as it is read. I can help with that if necessary, and I'd definitely appreciate the functionality of being able to tolerate comments. We just need to remember that they'll be stripped out with an open-save cycle, unless we do something more complicated.

kk7ds (changes requested):

chirp/directory.py
NGD-Skier: Agreed, not a great way to do it, and I was hoping to do it while reading the file (i.e.  I was hoping for a single read loop or something, where I could add a line that filters out comments), but it looks like .csv files are read two different ways in Chirp, once with a binary read into filedata in directory.py and then using csv.reader in generic_csv.py, so I used the temp file option to just do it in one place in the code and make it work.  After thinking on your comments more, I dug a little deeper and now have it filtering out the lines in two different places, but without having to use a temp file.

As for catching files that end in .csv but which aren't .csv files, I'm not sure what to do about that.  If someone calls a file a .csv file, but it isn't, then it seems like they're trying to do something wrong and might get whatever results, they get, no?  And isn't that the same behavior they'd get right now?   Do you currently have code that verifies that a file opened with a .csv extension actually is a .csv file before doing anything with it?  If so, let me know where, and maybe I could reuse that code.

Anyway, I rewrote the code to not use temp files and will update my branch in my fork with the new code (I think that's what I should do.  If it looks like that didn't work, I'll close this request and open a new one with my updated branch...  Let me know what you think.  :-)

Also, for future reference, if there is an easier way to "update a pull request" with new code, let me know.

PR #874, opened 2024-01-16
Fix/translation

dforsi:
This PR adds the handling of plurals, makes some more strings translatable and brings the Italian translation to 60%

### DIFF
--- a/chirp/locale/de.po
+++ b/chirp/locale/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-12-13 06:20+0900\n"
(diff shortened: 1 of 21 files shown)


dforsi:
The check_commit step fails with this error
```
+        catalog = {'value': _(self.THING),
Translated strings must be literals!
```
however that code that creates an error message when latitude and other values are out of bounds works for me and I see a popup like this:

image

kk7ds:
Yeah, the marking function needs to be used on literals. In non-python it's a hard requirement. In python it works on non-literals in certain cases but can fail in strange ways. We likely just need to tweak the code a bit to be able to mark the literal properly. I will have a look in a bit.

kk7ds:
This is how this needs to go (I know, it's annoying). It works for me in my native language (:P) and I think it should be good so I'll merge this. If something still doesn't look right, let me know.

Note I changed the commit message of the "fully translatable" commit since it was a bit ambiguous for the mailing list summary.

Thanks!

PR #881, opened 2024-01-16
Remove the "executable" permission from memory image files

dforsi:
This PR just removes the executable permission from some data files.


kk7ds:
I think this comes from windows users. The execute bit is interpreted as something else on some of the cygwin tooling. Might be mostly from the mercurial days, I'm not sure.

PR #882, opened 2024-01-16
Implement 'Open stock config directory' for Linux

dforsi:
This PR enable the menu item 'Open stock config directory' for Linux. Tested on a Debian system.


chirp/wxui/main.py
```diff
@@ -603,7 +603,7 @@ def add_stock(fn):
 
         stock.Append(wx.MenuItem(stock, wx.ID_SEPARATOR))
 
-        if sys.platform in ('darwin', 'win32'):
+        if sys.platform in ('darwin', 'linux', 'win32'):
```
kk7ds, line 606: There aren't any other platforms excluded now, so this could just go away now.

PR #883, opened 2024-01-18
Updated Spanish Translation

MELERIX:


kk7ds:
I had to rebase this, and I also squashed the two commits into one, I hope that's okay. Since the first line of each commit message goes to the users list "Update Spanish translation" is meaningful to  them ... most won't know what `es.po` is :)

PR #886, opened 2024-01-20
Add a GUI option to set restore tabs on start

dforsi:
This PR adds under the "View" menu an option 'Restore tabs on start' to toggle the restore_tabs preference that was already honored by CHIRP if present in chirp.config, but couldn't be set or unset with the GUI.

When set, it has the same effect as if the command line option `--restore` was always specified (the welcome page and the menu item "File / Open Recent / Restore %i tabs" aren't visibile).

When not set, behavior is like before this PR.

### DIFF
--- a/chirp/locale/de.po
+++ b/chirp/locale/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-01-16 21:38+0100\n"
+"POT-Creation-Date: 2024-01-20 12:27+0100\n"
 "PO-Revision-Date: 2012-10-02 22:11+0100\n"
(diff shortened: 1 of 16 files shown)


kk7ds:
I'm just curious, but do you actually want to use this or were you just closing the loop on the feature? I added this for myself a while back for iterative development and ended up finding it a lot less useful than I expected.

dforsi:
> I'm just curious, but do you actually want to use this or were you just closing the loop on the feature?

I want to use it because I have only one radio+cable supported by
CHIRP, so why should I manually click File, then Open or Open recent
and manually load the same file each time?
My main use case would be copying repeater settings so I will probably
also keep another tab open to do copy and paste.

-- 
Daniele Forsi

kk7ds:
Okay, cool, just checking. I was kinda excited about it when I added it and then after using it a bit I decided it wasn't as convenient as I thought. Maybe I should try again :)

PR #887, opened 2024-01-20
Autosize height of column headers of memedit

dforsi:
This PR fixes a small rendering glitch for column headers that span 2 lines, where letters on the second line are incomplete, see "Tone Mode", "Tone Squelch, "DTCS Polarity", "PTT ID" in the image below:

Before: (click to enlarge)
image

After: (click to enlarge)
image


kk7ds:
I don't see this on macos or Windows, but I guess this manifests on Linux?

PR #889, opened 2024-01-21
Fix check for lines added to cpep8.manifest and cpep8.blacklist and remove clean files from cpep8.blacklist

dforsi:
This PR changes the lines in `tools/check_commit.sh` that check for lines added to `tools/cpep8.manifest` and `tools/cpep8.blacklist` (and adds the hint to fix the code in cpep8.blacklist), then removes clean files from `tools/cpep8.blacklist`. 

The check for added lines couldn't work because it was using the file `added_lines` that contained the lines from `*.py` files and since `git diff` adds an header containing line starting with `+++`, adding the files `cpep8.blacklist` and `cpep8.manifest`  would always give a false positive. The way I found to ignore the lines with `+++` is piping to `tail` because a command using `git diff -G` searches the actual lines so it can't find the `+` prefix and so does `git grep`; another way could be using something like `git diff --shortstat` and grepping for the word `insertion` but it will break if the output gets translated in a future version of git.

### DIFF
--- a/tools/check_commit.sh
+++ b/tools/check_commit.sh
@@ -46,12 +46,12 @@ if grep -E "[^_]_\([^\"']" added_lines; then
     fail 'Translated strings must be literals!'
 fi
 
-if grep '/cpep8.manifest' added_lines; then
+if git diff ${BASE}.. 'tools/cpep8.manifest' | tail -n +5 | grep -q '^+'; then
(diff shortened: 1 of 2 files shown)


kk7ds:
> The check for added lines couldn't work because it was using the file `added_lines` that contained the lines from `*.py` files 

Yeah, I changed that recently to only show added lines from py files and didn't catch this, thanks.

PR #890, opened 2024-01-22
Fix marking tab as modified on menu delete

kk7ds:
This unifies our Edit->Delete and context->Delete paths to use the
same handler, which also fixes the issue where Edit->Delete does not
mark the tab as modified.

Fixes: #11090


kk7ds:
@dforsi FYI. I'll hold this until there's more meat for a build, but this should do it. Thanks for noticing/reporting :)

PR #895, opened 2024-01-28
17Pro (5RM): Refactor settings and add work mode menu

hgrasman:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/baofeng_uv17Pro.py
+++ b/chirp/drivers/baofeng_uv17Pro.py
@@ -21,7 +21,7 @@ from chirp import chirp_common, directory, memmap
 from chirp import bitwise
 from chirp.settings import RadioSetting, \
     RadioSettingValueBoolean, RadioSettingValueList, \
-    RadioSettingValueString, \
+    InvalidValueError, RadioSettingValueString, \
     RadioSettings, RadioSettingGroup
 import struct
 from chirp import errors, util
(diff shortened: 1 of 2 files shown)


hgrasman:
Feature 11109:
-Refactors 'get_settings(self)' menu top level construction as list->tuple pointer(?) so that configurations no longer need to be hard coded.  
-Cleans sublevel menu special cases into their respective 'get_settings_common_x' functions. I can undo this if people want, but I think this makes it a lot more clear what's going on here. No effect on driver behavior.

Feature 11108:
-Adds 5RM Work Mode Settings. This is using common memory mapping, so other radios may be able to enable this without issue. I cannot test, so leaving the flag only enabled for 5RM.
-Updates class definition DTMF Signal code list to be the right length (15) for 5RM

I'm sure there's loads of style war crimes.

hgrasman:
Stuck on this "savemode" key error. I didn't touch that, so I assume messing with the menu code is my hubris. Help wanted

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -742,6 +777,163 @@ def apply_Key2short(setting, obj):
                           RadioSettingValueBoolean(_mem.settings.fmenable))
         basic.append(rs)
 
+    def get_settings_common_workmode(self, workmode, _mem):
+
+        vfoA = RadioSettingGroup("vfoA", "VFO A")
+        vfoB = RadioSettingGroup("vfoB", "VFO B")
+
+        def convert_bytes_to_freq(bytes):
```
kk7ds, line 785: This gets copied as a closure into each driver that needs this functionality, so we have this duplicated in a number of places. We really need to put this in a single place (maybe `baofeng_common.py`?) instead of doing this. Fancy doing that refactor for karma points? :)

kk7ds:
Ah, I see you marked it ready, so I assume so.

chirp/drivers/baofeng_uv17Pro.py
hgrasman: Happy to! I'll take a look on a slow day if nobody beats me to it

PR #899, opened 2024-02-03
Leixen VV-898S/E dual bank radios added as JT270MH aliases

bgrahambo:
Fixing long running issue with newer Leixen VV-898S/E dual bank radios. Support was already added for identical radio JT270MH, so simply adding VV898SDualBank and VV898EDualBank as aliases to that one. Also bumping up rf.valid_bands up to 480000000 as repeatedly requested in this issue. I've tested this by uploading from OEM vendor software to radio, then downloading and uploadnig with CHIRP, and everything seems to check out. 

Related issue:
[https://chirp.danplanet.com/issues/4069](https://chirp.danplanet.com/issues/4069)


chirp/drivers/leixen.py
```diff
@@ -961,14 +961,25 @@ class JetstreamJT270MRadio(LeixenVV898Radio):
     _file_ident = b"JET"
     _model_ident = b'LX-\x89\x85\x53'
 
+class VV898SDualBank(chirp_common.Alias):
```
kk7ds, line 964: I'm actually trying to eliminate this alias behavior since we started stashing metadata in the file format, so they should just be subclasses.

chirp/drivers/leixen.py
```diff
@@ -961,14 +961,25 @@ class JetstreamJT270MRadio(LeixenVV898Radio):
...
+
+    '''Newer VV898S 1.06+ firmware that features dual memory banks'''
+    VENDOR = "Leixen"
+    MODEL = "VV-898S (1.06+ firmware)"
```
kk7ds, line 968: We really try not to put stuff like this in `MODEL`. If we *must* do this, then it should be in `VARIANT`.

Is there not any way we can detect which model is the right one when talking to the radio? Perhaps reading a chunk of memory to extract a version string or some indication that it's the newer model? It's really awful user experience for people to have to pick "the right one" from the model box and we really try to avoid that whenever possible.

chirp/drivers/leixen.py
```diff
@@ -1034,3 +1045,4 @@ class LeixenVV898SRadio(LeixenVV898Radio):
     _power_levels = [chirp_common.PowerLevel("Low", watts=5),
                      chirp_common.PowerLevel("Med", watts=10),
                      chirp_common.PowerLevel("High", watts=25)]
+
```
kk7ds, line 1048: Please don't add random whitespace damage :)

kk7ds (commented):
Please put "Fixes #4069" in your commit message so the issue tracker matches this up against it. It needs to be in the commit message, not in the github meta conversation.

Also, if we can't detect which version firmware the radio has, could we consider just an annotation on the radio matrix on the site? See the "Use model X" things for example.

kk7ds:
Looks like we can read the chunk at `0x168` before we start the download and compare it to the newer thing to see if it's the new radio or not. I have a proposal in another PR to allow one radio class to say "actually, the radio is this other one" during clone to avoid the user having to select the right one. Seems like this might be another good use for that.

bgrahambo:
> I have a proposal in another PR to allow one radio class to say "actually, the radio is this other one" during clone
That sounds like a great idea, but is a lot more significant than the couple aliases/subclasses in this PR. Do you want this re-submitted with the style/class changes fixed per your comments, or hold off to do something different?

kk7ds:
It is more work, but once we introduce a new vendor/model/variant we are pretty much stuck with it forever, lest we break people's images if we ever remove it. The detection thing still requires the subclasses, but we need to get them right for what we want to stick with long term.

How about you update this with subclasses and moving to `VARIANT` and then give me a bit to merge the infrastructure for the detection thing and then I'll update this with that approach and get some feedback from you and others? I have a JT270MH I can test with (somewhere).

I know the ticket is old and it sucks to wait longer, but at least there's a workaround for people at the moment...

Also, if you want to do the range thing in a separate PR (ideally should be a different commit anyway) I can get that merged sooner than this.

kk7ds:
I'll comment in the bug about this shortly, but this seems to work for me. If I select JT270MH it goes straight in, if I choose 898S, it does a detection, chooses the dual-bank model, and runs.

PR #903, opened 2024-02-05
Settings hardening

kk7ds:
- uv5r: Attempt to harden settings loading
- Refactor setting initialization for serviceability
- Leave invalid settings uninitialized


kk7ds:
@KC9HI Hopefully this will address the *loading* of images with invalid settings in most of the cases. I've tested with an image with invalid vhf/uhf upper/lower ranges, but not one with other corrupt settings. I'll shoot you an email tomorrow with some details about what this changes and how we can test some things. But, this lets me load an image with invalid limits, change settings, and even correct the limits by giving them sane values.

PR #904, opened 2024-02-05
th_uv88 : small fixes

velter:
Fix a communication error when using an RDP forwarded serial port under windows
Fix Number of channel (200 not 199)
Fix CTSS and DTCS code List


kk7ds:
Blessed by @KC9HI , I rebased this and will merge shortly.

PR #909, opened 2024-02-07
introlines settings adjustments to better match with the official programming software

velter:
- introlines are 16 char long but the 16th char is not displayed correctly so force the 16th to space
- a truncated introline1 to 12 char is stored in basicsettings

### DIFF
--- a/chirp/drivers/th_uv88.py
+++ b/chirp/drivers/th_uv88.py
@@ -126,11 +126,9 @@ struct {
 
 #seekto 0x1940;
 struct {
-  char name1[15];         // Intro Screen Line 1 (16 alpha text characters)
-  u8 unk1;
-  char name2[15];         // Intro Screen Line 2 (16 alpha text characters)
-  u8 unk2;
-} openradioname;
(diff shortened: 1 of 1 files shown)


kk7ds:
This is now in conflict after your other fix. I can probably resolve the conflicts for you if you need, but it'd be better if you can to confirm the changes.

Also, @KC9HI any complaints about these changes? Not sure I fully grok it but I imagine it's more obvious if you have/know the radio.

chirp/drivers/th_uv88.py
```diff
@@ -990,29 +989,47 @@ def get_settings(self):
...
+            setattr(obj1, atrb1, str(setting.value)[:12])
+            setattr(obj2, atrb2, setting.value)
+            return
+
+        def _name_validate(value):
+            # The 16th char is not displayed correctly.
+            # Force it to space
+            name = list(value)
+            name[15] = " "
+            return "".join(name)
```
kk7ds, line 1005: Maybe a cleaner way to do this is:
```
name = value[:15].ljust(16)
```

PR #911, opened 2024-02-08
introlines settings adjustments

velter:
to better match with the official programming software.
- introlines are 16 char long but the 16th char is not displayed correctly so force the 16th to space
- a truncated introline1 to 12 char is stored in basicsettings


kk7ds:
Thanks, I'll hold off merging this for a few days until we get a few more things to make up a build.

PR #914, opened 2024-02-09
KG-UV980P freq limit fix

mrtjr1:
Fixes: #11096
Update config map to send 3 blocks of 16 bytes to allow radio to accept the last sets of freq limits.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/kguv980p.py
+++ b/chirp/drivers/kguv980p.py
@@ -68,6 +68,7 @@ config_map = (          # map address, write size, write count
 
 
 MEM_VALID = 0x00
+MEM_INVALID = [0xFF, 0x80]
 TX_BLANK = 0x40
 RX_BLANK = 0x80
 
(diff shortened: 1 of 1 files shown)


mrtjr1:
Dan,  if you have a preferred way to replace the set_raw function I would
be happy to make the changes for the drivers of mine that use it.

Mel

kk7ds:
Replace it? It's okay to use it when it makes sense (like for clearing a whole structure like you do). However, it should be using bytes instead of strings (so `set_raw(b'\xFF'), 16)` for example).

mrtjr1:
Ok..I saw the depreciation warning and thought you wanted replaced.
I will work on fixing them to be byte compliant.

kk7ds:
I think you're seeing the "don't pass strings to this" warning? Yeah, that's all that needs changing, thanks :)

PR #916, opened 2024-02-10
th_uv88 : optional signaling implementation #11110

velter:
Add settings to configure :
- dtmf
- 2 tones
- 5 tones

updated test image file to have a wider variety of settings


chirp/drivers/th_uv88.py
```diff
@@ -1094,6 +1261,496 @@ def get_settings(self):
...
+        options = ['1', '2', '3', '4', '5', '6', '7', '8',
+                   '9', '10', '11', '12', '13', '14', '15', '16']
+        rx = RadioSettingValueList(options, options[_settings.dtmfch])
+        rset = RadioSetting("basicsettings.dtmfch", "Select channel", rx)
+        dtmfenc.append(rset)
+
+        # Encode channels
+        for i in range(16):  # 0 - 15
+            sigchan = RadioSettingSubGroup("dtmfencchan%d" % i,
+                                           "Channel %02d" % (i+1))
```
kk7ds, line 1493: Nice use of SubGroup. This is new, I just added it recently, but I think it really helps organize things better. Note that the "FM radio" section of this radio could probably use some similar cleanup, if you feel like it :)

I wish the style checker would complain, but python conventional style is space around operators (so `i + 1`), with `=` being the exception when used in parameter lists.

chirp/drivers/th_uv88.py
```diff
@@ -1094,6 +1261,496 @@ def get_settings(self):
...
+            options = []
+            for i in range(count):
+                options.append(format % (start+i*step))
+            rx = RadioSettingValueList(options, options[value])
+            rset = RadioSetting(name, display, rx)
+            setting.append(rset)
+
+        def _set_int10(setting, obj, atrb):
+            vx = float(str(setting.value))
+            vx = int(vx * 10)
```
kk7ds, line 1357: I don't think you should need these `int` and `float` casts here. If you do, maybe there's some work that needs doing elsewhere.

chirp/drivers/th_uv88.py
```diff
@@ -1094,6 +1261,496 @@ def get_settings(self):
...
+
+            rx = RadioSettingValueList(options, options[value])
+            rset = RadioSetting(name, display, rx)
+            setting.append(rset)
+
+        def _set_mask_array(setting, src, idx):
+            if bool(setting.value):     # Enabled = 1
+                vx = 1
+            else:
+                vx = 0
```
kk7ds, line 1385: I think you should be able to just `int(setting.value)` or just use `setting.value` without the cast, no? Not sure if you did this out of habit or need, but let me know and I could maybe make some more native fixes so they behave properly.

kk7ds (commented):
Not much to complain about, other than asking for space around operators and a few suggestions. Thanks for doing this work!

chirp/drivers/th_uv88.py
velter: I found the SubGroup while looking at recent commits :) Yes FM settings would also benefit the use of subgroup. I will add this to the #11153 issue

just commited the style fix

chirp/drivers/th_uv88.py
velter: You are right. Removing the cast do the job.

chirp/drivers/th_uv88.py
kk7ds: Sure, but int(bool) in python will give you a 0 or 1 and save you three lines. Anyway, I know it's copy/paste, just thought I'd point it out since you asked about pythonic code style :)

kk7ds:
This is still marked as draft, so I've been leaving it alone. No pressure, but if/when you're ready for it to merge just un-draft it .

velter:
I just undrafted it as it works well in the current state. I only own a TH-UV88 and got no feedback for RA89/QRZ-1. While those radios ae very close I was not able to do tests.
It seems that there is a conflict due to PR #906 merge. Im' not sure how to resolve these with git.

PR #920, opened 2024-02-11
Make the MessageBoxes stay on top of their parent Dialog

dforsi:
This PR avoids an issue (at least on Linux Gtk) when:

- a dialog is open (ChirpMemPropDialog or ChirpCloneDialog)
- an error message or an information message is displayed in a MessageBox (such as for an invalid entry or a communication error)
- the user clicks on the main window or on the dialog causing the MessageBox to get hidden behind the dialog:

image


kk7ds:
Thanks, I used to be better at this when I was a linux user and we were on GTK. On macOS, the dialogs are handled specially and are sort of modal for the whole application scope, so I don't notice the omissions as much.

PR #921, opened 2024-02-11
Square PNG icon

nils-werner:
On most Linux DEs, icons are usually expected to be square, and on GNOME a non-square application icon leads to a squished and ugly icon in the application launcher. This PR tries to mitigate that by replacing the non-square icon with a padded, square one. I didn't re-export the SVG but simply resized the existing PNG.


kk7ds:
Thanks, but you can't replace the existing one, which is used for other things. We can add another one and make the `.desktop` file use it if you want.

Here's one that I generated but never included the last time I did some maintenance. Let me know if this works, and if so, we should probably use it because it came from the SVG source directly.

chirpsquare

I'm surprised that @dforsi hasn't complained about this, so maybe we can get another data point about if we should always use the square version in the `.desktop` file?

dforsi:
I'm using KDE and the current rectangular icon looks good even when hovering with the mouse or with the keyboard:

image

kk7ds:
Can you try with the square one to see if it also works? It seems pretty dumb for gnome not to be able to do the right thing with a rectangular image, but maybe KDE will be happy with either?

dforsi:
The square icon works with KDE too.
I used the one by @nils-werner because I wasn't able to download yours from this page.

Can we use an icon 512x512 pixels instead of 531x531?

kk7ds:
Mine is 256x256. Does it need to be the higher res, or are you happy with a power-of-two 256? I can regenerate it, but it's more work since I already have the 256.

nils-werner:
Yeah, it seems pretty dumb to stretch it... but I think icons have always been square and a power of two, and on every platform. So it also makes sense to follow that convention/spec...

kk7ds:
So shall I push over this with my 256px one in a different place and change the `.desktop` to match?

kk7ds:
Actually, here, this makes the `chirp.png` one the square one and adds `chirp-logo.png` as the trimmed one. I'll change some of the other stuff that looks for the trimmed one. I need to do a couple test builds to make sure that's all right, but if either/both of you can test from source here to make sure it looks right, that'd be helpful.

dforsi:
Looks ok to me.
This square icon would probably look ok in the software center of KDE (the Discover application) which is stretching the rectangular icon.

PR #922, opened 2024-02-11
Fixes issue #9793

vdwel:
uv17: improved handling of memory corruption
uv13pro: added uv13pro

# CHIRP PR Checklist

* All tests must be passing.


chirp/drivers/baofeng_uv17.py
```diff
@@ -65,27 +64,33 @@ def _download(radio):
...
-        for addr in range(start_addr, start_addr + 0x1000,
-                          radio.BLOCK_SIZE):
-            frame = radio._make_read_frame(addr, radio.BLOCK_SIZE)
-            # DEBUG
-            LOG.debug("Frame=" + util.hexprint(frame))
+        if block_number not in memory_map:
+            """Memory block not found."""
+            raise errors.RadioError('Radio memory is corrupted. \
+                                     Fix this by uploading a backup image \
+                                     to the radio.')
```
kk7ds, line 71: This won't look like you want, because it includes all the indenting space in front of the other lines. Best to do this:
```
raise errors.RadioError('Radio memory is corrupted. '
                        'Fix this by uploading a backup image '
                        'to the radio.')
```

chirp/drivers/baofeng_uv17.py
```diff
@@ -65,27 +64,33 @@ def _download(radio):
...
     for block_number in radio.BLOCK_ORDER:
-        block_index = memory_map.index(block_number) + 1
-        start_addr = block_index * 0x1000
-        for addr in range(start_addr, start_addr + 0x1000,
-                          radio.BLOCK_SIZE):
-            frame = radio._make_read_frame(addr, radio.BLOCK_SIZE)
-            # DEBUG
-            LOG.debug("Frame=" + util.hexprint(frame))
+        if block_number not in memory_map:
+            """Memory block not found."""
```
kk7ds, line 68: Probably better to make this a comment. As a string it gets processed and stored in memory for no reason.

chirp/drivers/baofeng_uv17.py
```diff
@@ -65,27 +64,33 @@ def _download(radio):
...
+            raise errors.RadioError('Radio memory is corrupted. \
+                                     Fix this by uploading a backup image \
+                                     to the radio.')
+        else:
```
kk7ds, line 72: This doesn't need to be here because we'll abort with the raise, so you can keep all the stuff below one indent level less.

chirp/drivers/baofeng_uv17.py
```diff
@@ -105,10 +110,13 @@ def _upload(radio):
     radio.status_fn(status)
 
     for block_number in radio.BLOCK_ORDER:
-        # Choose a block number is memory map is corrupt
+        # Choose a block number if memory map is corrupt
         # This happens when te upload process is interrupted
```
kk7ds, line 114: Maybe fix the "te" typo while you're here? :)

chirp/drivers/baofeng_uv17.py
```diff
@@ -43,7 +43,6 @@ def _get_memory_map(radio):
         frame = radio._make_frame(b"R", addr, 1)
         baofeng_common._rawsend(radio, frame)
         blocknr = ord(baofeng_common._rawrecv(radio, 6)[5:])
```
kk7ds, line 45: I saw this while I was looking at this driver for that bug. If `_rawrecv()` returns a `bytes` then you just need this:
```
baofeng_common._rawrecv(radio, 6)[5]
```
Which will be an int. Example:
```
>>> b = b'\x01\x02\x03'
>>> b[1]
2
```
Maybe clean that up while you're here?

chirp/drivers/baofeng_uv17.py
```diff
@@ -149,7 +157,7 @@ class UV17(baofeng_uv17Pro.UV17Pro):
     upload_function = _upload
 
     MODES = ["FM", "NFM"]
-    BLOCK_ORDER = [16, 17, 18, 19,  24, 25, 26, 4, 6]
+    BLOCK_ORDER = [0x16, 0x17, 0x18, 0x19, 0x24, 0x25, 0x26, 0x4, 0x6]
```
kk7ds, line 160: Erm, these are not equivalent. I'm guessing they should have been hex to begin with and your removal of L46 makes these proper? Even still L46 wasn't converting these from decimal to hex equivalents or anything... Just making sure you're intending the change here.

kk7ds (commented):
Thanks for doing this, I'm sure the uv13 people will appreciate it :)

When you revise, can you make the first line of your commit message something suitable for the mailing list? Perhaps just "uv17: improved handling of memory corruption" would be best. Adding the uv13 in a separate commit will also make that visible to people. I can split those two things for you before merge too, if it would help.

chirp/drivers/baofeng_uv17.py
vdwel: It's cleaner this way. And makes the handling of corrupted memory blocks easier.

kk7ds: I'm definitely not arguing that it's not cleaner, I just want to make sure you're accounting for it and I'm not sure I see where.

17 is not equal to 0x17, but it looks like you were multiplying this by 0x1000 above, but that hasn't changed to match this.

chirp/drivers/baofeng_uv17.py
vdwel: Apple keyboard. :-S Sometimes skips what I type.

chirp/drivers/baofeng_uv17.py
vdwel: True... What was I doing there.. ? :-)

chirp/drivers/baofeng_uv17.py
vdwel: True.. It was a leftover from when I tried to make the drive so that it would fix the image. But I decided that the error makes more sense.

chirp/drivers/baofeng_uv17.py
vdwel: Done

chirp/drivers/baofeng_uv17.py
vdwel: I didn't notice anything strange with the message box. But your solution is how it's supposed to be. thnx

vdwel:
> Thanks for doing this, I'm sure the uv13 people will appreciate it :)
> 
> When you revise, can you make the first line of your commit message something suitable for the mailing list? Perhaps just "uv17: improved handling of memory corruption" would be best. Adding the uv13 in a separate commit will also make that visible to people. I can split those two things for you before merge too, if it would help.

I would appreciate it if you can split it for me.

kk7ds:
> I would appreciate it if you can split it for me.

Ack, will-do.

vdwel:
Thanks for the review!! I learned some things again. :-)

chirp/drivers/baofeng_uv17.py
```diff
@@ -65,6 +64,11 @@ def _download(radio):
     radio.status_fn(status)
 
     for block_number in radio.BLOCK_ORDER:
+        if block_number not in memory_map:
```
kk7ds, line 67: Ah, this is the answer to my question about changing the dec to hex. You're just using those as indexes into the array, so it doesn't matter what they are. I thought you were multiplying those values, like on L73 below. But, I see you're just *indexing* and using that, so makes sense.

vdwel: They are in BCD format. I was converting them, but it's pointless since I am only using the values for indexing, indeed.

kk7ds:
Split as requested. I'll merge this when it's time for the next build, thanks!

PR #927, opened 2024-02-12
KG-UV980P Fix missing channels in older images

mrtjr1:
Fixes: #11159
Not all channels showing in older Chirp images after 20240211 update 
Also fix set_raw deprecation warning for non-byte use

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/kguv980p.py
+++ b/chirp/drivers/kguv980p.py
@@ -68,7 +68,7 @@ config_map = (          # map address, write size, write count
 
 
 MEM_VALID = 0x00
-MEM_INVALID = [0xFF, 0x80]
+MEM_INVALID = [0x80]
 TX_BLANK = 0x40
 RX_BLANK = 0x80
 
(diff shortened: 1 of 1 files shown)


chirp/drivers/kguv980p.py
```diff
@@ -1011,21 +1011,27 @@ def get_memory(self, number):
...
-                                    _mem.rxfreq == 0x00000000)):
+           ((_valid != MEM_VALID) & ((_mem.rxfreq == 0xFFFFFFFF) or
+                                     _mem.rxfreq == 0x00000000))):
             mem.empty = True
             if _valid == 0x80:
                 LOG.debug("Ch %s was deleted by using radio menu" % number)
                 LOG.debug("Current Memory: \n%s" % _mem)
                 LOG.debug("Clearing Ch %s memory" % number)
-                _mem.set_raw("\xFF" * (_mem.size() // 8))
-                _nam.set_raw("\xFF" * (_nam.size() // 8))
```
kk7ds, line 1023: I just pushed this up, but for future reference, this is now available:

https://github.com/kk7ds/chirp/commit/e48bb5f1c662131e0ac5b6a5bf5670e4f9901e6e

Which could make these things:
```
_mem.fill_raw(b'\xFF')
_nam.fill_raw(b'\xFF')
```

Just FYI. I should have written that a long time ago :)

mrtjr1: Ok.  I will start to transition to this if it is what you prefer.  

Is there any reason to use one vs the other or should all set_raw move to the fill_rawm

kk7ds: It's just much less typing and much cleaner looking when you're filling with the same byte. No need to convert existing stuff if you don't want to, but going forward new code that fills with a single byte like this should ideally use this helper.

PR #928, opened 2024-02-12
Fix/ft1d

dforsi:
This PR:

- reorders some structs in the memory map
- removes a duplicated assignment that cause a warning to be  printed when installing the Debian package of chirp version 1:20240122-2
- fixes a small typo

A `#seekto` directive was useful because an "invalid negative seek" warning pointed a previously unknown field that was instead partially known and so could be deleted to fix the offset.


kk7ds:
> * removes a duplicated assignment that cause a warning to be  printed when installing the Debian package of chirp version 1:20240122-2

Probably running python 3.12 I'm guessing?

dforsi:
Running python 3.11

PR #931, opened 2024-02-14
python3-future is not needed anymore

dforsi:
This PR removes python3-future.

I'm using Debian testing and python3-future was removed recently, I uninstalled it today so I can't reinstall it easily, ouch!

Debian applied a patch to remove python3-future from its chirp package version 1:20240122-2 but I don't think that their chunk about urlilib is needed for the removal:
https://salsa.debian.org/debian-hamradio-team/chirp/-/commit/454c4d7f991d02061395fa2962957864705c9446


kk7ds:
Thanks, I need to go remove some `six` stuff as well...

PR #935, opened 2024-02-15
Feature/tests

dforsi:
This PR improves test coverage.

### DIFF
--- a/chirp/bitwise.py
+++ b/chirp/bitwise.py
@@ -67,7 +67,6 @@ import six
 from builtins import bytes
 
 from chirp import bitwise_grammar
-from chirp.memmap import MemoryMap
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


kk7ds:
This is still marked as draft so I'm waiting to merge, just FYI.

dforsi:
I would like to improve test coverage just deleting some more lines, for example do all the drivers in the following list really need to be executed directly? They aren't documented in the top-level README's.

```
grep -rl "if __name__" --include=*.py chirp
chirp/wxui/__main__.py
chirp/wxui/fips.py
chirp/platform.py
chirp/drivers/idrp.py
chirp/drivers/icx90.py
chirp/drivers/ic9x.py
chirp/drivers/tmv71_ll.py
chirp/drivers/id31.py
chirp/drivers/hf90.py
```
I would remove at least the code that tries to connect to a serial port with a constant Linux name, or change 

I would also like to fix some test that are failing but I don't know how.
For example there are 35 tests that apparently for the same reason; like this:
> FAILED tests/test_banks.py::TestCaseBanks::test_bank_index - TypeError: 'NoneType' object is not callable
fails at line 31 of base.py:
> self.parent = self.RADIO_CLASS(self.TEST_IMAGE)

kk7ds:
> I would like to improve test coverage just deleting some more lines, for example do all the drivers in the following list really need to be executed directly? They aren't documented in the top-level README's.

They're usually used for developers to do/test things directly, sometimes to grab just a firmware version, etc. I used to do more of that myself, but haven't in years with the presence of better tools. I'm not opposed to removing them necessarily, especially in cases where the code may not even work on python3. However, in general, I don't really think it hurts much to leave it, even if it's just useful as an example.

> chirp/wxui/__main__.py

This needs to stay.

> chirp/wxui/fips.py

This was for generating/updating the maps. That data doesn't change much (or ever), but that was the goal.

> chirp/platform.py

Again, for local/developer testing. It doesn't use the newer stuff that gets proper windows versions anyway, so it can go.

> chirp/drivers/idrp.py

Can be removed.

> chirp/drivers/icx90.py

This looks like it was useful to the author fairly recently, and is pretty well fleshed out. Not really sure if it's worth removing or not.

> chirp/drivers/ic9x.py
> chirp/drivers/tmv71_ll.py
> chirp/drivers/id31.py

Remove.

> chirp/drivers/hf90.py

Also recent, for probing firmware versions I think.

> I would also like to fix some test that are failing but I don't know how. For example there are 35 tests that apparently for the same reason; like this:
> 
> > FAILED tests/test_banks.py::TestCaseBanks::test_bank_index - TypeError: 'NoneType' object is not callable
> > fails at line 31 of base.py:
> > self.parent = self.RADIO_CLASS(self.TEST_IMAGE)

I'm not sure where/how you're seeing these fail. Can you give me an invocation to run?
```
$  tox -edriver -- -k test_bank_index
. . .
====================================== 13 passed, 287 skipped, 3764 warnings in 11.74s 
```

dforsi:
I was executing the wrong command:
> pytest tests/test_banks.py -k test_bank_index

your command succeeds
> tox -edriver -- -k test_bank_index

I was fooled by other tests that succeed when called directly with pytes, I still need to learn a lot about pytest.

PR #942, opened 2024-02-16
uvk5 egzumer: fix broken "Upload calibration" option

egzumer:


kk7ds:
Apologies if I broke that, but a variable on the object instance is the right way to do that instead of the global, so.. thanks :)

PR #949, opened 2024-02-20
Fixes #10600: Add support for the 2023 version of the Radioddity GA510

bsilvereagle:
# CHIRP PR Checklist

* All tests must be passing.


kk7ds:
Hi, I haven't looked at this in detail yet, but I just want to give you an early warning that it's not okay to just copy the other driver to a new file and make some tweaks for the new version. Unless it's a totally different radio, we'll need to work to create a subclass of the original one and just implement the differences.

kk7ds:
Looks like this is a subclass of the U17, so if the new GA510 is really just a rebadged UV-17 that should be easy. It looks like there's a bunch here that is copied from the UV17, but that you can just omit because it'll be inherited from the parent.

PR #951, opened 2024-02-21
TIDRADIO: Enforce usedflags at read time to prevent untouched memorie…

hgrasman:
…s from appearing full

Fixes #11185

Absolutely untested. This pc is still on suicide watch. trying to see if it passes tests at least.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -1081,7 +1081,10 @@ class TDH8(chirp_common.CloneModeRadio):
 
         if _mem.get_raw(asbytes=False)[0] == "\xff":
             mem.empty = True
+            self.set_usedFlag(mem)
             return mem
+
+        self.set_usedFlag(mem)
 
         # receiving frequency
         mem.freq = int(_mem.rxfreq) * 10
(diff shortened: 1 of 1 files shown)


hgrasman:
Working between bluescreens: tested very quick. Passes tests, fixes problem.

Merge it

cheers,
Hank

hgrasman:
Question for the future:

Is set_memory() called for every memory location edited by a copy/paste from another radio? If so, I can't think of why this would be an issue in the first place. Either way this will restore an image to proper function after the fact.

chirp/drivers/tdh8.py
```diff
@@ -1081,8 +1081,11 @@ def get_memory(self, number):
 
         if _mem.get_raw(asbytes=False)[0] == "\xff":
             mem.empty = True
+            self.set_usedFlag(mem)
             return mem
 
+        self.set_usedFlag(mem)
+
```
kk7ds, line 1088: These change the image just because we fetched a memory out of it, which is not cool. That alters the state even though the UI says it's not modified and in general is not a good thing to do in something that should otherwise be a read-only operation. You wouldn't want a word processor modifying your document just because you opened it and didn't change anything...

kk7ds (changes requested):
I'm not sure why you want to do this, but I think it's indicative of a different problem.

kk7ds:
> Is set_memory() called for every memory location edited by a copy/paste from another radio? If so, I can't think of why this would be an issue in the first place. Either way this will restore an image to proper function after the fact.

Yes, the only way a memory can be changed in a driver is via `set_memory()`. So, edits, pastes, even deletes on drivers that don't implement that separately.

hgrasman:
I agree modifying the memory on read is not good practice. I'm not sure what I was thinking. 

I'll look into the root cause again.

chirp/drivers/tdh8.py
hgrasman: In this case, those bits being set is a form of corruption, as it is an invalid configuration. 

I suppose an argument could be made that this change is to catch and restore, but that's not our business.

kk7ds: You can (should) be tolerant of corruption in `get_memory()`. You should not modify/fix memory until asked to update it I'd say. We have lots of cases where radios come from the factory with out-of-range values in fields where we need to be tolerant of them and take a default when and where we can. If there's no sane default, then raise an exception if we just need to signal it.

PR #953, opened 2024-02-21
always upload

kk7ds:
- Explain un-uploadable tabs to the user
- Make sure to always select a port in clone
- Log memory repr instead of str on validation fail


kk7ds:
PR check is a false positive

PR #954, opened 2024-02-21
uvk5: Fix bounds-checking for egzumer variant

kk7ds:
Fixes #11134

### DIFF
--- a/chirp/drivers/uvk5.py
+++ b/chirp/drivers/uvk5.py
@@ -763,6 +763,9 @@ class UVK5RadioBase(chirp_common.CloneModeRadio):
     def get_raw_memory(self, number):
         return repr(self._memobj.channel[number-1])
 
+    def _find_band(self, hz):
+        return _find_band(self._expanded_limits, hz)
+
     def validate_memory(self, mem):
         msgs = super().validate_memory(mem)
 
@@ -778,13 +781,13 @@ class UVK5RadioBase(chirp_common.CloneModeRadio):
             txfreq = mem.freq
 
         # find band
-        band = _find_band(self._expanded_limits, txfreq)
+        band = self._find_band(txfreq)
         if band is False:
             msg = "Transmit frequency %.4f MHz is not supported by this radio"\
                    % (txfreq/1000000.0)
             msgs.append(chirp_common.ValidationError(msg))
 
-        band = _find_band(self._expanded_limits, mem.freq)
+        band = self._find_band(mem.freq)
         if band is False:
             msg = "The frequency %.4f MHz is not supported by this radio" \
                    % (mem.freq/1000000.0)


chirp/drivers/uvk5.py
```diff
@@ -1013,6 +1016,11 @@ def get_memory(self, number2):
         else:
             mem.empty = False
 
+        # FIXME: This should set duplex='off' for channels that are outside
```
kk7ds, line 1019: @egzumer Can you tell me what a reasonable bound is for when we should allow the radio to transmit? I assume it can't actually transmit on 27MHz, right?

egzumer: It depends on radio settings (modulation, unlock settings, etc). If the settings are right you can transmit on any freq you can input.

kk7ds: Ah, okay, well, I'll just leave it as-is then.

egzumer:
Just a side note to make you aware, don't know if you want to deal with it or not, I know it is a mess and all, that is why I write it here.

Most of the UV-K5 firmwares (with one exception) has backwards compatible eeprom structure (for the most part), so the standard UV-K5 driver kind of works for all of them. I assume someone already complained it didn't work with 1o11 firmware and you added it before.
The same thing is with my firmware, there are forks that don't change memory structure at all, but they have different firmware strings, the driver will refuse to communicate with those.

I guess creators of other firmwares could just add their firmware strings to chirp, but a way to manually choosing a variant would also solve this.

kk7ds:
> Just a side note to make you aware, don't know if you want to deal with it or not, I know it is a mess and all, that is why I write it here.

Yeah, it's a mess :) I'm glad to have one of the firmware authors (you) involved at least, so thanks.

> Most of the UV-K5 firmwares (with one exception) has backwards compatible eeprom structure (for the most part), so the standard UV-K5 driver kind of works for all of them. I assume someone already complained it didn't work with 1o11 firmware and you added it before. The same thing is with my firmware, there are forks that don't change memory structure at all, but they have different firmware strings, the driver will refuse to communicate with those.
> 
> I guess creators of other firmwares could just add their firmware strings to chirp, but a way to manually choosing a variant would also solve this.

Well, causing trouble with people's radios is something we really try not to do. We just had a problem recently with people using the UV17 driver with the UV13Pro and causing some trouble because of assumptions the driver makes that are wrong for the UV13Pro. Luckily we were able to reverse the "damage" but we might not always be that lucky. When the original uvk5 driver was written, I asked for firmware version checking, but it never got done. Allowing people to pick any variant they want to "try" has bitten us in the past because, despite the warnings, people are (understandably) very unhappy when it goes badly.

I think I'm most happy to just have the firmware string whitelist in the base driver and add to that when another firmware version is reported to be working. If it becomes a problem of scalability, we can look at ways to allow testing (i.e. bypass the requirement to see if it works) but I don't think we're there yet. I've been jumping on uvk5 requests myself, especially for more "doesn't work for firmware X" things.

Incompatible forks can still use the mechanism to use an alternate subclass for a given variant, even if they have to be much more divergent from the base than your variant. Like I did with yours, I'll be glad to help those to "fit" as well as possible.

PR #955, opened 2024-02-22
Memedit extra fixes

kk7ds:
- Add support for integer mem.extra columns
- Add tooltips to memedit column headers
- uv_5118plus: Set docs on mem.extra fields

### DIFF
--- a/chirp/wxui/memedit.py
+++ b/chirp/wxui/memedit.py
@@ -559,6 +559,8 @@ def get_column_for_extra(radio, setting):
                                  label=label)
     elif isinstance(value, settings.RadioSettingValueBoolean):
         return ChirpFlagColumn(field, radio, label=label)
+    elif isinstance(value, settings.RadioSettingValueInteger):
+        return ChirpMemoryColumn(field, radio, label=label)
 
 
 class ChirpMemoryDropTarget(wx.DropTarget):
--- a/chirp/wxui/memedit.py
+++ b/chirp/wxui/memedit.py
@@ -73,6 +73,26 @@ class ChirpRowLabelRenderer(glr.GridDefaultRowLabelRenderer):
         self.DrawText(grid, dc, rect, text, hAlign, vAlign)
 
 
+DEFAULT_COLUMN_HELP = {
+    'freq': _('Receive frequency'),
+    'name': _('Memory label (stored in radio)'),
+    'tmode': _('Tone squelch mode'),
+    'rtone': _('Transmit tone'),
+    'ctone': _('Transmit/receive tone for TSQL mode, else receive tone'),
+    'dtcs': _('Transmit/receive DTCS code for DTCS mode, else transmit code'),
+    'rx_dtcs': _('Receive DTCS code'),
+    'dtcs_polarity': _('TX-RX DTCS polarity (normal or reversed)'),
+    'cross_mode': _('Complex or non-standard tone squelch mode '
+                    '(starts the tone mode selection wizard)'),
+    'duplex': _('Transmit shift, split mode, or transmit inhibit'),
+    'offset': _('Shift amount (or transmit frequency) controlled by duplex'),
+    'tuning_step': _('Frequency granularity in kHz'),
+    'skip': _('Scan control (skip, include, priority, etc)'),
+    'power': _('Transmit Power'),
+    'comment': _('Human-readable comment (not stored in radio)'),
+}
+
+
 class ChirpMemoryColumn(object):
     DEFAULT: object = ''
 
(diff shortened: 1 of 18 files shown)


kk7ds:
@KC9HI This fixes your bug about integer mem.extra fields. But also:

This adds tooltips for the memedit column headers by default:

image

But it also makes it so that mem.extra setting docs show up there for those fields:

image

Not sure if you noticed but I added similar support for this to the settings editor:

image

I think it'd be nice if some of the more popular radios started setting "docs" on the the settings in mem.extra and in settings. Example here for the Ruyage radio I used to fix the missing integer column.

Thoughts?

chirp/wxui/memedit.py
```diff
@@ -73,6 +73,25 @@ def Draw(self, grid, dc, rect, row):
...
+    'dtcs': _('Transmit/receive DTCS code for DTCS mode, else transmit code'),
+    'rx_dtcs': _('Receive DTCS code'),
+    'dtcs_polarity': _('TX-RX DTCS polarity (normal or reversed)'),
+    'cross_mode': _('Complex or non-standard tone squelch mode'),
+    'duplex': _('Transmit shift, split mode, or transmit inhibit'),
+    'offset': _('Shift amount (or transmit frequency) controlled by duplex'),
+    'tuning_step': _('Frequency granularity in kHz'),
+    'skip': _('Scan control (skip, include, priority, etc)'),
+    'power': _('Transmit Power'),
+    'comment': _('Human-readable comment (not stored in radio)'),
```
kk7ds, line 91: @KC9HI Specifically, any comments or edits to these descriptions to make them more useful?

KC9HI: @kk7ds The last example looks like it is from a Quansheng UV-5K type radio. I like what I see.

I don't know if I noticed this or not. If I did, it didn't sink in as a new feature. I don't spend much time hovering over the column headers. ;)

Will the 'memedit' descriptions only be in English only or will they be override-able by the language files?

I would like to see this done for the more popular radios, too. It looks like the Baofeng UV-5R, Baofeng BF-888 and Baofeng UV-82 radios rank 1, 2 and 3 on the list. Updating the uv5r.py driver would take care of 1 and 3 (and many other variants). The '888' has less features and settings so it should be the quickest (and hopefully easiest) to complete.

The 'memedit' descriptions are pretty good as they are. I don't know if I could come up with anything better, or not. 'freq', 'name', 'power' and 'comment' are about as simple and to the point as they could be. Tones and DTCS are probably the settings that causes the most confusion to new users. So maybe some time should be spend with them. Maybe 'Cross Mode' (the 'M' needs to be uppercase like it is in 'Tone Mode') could mention something like...

'Complex or non-standard tone squelch mode (starts the tone selection wizard)'

kk7ds: > @kk7ds The last example looks like it is from a Quansheng UV-5K type radio. I like what I see.

Yep.

> I don't know if I noticed this or not. If I did, it didn't sink in as a new feature. I don't spend much time hovering over the column headers. ;)

The "doc" for regular settings was a feature from the old UI from ages ago, but not very widely used. I only recently added it to the new UI. The column headers here are new in this PR, as is supporting the doc string for `mem.extra` settings.

> Will the 'memedit' descriptions only be in English only or will they be override-able by the language files?

They're marked (that's what the `_("foo")` means) so they are translate-able yeah.

> I would like to see this done for the more popular radios, too. It looks like the Baofeng UV-5R, Baofeng BF-888 and Baofeng UV-82 radios rank 1, 2 and 3 on the list. Updating the uv5r.py driver would take care of 1 and 3 (and many other variants). The '888' has less features and settings so it should be the quickest (and hopefully easiest) to complete.

Yeah, that'd be great. Just doing `setting.set_doc()` is all that is required to make that work. I've added it to some of my pet drivers. I figure I'll also go through and flesh out some more in the uvk5 since I've been in there a lot recently.

> The 'memedit' descriptions are pretty good as they are. I don't know if I could come up with anything better, or not. 'freq', 'name', 'power' and 'comment' are about as simple and to the point as they could be. Tones and DTCS are probably the settings that causes the most confusion to new users. So maybe some time should be spend with them. Maybe 'Cross Mode' (the 'M' needs to be uppercase like it is in 'Tone Mode') could mention something like...
> 
> 'Complex or non-standard tone squelch mode (starts the tone selection wizard)'

Yeah, it's probably best to keep them reasonably short (i.e. a few sentences at most) and we can't really explain everything about how to program a radio in a tooltip. But, it'd be nice to provide some hints.

Should I merge this as-is or do you want to mull the existing things for a while? The only reason not to merge is that @MELERIX will show up tomorrow with translations because he's super fast like that :) Any time we change these, the translators will have to update them, so we shouldn't do it too often.

KC9HI: Sorry, I went QRT before you sent your last reply. I would support merging this 'as-is'. Thanks!

PR #956, opened 2024-02-22
Allow deleting errored memories

kk7ds:
- uvk5: Fix non-byteclean set_raw() usage
- Allow deleting error'd memories

### DIFF
--- a/chirp/wxui/memedit.py
+++ b/chirp/wxui/memedit.py
@@ -1461,6 +1461,13 @@ class ChirpMemEdit(common.ChirpEditor, common.ChirpSyncEditor):
         for row in rows:
             if not self._memory_cache[row].empty:
                 self.delete_memory_at(row, event)
+            elif row in self._memory_errors:
+                LOG.warning('Attempting to clear memory error '
+                            '%r for memory %i at row %i with delete',
+                            self._memory_errors[row], self.row2mem(row), row)
+                self.delete_memory_at(row, event)
+            else:
+                LOG.debug('Not re-deleting empty memory at row %i', row)
 
         if not shift_up:
             return
--- a/chirp/drivers/uvk5.py
+++ b/chirp/drivers/uvk5.py
@@ -1941,10 +1941,10 @@ class UVK5RadioBase(chirp_common.CloneModeRadio):
         _mem4 = self._memobj
         # empty memory
         if mem.empty:
-            _mem.set_raw("\xFF" * 16)
+            _mem.set_raw(b"\xFF" * 16)
             if number < 200:
                 _mem2 = self._memobj.channelname[number]
-                _mem2.set_raw("\xFF" * 16)
+                _mem2.set_raw(b"\xFF" * 16)
                 _mem4.channel_attributes[number].is_scanlist1 = 0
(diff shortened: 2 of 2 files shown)


chirp/wxui/memedit.py
```diff
@@ -1461,6 +1461,13 @@ def _delete_memories_at(self, rows, event, shift_up=None):
         for row in rows:
             if not self._memory_cache[row].empty:
                 self.delete_memory_at(row, event)
+            elif row in self._memory_errors:
+                LOG.warning('Attempting to clear memory error '
+                            '%r for memory %i at row %i with delete',
+                            self._memory_errors[row], self.row2mem(row), row)
+                self.delete_memory_at(row, event)
```
kk7ds, line 1468: @hgrasman This should fix the UI and make it possible to delete memories that failed to load. So at least you can try to clean it up from the UI by deleting a bad memory and letting the driver wipe it in `set_memory()` or `erase_memory()` depending.

hgrasman: Hey Dan, I don't think this has fixed the issue. It would seem the driver is not evaluating the usedflags after deletion, only during setting. 

This goes for already empty memories, and memories with values. 

I don't know why, or how to find out. Any advice?

kk7ds: The only thing this does is allow you to run erase/set_memory on memories that failed to load. The driver clearly is still broken in get_memory() some way, causing it to say a memory is non-empty when it should be. That's going to have to be a driver fix of course.

hgrasman: Right. My understanding of the issue is that this flag is evaluated during set_memory calls and not get_memory calls (as to not modify the memory on read).

This issue presented itself in two ways:
1) set_memory returns on an empty memory before evaluating flags. I just fixed this. Deleting a memory now sets the flags correctly.
2) set_memory is not called when attempting to delete a memory that is already empty? Which means the faulty memories cannot be repaired without writing something in and then deleting. Which is probably okay? because the fix to #1 should prevent this from ever happening again. But that also means this method of fixing the damage that already exists in some images does not work. 

For your consideration. I'll be finishing up the fix for #1 soon, so this may all be a moot point. I made this comment before figuring out that the two are unrelated.

kk7ds: > Right. My understanding of the issue is that this flag is evaluated during set_memory calls and not get_memory calls (as to not modify the memory on read).
> 
> This issue presented itself in two ways:
> 
> 1. set_memory returns on an empty memory before evaluating flags. I just fixed this. Deleting a memory now sets the flags correctly.

Do you mean `get_memory()` here?

> 2. set_memory is not called when attempting to delete a memory that is already empty? Which means the faulty memories cannot be repaired without writing something in and then deleting. Which is probably okay because the fix to 1) should prevent this from ever happening again. But that also means this method of fixing the damage that already exists in some images does not work.

Yeah, delete is not run on empty memories, for obvious reasons.
 
> For your consideration. I'll be finishing up the fix for 1) soon, so this may all be a moot point. I made my last comment before figuring out that the two are unrelated.

Based on the above question, I'm not sure whether the problem is in get or set. But, push it up and let's discuss in context there instead of on this merged ui patch :)

PR #964, opened 2024-02-24
icf: Warn if more than one response received

kk7ds:


kk7ds:
Not needed, bug was caused by a hardware issue.

PR #968, opened 2024-02-27
K5 Plus: Register new model -> child of 5RM driver

hgrasman:
Fixes: #11119

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/baofeng_uv17Pro.py
+++ b/chirp/drivers/baofeng_uv17Pro.py
@@ -1332,3 +1332,9 @@ class BF5RM(UV17Pro):
             except ValueError:
                 pass
         super().check_set_memory_immutable_policy(existing, new)
+
+
+@directory.register
+class BFK5Plus(BF5RM):
+    VENDOR = "Baofeng"
+    MODEL = "K5-Plus"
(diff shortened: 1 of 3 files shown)


hgrasman:
Radio Baofeng_K5_Plus should ~~not~~ override check_set_memory_immutable_policy

Just like the 17 pro and 5RM

hgrasman:
@kk7ds 

Hey Dan. Now sure how to decypher what's going on here. The offending code seems to be something you added to the other models after some of my changes. I'm about to add another model inheriting from this driver, so I'd like to understand.

kk7ds:
Which are you talking about... the unit failure? If so, you need to add this radio to the immutable whitelist in the test file (`test_chirp_common`) **if** the airband override in the 5RM applies to the 5RH as well. If it doesn't, then it shouldn't be a subclass of it and instead should subclass the 17Pro itself.

hgrasman:
> Which are you talking about... the unit failure? If so, you need to add this radio to the immutable whitelist in the test file (`test_chirp_common`) **if** the airband override in the 5RM applies to the 5RH as well. If it doesn't, then it shouldn't be a subclass of it and instead should subclass the 17Pro itself.

I suppose I don't know the answer to that, other than that the new radio in question is "identical" with a different form factor. So I would assume that's correct. Unless you oppose, I'll add it. 

I assume the test matrix is mad because I added a model without updating unit tests? Do you have a one-pager tutorial for how to do that properly?

kk7ds:
If you don't know for sure I I think you should not subclass the special one. We can always fix it later if people complain. Perhaps you can tell from the public specs whether it supports airband?

For the matrix just rebase on master, delete the file and re-run the `makesupported` to regenerate the file, commit/resolve/whatever and you should be good. If you cant figure it out I can do it for you.

hgrasman:
> If you don't know for sure I I think you should not subclass the special one. We can always fix it later if people complain. Perhaps you can tell from the public specs whether it supports airband?
> 
> For the matrix just rebase on master, delete the file and re-run the `makesupported` to regenerate the file, commit/resolve/whatever and you should be good. If you cant figure it out I can do it for you.

I see now that this is just enforcing the AM mode. Yes it is appropriate for this radio, so I'll add the whitelist. This radio is a 5RM in hardware and firmware (just a cheap attempt to steal quansheng k5 customers).

I'll see if I can figure out the test matrix. I'm just ignorant about how that works at the moment.

hgrasman:
@kk7ds 

Update I don't think I'm figuring this out. I think I've got a 40% correct guess on what's going on. If you're up for it, please explain what to do, and do it for me this time.

PR #969, opened 2024-02-28
TIDRADIO: Fixed false empty channels on deletion with robust usedflag…

hgrasman:
…s logic.

FIXES: #11131
FIXES: #11207

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -129,16 +129,7 @@ struct{
 }fmmode[25];
 
 #seekto 0x1A08;
-struct{
-    u8 block8:1,
-       block7:1,
-       block6:1,
-       block5:1,
-       block4:1,
-       block3:1,
-       block2:1,
-       block1:1;
-} usedflags[25];
+bit usedflags[200];
 
 #seekto 0x1a28;
 struct{
@@ -204,16 +195,7 @@ struct {
 } vfob;
 
 #seekto 0x1B78;
-struct{
-   u8  block8:1,
-       block7:1,
-       block6:1,
-       block5:1,
-       block4:1,
-       block3:1,
-       block2:1,
-       block1:1;
-} fmusedflags[4];
+bit fmusedflags[32];
 
 #seekto 0x1c08;
 struct {
@@ -474,16 +456,7 @@ struct{
 }endcode;
 
 #seekto 0x1908;
-struct{
-    u8 block8:1,
-       block7:1,
-       block6:1,
-       block5:1,
-       block4:1,
-       block3:1,
-       block2:1,
-       block1:1;
-} usedflags[25];
+bit usedflags[200];
 
 #seekto 0x1928;
 struct{
@@ -498,16 +471,7 @@ struct{
 } scanadd[25];
 
 #seekto 0x1948;
-struct{
-  u8  block8:1,
-       block7:1,
-       block6:1,
-       block5:1,
-       block4:1,
-       block3:1,
-       block2:1,
-       block1:1;
-} fmusedflags[4];
+bit fmusedflags[32];
 
 #seekto 0x1958;
 struct {
@@ -1067,12 +1031,6 @@ class TDH8(chirp_common.CloneModeRadio):
     def _get_scan(self, number):
         return self._memobj.scanadd[number]
 
-    def _get_block_data(self, number):
-        return self._memobj.usedflags[number]
-
-    def _get_fmblock_data(self, number):
-        return self._memobj.fmusedflags[number]
-
     def get_memory(self, number):
         _mem = self._get_mem(number)
         _nam = self._get_nam(number)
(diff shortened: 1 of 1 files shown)


chirp/drivers/tdh8.py
```diff
@@ -1315,6 +1302,22 @@ def set_memory(self, mem):
...
                 _usedflags.block8 = 0
 
+    def set_memory(self, mem):
+        _mem = self._get_mem(mem.number)
+        _nam = self._get_nam(mem.number)
+ 
+        self.evaluate_usedflags(mem)
+
+        if mem.empty:
+            _mem.set_raw("\xff" * 16)
```
kk7ds, line 1312: Use `fill_raw()` and you won't have to do the size.

chirp/drivers/tdh8.py
```diff
@@ -1315,6 +1302,22 @@ def set_memory(self, mem):
...
+            if self.MODEL in H8_LIST:
+                _nam.set_raw("\xff" * 16)
```
kk7ds, line 1314: And if the structures are right, `fill_raw()` here will avoid the need for the conditional.

chirp/drivers/tdh8.py
```diff
@@ -1254,20 +1254,7 @@ def get_block_xy(self, number):
...
-        if mem.empty:
-            _mem.set_raw("\xff" * 16)
-            if self.MODEL in H8_LIST:
-                _nam.set_raw("\xff" * 16)
-            else:
-                _nam.set_raw("\xff" * 8)
-            return
-
-        _mem.set_raw("\x00" * 16)
-
```
kk7ds, line 1270: Okay so the fix is to do the (ridiculous) flag setting below all the time, and *then* mark out the memory and bail after that right?

Can we name it something other than "evaluate_usedflags". It's not really evaluating anything, it's setting, updating, etc.

Also, when I was reviewing this driver I tried to help the submitter make it less crazy than the if-per-bit thing. Can we refactor that as part of this? I can stack something on top here if you want.

hgrasman:
@kk7ds 

-When/after modifying a memory, set_memory() within the driver is called, which evaluates the usedflag. I didn't design that, but I've observed it to be true. Previously that function returned early on an empty memory, causing the usedflag to remain 1 forever once set. The logic to evaluate the usedflags has been moved before that check, and now deleting memory works as expected.

-The other item I mentioned is that chirp ignores attempts to delete already empty memory. That's a great idea, but it prevents cleaning up these errored flags if they already exist. I guess you could program a memory in and then delete it, but that's roundabout.

chirp/drivers/tdh8.py
hgrasman: Saw this after typing my shpeil:

Yes. Simple fix, elusive problem.

Sure. I'll think on it

Sure. It aggrivated me too. I can do that quick.

kk7ds:
> -The other item I mentioned is that chirp ignores attempts to delete already empty memory. That's a great idea, but it prevents cleaning up these errored flags if they already exist. I guess you could program a memory in and then delete it, but that's roundabout.

Yeah, but I don't think there's a cleaner way to do that in a way that doesn't break the model for everyone just to fix this driver's issue. So I think it's best to just let people fix this one time, once it's correct.

TBH, I'm not sure why this doesn't fail tests, so maybe I should look and see if there's a gap there that I could add to catch this in the future (or other broken drivers).

chirp/drivers/tdh8.py
```diff
@@ -1254,66 +1240,27 @@ def get_block_xy(self, number):
...
+    def set_usedflags(self, mem):
+        # When the channel is empty, you need to set "usedflags" to 0,
+        # When the channel is used , you need to set "usedflags" to 1. 
+        block_xy = self.get_block_xy(mem.number)
+        _usedflags = self._get_block_data(block_xy[0])
+        if mem.empty:
+            _usedflags.flags &= ~( 1 << (block_xy[1] - 1))
+        else:
+            _usedflags.flags &= (1 << (block_xy[1] - 1))
+
```
kk7ds, line 1252: Much better, thanks :)

chirp/drivers/tdh8.py
```diff
@@ -130,14 +130,7 @@
...
 struct{
-    u8 block8:1,
-       block7:1,
-       block6:1,
-       block5:1,
-       block4:1,
-       block3:1,
-       block2:1,
-       block1:1;
+    u8 flags;
```
kk7ds, line 133: This doesn't need to be inside a struct anmore, it can just be at the top level. But, there's also a "bit" type which never gets used, but it should be used more. So:
```
bit flags[200];
```

And then you can do something like:
```
memobj.flags[memory.number - 1] = int(mem.empty)
```
And not have to do the masking below.

Neither are critical, but just FYI if you want to clean it up more ;)

hgrasman:
@kk7ds 

Should be pretty now. fmused flags got the same treatment

chirp/drivers/tdh8.py
```diff
@@ -1243,77 +1201,19 @@ def _get_scan_list(self, scan_data):
...
-
-        return ([x_list, y_list])
-
     def set_memory(self, mem):
         _mem = self._get_mem(mem.number)
         _nam = self._get_nam(mem.number)
 
+        # When the channel is empty, you need to set "usedflags" to 0,
+        # When the channel is used , you need to set "usedflags" to 1.
+        self._memobj.usedflags[mem.number - 1] = int(~mem.empty)
```
kk7ds, line 1210: Noice, this is very clean. But, I think you should use `not` here instead of `~`:
```
>>> not False
True
>>> ~False
-1
```
So you're actually setting -1 instead of 1, which based on the bit math in the bitwise object will make this weird. I assume you tested this with the radio?

```
>>> 0x80 |  (-1 << 3)
-8
>>> 0x80 |  (1 << 3)
136
```
I think it will still be okay because of the masking that goes on, but it's not super great for understandability in the future :)

kk7ds:
Okay yeah, I did some poking at the memory object and it is doing the right thing, it's just confusing. I need to do some cleanup of deprecated items in here too, so I'll hit this when I do just to make that more intuitive. Thanks!

kk7ds:
Unit failure is unrelated - DMRMARC is down or something.

PR #970, opened 2024-02-28
baofeng_common.py: gracefully handle bcd coding errors - fixes #11183

KC9HI:
The BTech UV-5X3 do not ship from the factory with valid values for all of the subvfo settings. This causes the Settings tabs to not display.

This patch allows the tabs to display and highlights the settings that need a valid value.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/uv5x3.py
+++ b/chirp/drivers/uv5x3.py
@@ -23,7 +23,7 @@ from chirp.settings import RadioSettingGroup, RadioSetting, \
     RadioSettingValueBoolean, RadioSettingValueList, \
     RadioSettingValueString, RadioSettingValueInteger, \
     RadioSettingValueFloat, RadioSettings, \
-    InvalidValueError
+    InvalidValueError, InternalError
 
 LOG = logging.getLogger(__name__)
 
@@ -136,7 +136,7 @@ class UV5X3(bfc.BaofengCommonHT):
     SCODE_LIST = LIST_SCODE
 
     MEM_FORMAT = """
-    #seekto 0x0000;
+    // #seekto 0x0000;
     struct {
       lbcd rxfreq[4];
       lbcd txfreq[4];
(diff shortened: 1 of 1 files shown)


chirp/drivers/baofeng_common.py
```diff
@@ -448,8 +448,8 @@ def bcd_decode_freq(bytes):
     real_freq = 0
     for byte in bytes:
         if byte > 9:
-            msg = ("Found Invalid BCD Encoding")
-            raise InternalError(msg)
+            LOG.debug('Found Invalid BCD Encoding')
+            return chirp_common.format_freq(0)
```
kk7ds, line 452: I think this is the wrong thing to do here. I think this generalized helper *should* fail when we can't decode something that should be BCD. Instead, we should catch the failure where it happens and make the decision there about what to do. I think the way to tell it's wrong is that we'll see a random "Invalid BCD encoding" in the log, but we won't know what was actually wrong. Would doing it differently be okay?

So if you don't change this and leave the raise, put the exception handling where you use it:
```
try:
    value = bfc.bcd_decode_freq(foo)
except settings.InternalError:
    LOG.debug('Failed to decode VFOb')  # or whatever
    value = chirp_common.format_freq(0)
```
What do you think?

KC9HI: Let me give it a try. Thanks. It might not be until tomorrow evening.

KC9HI: My initial testing looks like this is going to work. I will expand it to the other 5 subVFO settings and perform better testing tomorrow. Thanks again!

kk7ds: Cool, thanks!

KC9HI:
@kk7ds This turned out really good. I have gone back to load images from 2016 that wouldn't show the Settings tabs and they all now open with a highlighted setting wherever a valid frequency was not found.

chirp/drivers/uv5x3.py
```diff
@@ -855,54 +855,72 @@ def apply_freq(setting, obj):
         rs.set_apply_callback(apply_freq, _mem.vfo.b)
         work.append(rs)
 
-        val = bfc.bcd_decode_freq(_mem.subvfoa.vhf.freq)
-        if int(float(val)) == 0:
-            val = str(int(_mem.limits.vhf.lower)) + ".000000"
+        try:
+            val = ""
```
kk7ds, line 859: Hmm, why do you have these? They're unnecessary (as long as you have the thing on L863 below in the exception handlers. I can clean these up before merge if you want.

Otherwise yeah, this looks much nicer!

KC9HI: The reason is because without them, if the there is a bcd coding error, the 'val' from the previous vfo a/b will be used. It bugged that the value from vhf2 would end up in uhf if the image didn't have a uhf value established at the factory. So I guess using something like val1, val2, val3, val4, val5 and val6 could be used to avoid the problem. Do you have a different idea?

kk7ds: Erm, I don't see any way that could be true. As soon as L860 runs, this copy of val is gone. If that fails to run, then L863 will wipe it away. In either case, this value is never seen by the first line that actually uses it, which is L864.

I'm wondering if you saw that behavior before you added the `val = "000.000"` in the exception handler of each? Because if you didn't have that, then yeah, a failure in the "VFO B VHF" case would re-use the value from the "VFO A VHF" case. But what you have here can't.

If I'm wrong, can you tell me how to repro? Shoot me an image or something and I'll test loading it with these lines gone and see what you see.

KC9HI: You may be right. Let me test it again.

KC9HI:
You were right. It must have been the order that I did my testing.

kk7ds:
> You were right. It must have been the order that I did my testing.

Nice, thanks. One of the benefits of being older than dirt (talking about myself here) is having made a bunch of those mistakes myself so I can "reconstruct the crime scene" when others do :P

PR #971, opened 2024-02-28
Update Spanish Translation

MELERIX:
minor fixes for consistency


kk7ds:
This just failed because repeaterbook was slow.

PR #972, opened 2024-02-29
Fix TD-H8 overlapping memory

kk7ds:
- tdh8: Fix bitwise errors and memory structure
- Show end address in browser

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -54,12 +54,6 @@ struct {
   u8 unused10;
 } memory[200];
 
-#seekto 0x0D38;
-struct {
-  char name[8];
-  u8 unknown2[8];
-} names[200];
-
 #seekto 0x0CA8;
 struct {
   u8 txled:1,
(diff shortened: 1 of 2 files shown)


kk7ds:
@hgrasman Here's my deprecation and error cleanup, with the "not" change.

Note that the errors were signaling an overlapping memory structure, which is why the H8 and H3 were having to index into that array with N and N-1 respectively. The original author just started that too early, overlapping with the previous structure, and then never used the `[0]` spot. 🤦

hgrasman:
I'll never talk poorly of someone else's free labor, but this driver lives in my nightmares.

I don't even have a TIDRADIO anymore

Thanks for cleaning up my strange math. I'm a matlab guy by day, and the ~ as a logical not worked it's way in. The fact that it worked was pure luck 😂

kk7ds:
Yeah, I had lots of "feedback" for this one before merge, but it was just not happening so I cut my losses :)

Anyway, I think this is good to go, so I'll merge.

PR #977, opened 2024-03-05
Fixes issue #11221

vdwel:
UV13Pro and UV17: channel names were displayed wrong above channel 372

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/baofeng_uv17.py
+++ b/chirp/drivers/baofeng_uv17.py
@@ -290,9 +290,15 @@ class UV17(baofeng_uv17Pro.UV17Pro):
     } vfo;
 
     #seekto 0x4000;
-    struct {
+    struct channelname {
       char name[11];
-    } names[999];
+    };
+
+    struct channelname names1[372];
+    #seek 0x4;
+    struct channelname names2[372];
+    #seek 0x4;
+    struct channelname names3[255];
 
     #seekto 0x8000;
     struct {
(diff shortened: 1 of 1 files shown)


chirp/drivers/baofeng_uv17.py
```diff
@@ -290,9 +290,23 @@ class UV17(baofeng_uv17Pro.UV17Pro):
...
+
+    #seek 0x4;
+    struct {
+      struct channelname names[372];
+    } names2;
+
+    #seek 0x4;
+    struct {
+      struct channelname names[255];
+    } names3;
```
kk7ds, line 309: Why put these in structs? That just increases the indirection you need. Just do this:
```
struct channelname names1[372];
#seek 4;
struct channelname names2[372];
#seek 4;
struct channelname names3[255];
```

chirp/drivers/baofeng_uv17.py
```diff
@@ -408,9 +422,20 @@ def get_raw_memory(self, number):
         _mem = self._memobj.mem1.mem[number]
         return _mem
 
+    def get_channel_name(self, number):
+        number = number - 1
+        if number >= 744:
+            _name = self._memobj.names3.names[number - 744]
```
kk7ds, line 428: Then this can be just:
```
self._memobj.names3[number - 744]
```

kk7ds (commented):
Cool, glad that was an easy fix, thanks! I think you're making the memory object more complicated than you need to. I can fix this for you if you can't do it quickly.

kk7ds:
Also, I'll fix the commit message. It's sort of upside down. The second line that describes what it does should be first, because that will be sent to the mailing list. And the fixes tag needs to not have "issue" in between "fixes" and the number or it won't properly reference the actual ticket.

PR #988, opened 2024-03-12
snapcraft: use archive.chirpmyradio.com

goldstar611:
(my sync script removed my change from master branch, I forgot about that)

Congrats on the new domain, Dan!

This PR updates the snapcraft.yml file so that the final curl command gets the tarball and not another redirect.

Tony

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -43,9 +43,9 @@ parts:
       - PYTHONPATH: ""
     override-pull: |
       sudo apt install -y curl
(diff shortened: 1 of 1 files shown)


kk7ds:
Sorry, I should have thought to give you a heads up before I put this redirect in place... I'll merge this when we get some content for the next build, thanks!

goldstar611:
No worries, it's on me for not using the follow redirects parameter in curl.

kk7ds:
That gets me every time too. I wish it was the default instead of the exception, but I know there are good reasons for it.

PR #989, opened 2024-03-14
TIDRADIO H3/H8 fixes

kk7ds:
tdh8: Fix TD-H3 power levels
tdh8: Simplify and fix offset/split/off decoding
tdh8: Simplify skip/add data structure
tdh8: Fix byte ordering for used flags
tdh8: Late-evaluate mic gain setting

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -117,7 +117,7 @@ struct {
 
 #seekto 0x0CD8;
 struct{
-    u8 fmblock[4];
+    lbcd fmblock[4];
 }fmmode[25];
 
 #seekto 0x0D48;
@@ -133,9 +133,7 @@ lbit usedflags[200];
 lbit scanadd[200];
 
 #seekto 0x1B38;
-struct{
-    u8 vfo[4];
-}fmvfo;
+lbcd fmvfo[4];
 
 #seekto 0x1B58;
 struct {
@@ -367,7 +365,7 @@ struct {
 
 #seekto 0x0CD8;
 struct{
-    u8 fmblock[4];
+    lbcd fmblock[4];
 }fmmode[25];
 
 #seekto 0x0D48;
@@ -497,10 +495,7 @@ struct {
 } vfob;
 
 //#seekto 0x1978;
-struct{
-    u8 vfo[4];
-}fmvfo;
-
+lbcd fmvfo[4];
 
 #seekto 0x1c08;
 struct {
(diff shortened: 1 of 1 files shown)


kk7ds:
Even the out of range thing is not quite right, but perhaps just because of the used-or-not flags, per this bug:

https://chirpmyradio.com/issues/11233

Also maybe @hgrasman in case you didn't see my mention on the bug tracker, there are still issues with the used flags and this driver:

https://chirpmyradio.com/issues/11244

hgrasman:
Hey all,

I’ve never owned an h3, but I assume that’s a simple band range define
incorrectly. I know this isn’t directed at me, but that’s where I’d look.

Ah rats. I don’t know if I will be able to look into it any time soon :/
I’ll take a quick one now, but no promises.

-Henry

hgrasman:
Update: As expected, the ranges were all wrong. Check out my pull request
as a good starting point.

It passed all the driver tests, so...
I even almost got syle right

-H

songjunbiao:
I pulled down that branch and realized that the contents of the dropdown box were not showing up... am I doing something wrong?

kk7ds:
@songjunbiao which drop-down?

songjunbiao:
All drop-down menus for optional functions

songjunbiao:
image
Here's the picture I provided

songjunbiao:
I'll take your advice.

songjunbiao:
I've fixed the problem with the feedback above. How do I submit the code?

kk7ds:
@songjunbiao which problem have you fixed exactly and how? Feel free to submit a new PR if you like and you can include these commits if necessary.

songjunbiao:
We're ready to offer you samples of H3 and H8. 
Do you need a model number agreement? 
The problem with some of the channels not being written is that during the merger, you optimized some of the code so that it didn't match the protocols, and then you couldn't write to the new channels.

songjunbiao:
We've sent you H3 and H8. 
Can we work together to fix the problem?
I've been looking at your code optimization, and it doesn't match the assisted design. 
I think I need to inform you of our agreement.

kk7ds:
Yeah, If you can explain the problem I'm happy to work on it. The optimization came because the original code was clearly not right and many users were complaining. We tried to fix it but without access to the radios we were not sure, which is why I haven't merged this subsequent fix and was looking for help from you.

songjunbiao:
I'm sending you the protocols for H3 and H8 to your e-mail.
And you'll have the H3 and H8 machines in about a week. 
You'll help me fix the bugs then, okay?

songjunbiao:
The usedflags issue has been resolved, and the newly programmed channels are now in effect. 
Thank you very much. I've tested it. 
bug/11233/tdh8range There are still a couple of problems with the branch.
Settings The list of all dropdown boxes is not displayed.
Extra for channels is blank
Settings FM Channel 1 is not working (fmusedflags) FM Other channels are working.
I'll keep testing. Thanks for the fix.

songjunbiao:
image
Optional function with drop-down box Not shown anymore
image
Extra is blank.

kk7ds:
The drop-down thing is unrelated to this driver and just because this is based on an earlier commit in the middle of a change to the settings editor. I just rebased this on master where the issue is resolved, so if you pull again it should work.

Thanks for confirming the usedflags fix. I will take a look at the FM thing.

kk7ds:
Okay, the FM code was very complex and buggy as well, so I simplified it a lot and it appears to work for me, but if you could test with a radio that would be great.

songjunbiao:
Thank you very much for the fix. 
Once you've received the radio we sent you, let's talk about any questions or suggestions you may have about the experience.
I'll keep testing it, and if there are any problems, I'll give you feedback and suggestions.

PR #990, opened 2024-03-14
Tidradio H3: Fixed ranges to include all correct bands

hgrasman:
# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -893,8 +893,7 @@ class TDH8(chirp_common.CloneModeRadio):
     _memsize = 0x1eef
     _ranges_main = [(0x0000, 0x1eef)]
     _idents = [TD_H8]
-    _vhf_range = (136000000, 175000000)
-    _uhf_range = (400000000, 521000000)
+    _bands = [(136000000, 175000000), (400000000, 521000000)]
     _aux_block = True
     _tri_power = True
     _gmrs = False
@@ -952,10 +951,10 @@ class TDH8(chirp_common.CloneModeRadio):
         rf.valid_modes = ["FM", "NFM"]
         rf.valid_tuning_steps = STEPS
 
-        normal_bands = [self._vhf_range, self._uhf_range]
+        normal_bands = self._bands
 
         if self._mmap is None:
-            rf.valid_bands = [normal_bands[0], normal_bands[1]]
+            rf.valid_bands = [b for b in normal_bands]
         else:
             rf.valid_bands = normal_bands
         rf.memory_bounds = (1, 199)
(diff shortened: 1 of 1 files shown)


hgrasman:
Untested, unfinished (probably). Please try this (someone with an H3 and more time)

@songjunbiao @kk7ds

songjunbiao:
TD-H3 Ham and TD-H3 GMRS can only support the same range as the original one, only TD-H3 can support more range.

chirp/drivers/tdh8.py
```diff
@@ -1646,11 +1645,7 @@ def _filter(name):
         _fmv_temp = ''
         for x in range(_fmv_len, -1, -1):
             _fmv_temp += str(_fmv[x])[2:]
-        try:
-            _fmv_block = int(_fmv_temp)
-        except ValueError:
-            LOG.error('Invalid fmv_block %s' % _fmv_temp)
-            _fmv_block = 0
```
kk7ds, line 1653: Why are you reverting this change? It was a recent fix for a reason.

chirp/drivers/tdh8.py
```diff
@@ -2340,8 +2337,10 @@ class TDH3_GMRS(TDH3):
     MODEL = "TD-H3-GMRS"
     ident_mode = b'P31184\xff\xff'
     _gmrs = True
-    _vhf_range = (136000000, 175000000)
-    _uhf_range = (400000000, 521000000)
+    _bands = [(50000000, 76000000), (76000000, 108000000),
+              (108000000, 136000000), (136000000, 174000000),
+              (174000000, 350000000), (350000000, 400000000),
+              (400000000, 470000000), (470000000, 600000000)]
```
kk7ds, line 2343: Okay, but if you look at the issue attached to my attempt, this didn't work for the reporter. I'm not sure if it didn't work because of the used flags thing, or because there is more required to be done in the driver, which is what I was hoping to get feedback on. You don't have an H3 so you can't actually test that this works right?

chirp/drivers/tdh8.py
hgrasman: I don't remember editing this, I must have fat thumbed something

chirp/drivers/tdh8.py
hgrasman: Correct, I cannot test. 

I couldn't find anything that references the _uhf_range etc (now _bands) either. So I assumed by updating all of those instances it would have the correct bands available. If that doesn't work, it might be hard-coded somewhere. I don't have a lot of time lately, hoped this would work/move the right direction at least

kk7ds: Okay I missed the request for test above, so I see now.

@songjunbiao will you confirm if these band expansions work properly?

songjunbiao: I've taken his advice and merged the codes, so far only the H3's work, the others don't, and I've wrapped H3 GMRS and H3 Ham.

kk7ds:
Okay, so I'm going to stick with my in-progress band expansion patch instead of this, @hgrasman because it addresses the receive-only-ness aspect, which we need for some of the bands. But I'll try to incorporate some of the cleanup you did here to avoid the `_vhf` and `_uhf` stuff. So, we can close this I think.

hgrasman:
> and

Fine by me! 

I threw that together quick thinking that other person was planning to work on it more, but didn't know where to start.

kk7ds:
Yeah, we've got three parallel threads here and I need to collapse that. But I think I see the problem and it stems from me making you convert to the bit array, so it's my mess to clean up.

PR #991, opened 2024-03-15
Bug/11233/tdh8range

songjunbiao:
Fix inability to write new channel information
and add H3 frequency range
and bug fixes

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -29,7 +29,7 @@ from chirp import bandplan_na
 
 LOG = logging.getLogger(__name__)
 
-
+AIRBAND = (108000000, 135999999)
 MEM_FORMAT = """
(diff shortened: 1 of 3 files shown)


kk7ds:
This reverts a lot of cleanups we have made and moves back to individual elements for each of the flags. That's not okay and also matches the original code that the users reported does not work properly. Please help me understand what is wrong about the current code, or please iterate from there and don't reintroduce the inefficient flag scheme.

songjunbiao:
I'll email you the H3 and H8 agreements.
In about a week, you'll have the H3 and H8 machines.
When you do, you can help me fix the bugs, okay?
Thank you very much.

kk7ds:
Okay, I'm going to close this now as I think we're resolved after merging of #989. Thanks!

PR #998, opened 2024-03-19
Related to #11255 - Enable offsets to exceed band limits for radios that support it

ho0ber:
Related to #11255 - I think it's likely this approach might need to change, but didn't want to submit the feature request empty-handed.

I'm happy to add testing as well, once I get the test suite working locally.

# CHIRP PR Checklist

* All tests must be passing.

### DIFF
--- a/chirp/drivers/ft4.py
+++ b/chirp/drivers/ft4.py
@@ -677,6 +677,7 @@ class YaesuSC35GenericRadio(chirp_common.CloneModeRadio,
     BAUD_RATE = 9600
     MAX_MEM_SLOT = 200
     NEEDS_COMPAT_SERIAL = False
+    DUPLEX_OFF_VIA_OFFSET = False
 
     # These settings are common to all radios in this family.
     _valid_chars = chirp_common.CHARSET_ASCII
@@ -1149,11 +1150,33 @@ class YaesuSC35GenericRadio(chirp_common.CloneModeRadio,
                 mem.offset = txfreq
             else:
                 mem.offset = int(_mem.offset) * 25000 * freq_offset_factor
+            if self.DUPLEX_OFF_VIA_OFFSET and mem.duplex in ['+', '-']:
+                band = self.band_for_freq(mem.freq)
+                if band is not None and self.DUPLEX_OFF_VIA_OFFSET[band]:
+                    sign = 1 if mem.duplex == "+" else -1
+                    tx = mem.freq + (sign * mem.offset)
+                    if self.band_for_freq(tx) is None:
+                        mem.duplex = "off"
+                        mem.offset = 0
             self.decode_sql(mem, _mem)
             mem.power = POWER_LEVELS[2 - _mem.tx_pwr]
             mem.mode = ["FM", "NFM"][_mem.tx_width]
             mem.tuning_step = STEP_CODE[_mem.step]
         return mem
+
+    def band_for_freq(self, freq):
+        """
+        We need to find the valid_bands list index a frequency falls
+        within, if any, to help determine if DUPLEX_OFF_VIA_OFFSET
+        logic is required
+        """
+        band_index = None
+        for i, limits in enumerate(self.valid_bands):
+            lo, hi = limits
+            if lo <= freq <= hi:
+                band_index = i
+                break
+        return band_index
 
     def enforce_band(self, memloc, freq, mem_num, sname):
         """
@@ -1189,20 +1212,26 @@ class YaesuSC35GenericRadio(chirp_common.CloneModeRadio,
         _mem.tx_width = mem.mode == "NFM"
         _mem.step = STEP_CODE.index(mem.tuning_step)
 
+        duplex = mem.duplex
+        offset = mem.offset
+        if self.DUPLEX_OFF_VIA_OFFSET and duplex == "off":
+            band = self.band_for_freq(mem.freq)
+            if band is not None and self.DUPLEX_OFF_VIA_OFFSET[band]:
+                duplex, offset = self.DUPLEX_OFF_VIA_OFFSET[band]
+
         freq_offset_factor = self.freq_offset_factor
         # FT-25R/65R Asia version (US version is 0)?
         if self.subtype == 3 and freq_offset_factor == 2:
             freq_offset_factor = 1  # 25000 scaler
-        _mem.offset = mem.offset / (25000 * freq_offset_factor)
-        duplex = mem.duplex
+        _mem.offset = offset / (25000 * freq_offset_factor)
         if regtype in ["memory", "pms"]:
             ndx = num - 1
             store_bit(self._memobj.enable, ndx, True)
             store_bit(self._memobj.scan, ndx, SKIPS.index(mem.skip))
             nametrim = (mem.name + "        ")[:8]
             self._memobj.names[ndx].chrs = bytearray(nametrim, "ascii")
-            if mem.duplex == "split":
-                txfreq = mem.offset / 10
+            if duplex == "split":
+                txfreq = offset / 10
             self._memobj.txfreqs[num-1].freq = txfreq
         _mem.duplex = DUPLEX.index(duplex)
         if regtype in ["vfo", "home"]:
@@ -1331,6 +1360,11 @@ class YaesuFT65RRadio(YaesuFT65GenericRadio):
     DUPLEX_AUTO = DUPLEX_AUTO_US
     legal_steps = STEP_CODE
     BAND_ASSIGNMENTS = BAND_ASSIGNMENTS_DUALBAND
+    DUPLEX_OFF_VIA_OFFSET = [  # matches the order of valid_bands
+        None,  # Don't modify broadcast FM memories
+        ("-", 99_950_000),
+        ("+", 99_950_000),
+        ]
 
 
 @directory.register
--- a/tests/base.py
+++ b/tests/base.py
@@ -132,6 +132,14 @@ class DriverTest(unittest.TestCase):
     def assertEqualMem(self, a, b, ignore=None):
         if a.tmode == "Cross":
             tx_mode, rx_mode = a.cross_mode.split("->")
+        if ignore is None:
+            ignore = []
+
+        if a.duplex == b.duplex == 'off':
+            # If we're asking for duplex=off, we should not obsess over the
+            # driver keeping track of our offset, as it may use the offset
+            # field to manage the TX behavior
+            ignore.append('offset')
 
         a_vals = {}
         b_vals = {}


chirp/drivers/ft4.py
```diff
@@ -616,6 +616,9 @@ def get_memory_mappings(self, memory):
     (136000000, 174000000),    # VHF
     ]
 
+# Presumably the same for both models, this can be added to FT65E once tested
+VALID_OFFSETS_FT65 = (50_000, 99_950_000)
```
kk7ds, line 620: I'm not sure what the units or meaning of this is really.. 50kHz to 99.950MHz?

kk7ds (commented):
I really don't like this, for a number of reasons, but I think you probably expected that possibility, from the verbiage in the bug (thanks for that btw).

I think that achieving tx-inhibit in this way is pretty strange and unfortunate if the radio really doesn't have that ability. Also, the driver already claims to support duplex=off (it's not just a baofeng thing). Are you perhaps just suggesting that the only way to achieve duplex=off _on the radio_ is by setting this invalid offset or something?

Wouldn't it be possible to handle this internal to the radio and basically say if any memory has a duplex with an offset that is outside the valid range, we consider that to be duplex=off for chirp's purposes? It wouldn't match the radio's behavior exactly I guess if the radio really shows it as +50MHz, but it would also mean that copying that memory to another radio that does support duplex=off (or even back into this one) it would just come in as a duplex=off memory instead of an offset=+50MHz one...

chirp/drivers/ft4.py
ho0ber: Yeah, apologies - this is just a range. The minimum configurable offset when programming via keypad is displayed as `0.050M` and you can increment in steps of 0.05 up to `99.950M`

kk7ds:
> Wouldn't it be possible to handle this internal to the radio 

Sorry, I meant "internal to the **driver**"

ho0ber:
> I really don't like this, for a number of reasons, but I think you probably expected that possibility, from the verbiage in the bug (thanks for that btw).

I absolutely anticipated this, because I also don't like it for a number of reasons. But I also don't like the idea of maintaining my own fork of Chirp to be able to program frequencies with tx-inhibit. Any means of accomplishing that basic functionality is far more important to me than being able to program any configuration via Chirp that is possible via FPP.

> I think that achieving tx-inhibit in this way is pretty strange and unfortunate if the radio really doesn't have that ability. Also, the driver already claims to support duplex=off (it's not just a baofeng thing). Are you perhaps just suggesting that the only way to achieve duplex=off on the radio is by setting this invalid offset or something?

Unfortunately, setting `duplex=off` in chirp and uploading to the radio does not inhibit tx. The state of 'off' is supported in the driver and the radio can store it (or at least, it comes back when you download from the radio), but it seems to be indistinguishable in function from setting duplex to blank.

So the only means to accomplish TX inhibit on a per-channel basis, either when programming from Chirp or when programming on the radio itself, is to set a duplex offset that will bring the transmit frequency outside of the radio's band restrictions (regardless of whether or not transmit range has been broadened).

> Wouldn't it be possible to handle this internal to the radio and basically say if any memory has a duplex with an offset that is outside the valid range, we consider that to be duplex=off for chirp's purposes? It wouldn't match the radio's behavior exactly I guess if the radio really shows it as +50MHz, but it would also mean that copying that memory to another radio that does support duplex=off (or even back into this one) it would just come in as a duplex=off memory instead of an offset=+50MHz one...

I was thinking about this - the point of improving compatibility with copying to other radios is compelling. The only concern I have is that if we wanted this to work, we'd need to have Chirp be smart enough to interpret `duplex=off` to mean "find an offset sufficiently large so that the transmit frequency is out of band", meaning that we'd also need to track not only the `valid_bands` but the actual unlocked valid bands range in order to ensure a tx frequency is out of band regardless and will error instead of transmit on PTT.

Alternatively, and here's where my inexperience with the inner workings of these radios comes in, it's possible that there is simply a bug in the driver and it would be possible to "fix" the duplex=off functionality so that it inhibits tx.

Thanks for taking the time to consider this - it's unfortunate that there are no obvious "right" answers here. It's a lovely little budget radio, but like all budget radios, it has a number of compromises.

I'll do a bit more testing with the official programming software to see if there's anything I'm missing that can inhibit tx

kk7ds:
> I was thinking about this - the point of improving compatibility with copying to other radios is compelling. The only concern I have is that if we wanted this to work, we'd need to have the driver be smart enough to interpret `duplex=off` to mean "find an offset sufficiently large so that the transmit frequency is out of band", meaning that we'd also need to track not only the `valid_bands` but the actual unlocked valid bands range in order to ensure a tx frequency is out of band regardless and will error instead of transmit on PTT.

Yes, this is what I was suggesting: handling it fully internal to the driver. It sounds like the logic you'd need is exactly what the user has to do on the radio anyway, so I imagine it won't be too hard. If the main frequency is VHF, -99MHz should get you out of band, and if you're in UHF, +99MHz should do it, no?

> Alternatively, and here's where my inexperience with the inner workings of these radios comes in, it's possible that there is simply a bug in the driver and it would be possible to "fix" the duplex=off functionality so that it inhibits tx.

I think it's very likely/possible that the driver is doing something wrong. Or maybe one of the other radios the ft4 driver supports honors the duplex=off properly but not all? What about just setting duplex=split with a TX frequency of 1MHz or something?

> Thanks for taking the time to consider this - it's unfortunate that there are no obvious "right" answers here. It's a lovely little budget radio, but like all budget radios, it has a number of compromises.
> 
> I'll do a bit more testing with the official programming software to see if there's anything I'm missing that can inhibit tx

I certainly support your desire to not carry a fork, so I'm glad you're taking that approach. But, I think that this is a very odd quirk and I'd really like to try to keep the fix/mitigation for this internal to the driver if at all possible. Anything in the core that we carry for the benefit of one driver is significant debt...

ho0ber:
> It sounds like the logic you'd need is exactly what the user has to do on the radio anyway, so I imagine it won't be too hard. If the main frequency is VHF, -99MHz should get you out of band, and if you're in UHF, +99MHz should do it, no?

Yeah, I'm pretty sure it's that simple. After a second look, I realize now that `valid_bands` already represents the "radio can't transmit outside of this even when unlocked" range that I'd need to do an extra paranoid check to make sure it's possible to reach outside a band edge with a 99 MHz offset (just in case the FT-76R comes out, inherits this code, but has a wider range for some reason).

> I certainly support your desire to not carry a fork, so I'm glad you're taking that approach. But, I think that this is a very odd quirk and I'd really like to try to keep the fix/mitigation for this internal to the driver if at all possible. Anything in the core that we carry for the benefit of one driver is significant debt...

That's a very good point. Outside of quirks like this, there's no good reason for a TX frequency to be fully outside the capabilities of a radio, and thus no reason for a user to program such a thing. Jumping through hoops to modify global validation logic for one weird quirk feels wrong when you frame it like this. The ugly code should live in the driver for the ugly radio.

I'll get to work on containing this in the driver (which should also solve the far-reaching bugs I created, causing the test failures in this branch). Thanks for steering me straight!

kk7ds:
> Yeah, I'm pretty sure it's that simple. After a second look, I realize now that `valid_bands` already represents the "radio can't transmit outside of this even when unlocked" range 

That's not entirely true - it's intended to be the range of frequencies that are resolvable (at all) by the VFO in the radio. So typical VHF ham radios will have 136-174 there...i.e. "can dial this frequency" not necessarily "can tx this frequency". Some (but far from all) radios will coerce out-of-tx-band channels to duplex=off to make it clear that they can't transmit there, but there are some other gotchas there.

BUT, I think your point is that you can just check for the same situation that is currently causing you trouble and use *that* to decide if you should fake out duplex=off. Basically, if an offset lands you outside of our own valid_bands for this radio, chirp will complain, so you might as well set duplex=off. So yes, that seems sane to me.

> I'll get to work on containing this in the driver (which should also solve the far-reaching bugs I created, causing the test failures in this branch). Thanks for steering me straight!

Cool, thanks for working on this and glad to have someone else around that can work on this. There have been some bug reports in the past and scaring up someone who can work on it is hard sometimes.

ho0ber:
I think I'm relatively happy with where I got this to (assuming driver tests pass). Any and all feedback would be most welcome.

I opted to make this behavior configurable per radio subclass within the driver so I'm not accidentally changing behaviors for other radios where duplex=off works properly, or where it doesn't work but users might be relying on current behavior.

I would have enabled the conversion of out-of-bad offsets for all radios in the family, but I wouldn't want someone to load up their favorite image, have it convert a bunch of out-of-band offsets to duplex=off, and then unknowingly upload them to the radio with transmit enabled.

I also opted to make the offset value itself configurable, just in case a new radio enters the family with this issue but where the -/+ split rule needs to change or where the maximum offset is not the same as the FT-65. It also would encourage anyone configuring this feature for a radio to evaluate what safe direction and offsets would be.

I am much happier with this than the initial solution I proposed.

ho0ber:
Okay, I'm finally _actually_ happy with the code and tests are passing. Let me know how it looks!

tests/base.py
```diff
@@ -188,6 +188,27 @@ def assertEqualMem(self, a, b, ignore=None):
...
+                if matching_band is not None and \
+                   self.radio.DUPLEX_OFF_VIA_OFFSET[i] is not None:
+                    duplex, offset = self.radio.DUPLEX_OFF_VIA_OFFSET[i]
+                    if a.offset == offset and a.duplex == duplex and \
+                       b.duplex == "off":
+                        continue
+                    if b.offset == offset and b.duplex == duplex and \
+                       a.duplex == "off":
+                        continue
+
```
kk7ds, line 211: First, major props for adding test coverage. *Nobody* helps me with test coverage :)

However, I think maybe we should put this in a different spot. You're checking for your flag, but that flag isn't on the base radio type, it's really an internal thing to just your driver. Like, if someone else just happened to implement that same flag in a way that means something very different, this might be triggered in a good or bad way. This is a pretty deep spot to embed that logic (even though I recognize it's not too dissimilar from some other things here).

So I think we should still do this check, but let's put it in the unit tests, where we have some more targeted tests for specific radios and situations, if that's okay. Here's an example of a radio-specific unit test file for some of the cloning routines in one driver:

https://github.com/kk7ds/chirp/blob/master/tests/unit/test_th9800.py

I think that sort of thing is best for here, that way you can write it targeted for just the ft4 and not add these assertions that could potentially affect everything. If so inclined, that would also give you a spot to add further test coverage for all the ft4 weirdness if you so chose :)

I'll be glad to move/rewrite this in that form for you if you like, happy to do that, but don't want to take and mangle your code if you'd rather.

ho0ber: In this case, I had to modify `assertEqualMem` in order for most of the tests associated with the radio to pass, as many of the standard tests now trigger the special logic and fail the assertion because it's no longer a given that a `+` or `-` duplex will stay that way, it may instead become an `off`. Offsets as well would change in some circumstances. So the goal of this wasn't to be a unit test so much as it was to address the fact that this driver-specific logic affects some broad assertions - namely that for the FT-65R, out-of-bound offsets with a +/- duplex should be treated as equivalent to `duplex=off` when comparing memories.

If there's a good way to ensure that the ft4 driver tests have an overridden `assertEqualMem`, then I agree that this would belong somewhere driver-specific. Beyond that, I guess we could also branch on driver name for this special logic, so at least we don't run the risk of someone else using the same constant name for other purpose in a different driver (and then it would also be exceptionally clear that this quirk is driver-specific).

There's probably a clever way to inhibit those assertion fails in `assertEqualMem` and fill in equivalent driver-specific testing, but I wasn't coming up with anything when I tried.

kk7ds: Ohh, I see, I read this too fast during dinner. You're bailing in this scenario and not running stuff below, I see, I was thinking this was just checking those limits, but now I see what's going on.

The problem with that is that you're basically eliminating all the test assertions for radios with that property, which is, uh, cheating :)

Let me think on this a bit tomorrow and see what I can come up with. In general, we should not be modifying the tests, but making the driver do the right thing of course.

kk7ds: Okay, so, with this test change reverted, I see failing one test:
```
FAILED tests/test_drivers.py::TestCaseEdges_Yaesu_FT-65R::test_get_set_specials - AssertionError: {'fre[67 chars]x': 'off'...
```

And it looks like a legit fail to me. The test tried to set a `duplex=off` memory, but got back a `duplex=+` memory. That's because while you converted `duplex=off` to "big offset" in `set_memory()`, your `get_memory()` didn't turn that back into `duplex=off` on the way back out. So that's a bug in your driver, as get/set should be identical in this case.

We can make `assertEqualMem()` ignore the offset discrepancy if `duplex=off`, but that's not the actual problem here that I see.

chirp/drivers/ft4.py
```diff
@@ -1149,12 +1150,35 @@ def get_duplex(freq):    # auto duplex to real duplex
                 mem.offset = txfreq
             else:
                 mem.offset = int(_mem.offset) * 25000 * freq_offset_factor
+            if self.DUPLEX_OFF_VIA_OFFSET and \
+               regtype in ["memory", "pms"] and mem.duplex in ['+', '-']:
```
kk7ds, line 1154: This is the source of the only error I see. The `regtype in ["memory"...]` bit excludes the special memories from this big-offset handling, which means the special memory test fails to be able to set `duplex=off`, because it just sees `off` turn into `+99MHz` as expected. Why this filter? If I remove the regtype check, all the tests pass for me.

kk7ds:
I think this should pass. I just pushed a few commits on top of yours, unchanged, so it's easy to revert back if I'm missing something about why we need that `regtype` filter.

ho0ber:
> The problem with that is that you're basically eliminating all the test assertions for radios with that property, which is, uh, cheating :)

The intent was that this was only eliminating the comparison for the `duplex` and `offset` fields when comparing memories for radios with that property, and only doing so if they mismatched in exactly the intended way. I don't think that would have been "cheating" just performing comparison for those fields ahead of the assert statement.

> I think this should pass. I just pushed a few commits on top of yours, unchanged, so it's easy to revert back if I'm missing something about why we need that regtype filter.

Nope, I think I was just looking at the code too long and chasing at ghosts. I think I was convinced that the special memories were causing issues with the special duplex/offset conversion logic, but in hindsight, I'm not sure why. Another great victory for PR review!

Thanks for cleaning it up and keeping me honest! I'm eagerly awaiting test completion.

kk7ds:
> Thanks for cleaning it up and keeping me honest! I'm eagerly awaiting test completion.

Await no further 🎉 .. So, I will squash these into fewer pieces, just the test and driver fix. I also want to make a few style changes that I hope you won't mind. In most of chirp, we try to do line breaks with parens instead of backslashes to the point that I had to manually force my editor to match your style while making this fix.. so if you don't mind, I'll fix that up at the same time. Please accept apologies from my OCD :)

Also, since we just had a build, I'll hold off on merging this for a day or so until we have a little more meat for the next build, but otherwise should be fine.

Thanks!

kk7ds:
Oh and add the bug reference in the commit so it links properly.

ho0ber:
> so if you don't mind, I'll fix that up at the same time

Not at all! Style consistency is important. I've got no ego about my contribution, so feel free to shape it to match the codebase as much as you see fit.

> Also, since we just had a build, I'll hold off on merging this for a day or so until we have a little more meat for the next build, but otherwise should be fine.

For sure - can't be popping up the update notification on folks _every_ day.

Once we get this out the door, and now that I'm a bit more onboarded to the project, I'm going to take a crack at picking up where others left off on Radioddity GM-30 support. I'm sure I'll continue to make some mistakes along the way, so your continued guidance will be much appreciated.

kk7ds:
Cool, I'm sure GM-30 people will appreciate that :)

PR #999, opened 2024-03-20
tdh8: Fix variants with two power levels (H3)

kk7ds:
Fixes: #11257


songjunbiao:
Did you get the radio we sent you? 
If you have any suggestions on how to use it, we can discuss them.
Thank you for your support.

PR #1003, opened 2024-03-31
radioreference: fix utf-8 encoding issue with SUDS

trevorstarick:
# CHIRP PR Checklist

* All tests must be passing.

---

## Problem:
Current importer fails to handle "invalid tokens" which I believe are most likely caused by non-ascii accents frequently found in the Quebec region:

image

## Solution:
The PR attempts to fix that by adding a MessagePlugin to decode the input (while ignoring errors) and then re-encode it as utf-8. 

I tested it locally and was able to import the QC/Montreal region with no "invalid token" errors.

### DIFF
--- a/chirp/sources/radioreference.py
+++ b/chirp/sources/radioreference.py
@@ -26,6 +26,7 @@ CONF = config.get()
 try:
     from suds.client import Client
     from suds import WebFault
+    from suds.plugin import MessagePlugin
     HAVE_SUDS = True
 except ImportError:
     HAVE_SUDS = False
@@ -85,8 +86,14 @@ class RadioReferenceRadio(base.NetworkResultRadio):
                 "Suds library required for RadioReference.com import.\n" +
                 "Try installing your distribution's python-suds package.")
 
+        class UnicodeFilter(MessagePlugin):
+            def received(self, context):
+                decoded = context.reply.decode('latin1')
+                reencoded = decoded.encode('utf-8')
+                context.reply = reencoded
+
         self._auth = {"appKey": self.APPKEY, "username": "", "password": ""}
-        self._client = Client(self.URL)
+        self._client = Client(self.URL, plugins=[UnicodeFilter()])
         self._freqs = []
         self._modes = None
         self._zipcounty = None


chirp/sources/radioreference.py
```diff
@@ -24,6 +24,14 @@
 CONF = config.get()
 
 try:
+    from suds.plugin import MessagePlugin
+
+    class UnicodeFilter(MessagePlugin):
+        def received(self, context):
+            decoded = context.reply.decode('utf-8', errors='ignore')
```
kk7ds, line 31: Is `errors='ignore'` required for your test case or were you just trying to be extra cautious? If the latter, I think it would be better to just ignore results that have broken encoding (which really shouldn't happen anyway).

Also, could you move this down outside the import section? This defensive import is also from a really long time ago, so I should probably clean all this up anyway.

Thanks!

trevorstarick: The `errors='ignore'` can be replaced by `errors='replace'` but then you get weird characters within chirp and I can imagine this leading to other issues down the line: 

image

trevorstarick: I've moved the UnicodeFilter outside of the block to just above where the Client is created and left the import along with the other suds imports.

kk7ds: Hmm, the only thing that could mean is that the service is not using unicode I think. Can you tell me what query you're doing so I can try to reproduce?

trevorstarick: I'm using this query: 

image

When I logged which subcats failed, it pointed me towards this https://www.radioreference.com/db/browse/ctid/5210/ham. While it should be valid unicode, however it was inputted on the RadioRefrence side, it became invalid (it's supposed to be `Collège` instead of `Coll�ge`)

trevorstarick: There _should_ be two failed parsings, one in `county.cats` and another in `county.agencyList`.

kk7ds: Okay, yep, the problem is that it's not `utf-8`, it's `latin1`. If you change that input decoding to `latin1` (and remove the `errors=` hack) you will get no error *and* you'll get the proper string in chirp.

trevorstarick: Perfect, I'll make those changes and update the PR

trevorstarick: Regarding the encoding, I couldn't find any official source on what the API should be returning but according to [this](https://forums.radioreference.com/threads/utf-8-utf-16-encoding-support.300758/) forum post from 2014, it looks like it was originally `latin1` and then `utf-8` was "patched in".

kk7ds:
Cool, thanks. I'm going to squash this (per bullet #2) and merge.

PR #1008, opened 2024-04-04
Add Alinco DR-735T Driver

jacobcalvert:
# CHIRP PR Checklist
[Issue Link](https://chirpmyradio.com/issues/8885)

Fixes #8885 in CHIRP issue tracker. 

* All tests must be passing.


chirp/drivers/alinco_dr735t.py
```diff
@@ -0,0 +1,368 @@
...
+            _mem.mode = find_key_in(self.MODE_MAP, mem.mode)
+            _mem.subtone_selection = find_key_in(self.TONE_MODE_MAP, mem.tmode)
+            _mem.shift = mem.offset
+            _mem.used = 0x00 if mem.empty else 0x55
+            _mem.power_index = self.POWER_MAP.index(
+                mem.power) if mem.power in self.POWER_MAP else 0
+            _mem.skip = 0x01 if mem.skip == "S" else 0x00
+            try:
+                _mem.rx_tone_index = ALINCO_TONES.index(
+                    mem.rtone if mem.rtone else 67.0)
```
kk7ds, line 264: Why are you doing `if mem.rtone else`? There should be no case where `mem.rtone` is falsey.

chirp/drivers/alinco_dr735t.py
```diff
@@ -0,0 +1,368 @@
...
+                                                  "not support " +
+                                                  "tone %.1fHz" % mem.rtone)
+            try:
+
+                _mem.tx_tone_index = ALINCO_TONES.index(
+                    mem.ctone if mem.ctone else 67.0)
+            except ValueError:
+                raise errors.UnsupportedToneError("This radio does " +
+                                                  "not support " +
+                                                  "tone %.1fHz" % mem.ctone)
```
kk7ds, line 276: Not a big deal, but you don't need the `+` to concat here, adjacent strings are automatically concatenated.

chirp/drivers/alinco_dr735t.py
jacobcalvert: Hi Dan, thanks for looking at my PR.  One of the driver tests was failing on mem.rtone being None in some cases it seemed, although I couldn't really tell given the test output was a little hard to understand since I'm new to the project. I can remove the falsesy check since the exception should catch any wrong value (I added the try later). Happy to change that.

kk7ds: Okay, yeah, I'm not sure when/where that would be but I'd rather not have the implication here and handle any failure to find the tone values with the exception handler. Thanks!

kk7ds:
Please squash all these into a single commit since they're all just iterative attempts to get to the working driver (per the PR guidelines above). Also, please put your `Fixes: #8885` in that commit message when you do so the thing gets referenced properly. The github PR metadata doesn't ultimately make it into the repo the issue system sees.

I'll merge this (and my own pending patch) in the next few days, I just spread these out a bit to keep the builds from being too frequent.

PR #1015, opened 2024-04-14
Update Italian Translation

giovanni69:
# CHIRP PR Guidelines


giovanni69:
Update Italian Translation

PR #1018, opened 2024-04-16
New driver for Yaesu FTM-6000

Old-Phart:
Supports FTM-6000 and includes FTM-200; Issues 9719 & 11137


Old-Phart:
Oops. Way too many conflicts. Returning to VS to clean them up.

PR #1031, opened 2024-05-11
Fixup rpttool

cmlara:
Fix idrp python3 conversion

Fixes #11252

Co-Authored-By: Dan Smith <dsmith@danplanet.com>

### DIFF
--- a/chirp/drivers/idrp.py
+++ b/chirp/drivers/idrp.py
@@ -81,7 +81,7 @@ def set_freq(pipe, freq):
     resp = send(pipe, buf)
     for frame in resp:
         if len(frame) == 6:
-            if frame[4] == b"\xfb":
+            if frame[4] == 251:
                 return True
 
     raise errors.InvalidDataError("Repeater reported error")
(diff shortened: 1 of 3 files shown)


chirp/drivers/idrp.py
```diff
@@ -81,7 +81,7 @@ def set_freq(pipe, freq):
     resp = send(pipe, buf)
     for frame in resp:
         if len(frame) == 6:
-            if frame[4] == b"\xfb":
+            if frame[4] == 251:
```
kk7ds, line 84: Ah yep, this was incorrect py2->py3 conversion. I figured it was something like this.

kk7ds (commented):
The cpep8 manifest complaint looks incorrect. The rest of this looks good to me. Still draft because you're still working on it, or what? If you respin, please put "Fixes #11252" in one (or all) of the commit messages so that it gets tagged to the bug. It has to be that pattern or the bug tracker won't notice. I'll try to remember to do it myself before I merge if this is ready.

Either way we just had a build today so I'll hold off merging for a few days.

Thanks!

kk7ds:
See if that works or gets you close. That function is obsolete for everything else other than idrp, so it never got fixed for py3.

kk7ds:
Ah, PR check fail was because it wasn't rebased on master and `cpep8.manifest` was trimmed between what you were based on and current.

cmlara:
I have to attend a family event, however I was able to run a quick test with your latest change and rpttool reports successfully changing the frequency.

I can return this evening to cleanup the commit messages and do any other fix-up work you might want.

kk7ds:
Cool, glad to hear it. I think we should just squash this into one "fix idrp" commit with the proper bug reference and I'll get it merged sometime next week. So, no rush and/or I can just do it for you.

Thanks much!

cmlara:
Cleaned up the commit message and squashed.  Thanks for the help with the bcd_encode portion to finish this up.

No rush on the merge, take it at your leisure.

PR #1034, opened 2024-05-16
Allow sorting columns in the memedit view

kk7ds:
- **Simplify row label iteration**
- **Support sorting columns in the view**
- **Allow unsorting by clicking the header corner**

### DIFF
--- a/chirp/locale/de.po
+++ b/chirp/locale/de.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-05-01 19:23-0700\n"
+"POT-Creation-Date: 2024-05-16 15:11-0700\n"
 "PO-Revision-Date: 2012-10-02 22:11+0100\n"
 "Last-Translator: Benjamin, HB9EUK <hb9euk@hb9d.org>\n"
 "Language-Team: German\n"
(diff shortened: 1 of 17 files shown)


kk7ds:
@KC9HI You're good at breaking stuff. Can you take a poke at this? No rush. Things to look for are failures during sorting, and/or the wrong memory showing up in a slot, edits being made against the wrong memory, etc.

Note that if you edit a field in a sorted column, it doesn't re-sort the memories after the edit until/unless you re-sort the column. I think it would be annoying to have memories jump around when you edit them, but let me know what you think.

kk7ds:
This is different from the sorting in the right-click menu. This is just sorting the "view" like legacy chirp could do. Many people have asked for this in a few different venues. Here are a couple bugs about it specifically:

https://www.chirpmyradio.com/issues/11338
https://www.chirpmyradio.com/issues/10750

Personally, I don't really ever care to sort my memories by anything other than location, but some people seem to want to do it. I guess if you're looking for a particular channel, you might want to sort by frequency and then go look for it in the place it should be, or something. Unfortunately it adds a lot of complexity, and as such, I'll have to disable a number of things *while* the view is sorted (such as import, the "other" sorting, clustering, etc).

After I pushed this last night and slept on it, I realized a few more bugs with it, so I just pushed some fixes for those. Hopefully that makes this work consistently now.

So the goal here is that if you (for example) sort by frequency and make a change to memory 7, then sort by the default (location) the changes you made to memory 7 while it was in some other order still look consistently applied to memory number 7 :) Copy and paste and edits like you were doing there are exactly the sort of thing I'm looking for, bug-wise.

KC9HI:
Ahhh... Now I think I understand. Just a new/different method of finding a memory to edit, then once edited, return back the natural 'location' order. Now it makes sense.

I will grab it and take another look after supper (I should be getting the call soon).

kk7ds:
Yeah, I guess that's one reason for it. Looking at one of the other bugs, another example might be searching for duplicates, without changing the actual ordering in the radio.

But yeah, this is why I'm trying to call this "sorting the view" instead of "sorting the memories". The former just changes the order in which you see stuff, the latter changes the ... actual order.

No rush, enjoy supper, thanks!

KC9HI:
OK. I think it is working. I've had a couple of crashes where CHIRP closes. Unfortunately I didn't have logging enabled. It is always when I am trying to select a field for editing. I do see the following in the log but I don't know if it is related...

DEBUG: Ignoring drag from row 23 -> 23

kk7ds:
Hmm, okay I haven't done much testing on windows, so I better try to reproduce that ...

kk7ds:
I wasn't able to reproduce any funny behavior or crashes on Windows. Have you seen more of this?

KC9HI:
After I submitted a PR, I tried again last night. I was not able to recreate the crash. I meant to try again today but haven't gotten around to it yet.

Also, when I had the issue, it was the previous version. And I was trying to use it in a way that was not intended.

kk7ds:
Ah, good to know. The thing that made the previous version mess up the numbering could have been pushing things outside the grid, which can cause serious problems, so hopefully that was it.

Thanks!

PR #1035, opened 2024-05-17
Add Retevis RA87 - fixes #10871

KC9HI:
# CHIRP PR Guidelines


chirp/drivers/retevis_ra87.py
```diff
@@ -0,0 +1,1250 @@
...
+            _mem.duplex = DUPLEX_POSSPLIT
+            _mem.offset = mem.offset / 1000
+        elif mem.duplex == '-':
+            _mem.duplex = DUPLEX_NEGSPLIT
+            _mem.offset = mem.offset / 1000
+        elif mem.duplex == '':
+            _mem.duplex = DUPLEX_NOSPLIT
+        else:
+            LOG.error('%s: set_mem: unhandled duplex: %s' %
+                      (mem.name, mem.duplex))
```
kk7ds, line 683: The UI should not send you any duplexes other than those your radio claims to handle... Is this for a specific behavior you saw, or are you just being comprehensive? If it can happen for some reason, this should probably raise an error so there's some indication to the user that something went wrong, right? Otherwise it'll just flip back when it refreshes.

chirp/drivers/retevis_ra87.py
```diff
@@ -0,0 +1,1250 @@
...
+        #
+        options = ["TX Inhibit", "TX/RX Inhibit"]
+        rs = RadioSettingValueList(options, options[_dtmfd.stuntype])
+        rset = RadioSetting("dtmfd.stuntype", "Stun Type", rs)
+        dtmf_dec.append(rset)
+
+        # DTMF - Autodial Memory
+        codes = self._memobj.dtmf_codes
+        i = 1
+        for dtmfcode in codes:
```
kk7ds, line 1162: Protip life hack for you. Whenever you need to iterate a list *and* have the index, you can do this:
```
for i, dtmfcode in enumerate(codes):
```
Now, `i` will be zero-origin isntead of 1 like you're doing here, but  it looks like you're using it as `i-1` anyway.

chirp/drivers/retevis_ra87.py
KC9HI: This is basically code borrowed from the anytone778uv.py driver. DUPLEX_NOSPLIT is just a variable that holds the value of 0x00 which is the value of 'duplex' when the radio's memory is set to simplex (no split).

    u8   step:4,          // STEP                                    // 5
         unk1:2,
         duplex:2;        // Duplex 0: Simplex, 1: Plus, 2: Minus

Do you just want me to rename the variable to something like DUPLEX_SIMPLEX, replace it directly with 0x00 or am I not understanding what you are trying to tell me?

chirp/drivers/retevis_ra87.py
KC9HI: Thanks for the tip. Appreciated.

chirp/drivers/retevis_ra87.py
kk7ds: Nope, I'm just wondering about the else condition for `mem.duplex` and the error log. I'm just wondering if you added that because you saw the UI send you a duplex that wasn't in the radio's supported_duplexes?

KC9HI: No. It it there only because it was in the anytone778uv.py driver that I borrowed it from.

kk7ds: Okay, gotcha, just wanted to make sure it wasn't signaling something else going on.

KC9HI: There is a similar else condition in both get_memory() and set_memory(). Should I just remove them both?

kk7ds: No need to remove them, I just wanted to make sure there wasn't some reason for it :)

PR #1041, opened 2024-05-21
Update Spanish Translation

MELERIX:


ajraymond:
@MELERIX I added a mechanism to translate the "Welcome to CHIRP" image shown when starting the software. Since it's now merged, you might want to translate it to Spanish, like I did for French. See https://github.com/kk7ds/chirp/pull/1043/files

ajraymond:
Yes, in your case, welcome_screen_es.xcf -> welcome_screen_es.png.

I don't know if these are the exact parameters needed, but these are the export options I used:
image

kk7ds:
There's a `.xcf` file (GIMP format) in the tree you can use to clone/edit the original instead of just creating something new: 

https://github.com/kk7ds/chirp/blob/master/chirp/share/welcome_screen.xcf

I definitely prefer that for consistency :)

PR #1046, opened 2024-05-23
Make 'reveal backup location' create folder if it doesn't exist

ajraymond:
If you run "Help/Show image backup location" before having made a backup, the action fails silently (with an error on the console), because the backup folder doesn't exist.

This MR includes a fix that attempts to create the directory before "revealing" it.


kk7ds:
I might argue it would be better to just tell the user "no backups yet" instead of show them an empty directory, but this is obviously better than doing nothing ;)

Thanks!

PR #1047, opened 2024-05-24
ADDED: Welcome Screen in Spanish

MELERIX:


kk7ds:
I should have known this would be only a few minutes of turnaround time for you :)

PR #1052, opened 2024-05-27
Limit Repeaterbook Queries to Open Repeaters

alrs:
This adds a checkbox to the Repeaterbook query window to limit results to open repeaters only.

Is this small enough to be a pull request only, or should I create an issue?


kk7ds:
Yeah, it's okay. I can fix up the style, locale, and test issues before I merge this, so don't worry about them. Thanks!

PR #1053, opened 2024-05-28
Fix python suds dep

ajraymond:
Following discussions in https://github.com/kk7ds/chirp/pull/1050, this MR makes the python3-suds dependency non-optional, and updates the instructions found in the INSTALL file, pointing to the Wiki directly.


kk7ds:
Thanks 👍

PR #1056, opened 2024-06-03
Alias Baofeng UV-25 to UV-17Pro driver

ajraymond:
Word on the street seems to be that the Baofeng UV-25 is compatible with the UV-17Pro driver (ex: https://www.youtube.com/watch?v=FJStvAKlx68).

This should address https://chirpmyradio.com/issues/11260


kk7ds:
The magic incantation for closing issues in the commit message is `Fixes: #11260`. Also, please squash "fix tests" commits. No need to keep version history of false starts forever :)

PR #1058, opened 2024-06-06
Fix radtel rt730 bugs

KC9HI:
# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -1342,7 +1342,7 @@ class TDH8(chirp_common.CloneModeRadio):
             try:
                 _nam.name[i] = mem.name[i]
             except IndexError:
-                _nam.name[i] = "\xFF"
+                _nam.name[i] = "\x00"
 
         txtone, rxtone = chirp_common.split_tone_encode(mem)
 
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -1005,6 +1005,7 @@ class TDH8(chirp_common.CloneModeRadio):
     ident_mode = b'P31183\xff\xff'
     BAUD_RATE = 38400
     NEEDS_COMPAT_SERIAL = False
+    MODES = ["FM", "NFM"]
     _memsize = 0x1eef
     _ranges_main = [(0x0000, 0x1eef)]
     _idents = [TD_H8]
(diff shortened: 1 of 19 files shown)


KC9HI:
@kk7ds Hi Dan. I am having trouble with the 3rd commit. I am trying to add the ability to paste a memory row with a forced AM frequency into a memory row with an existing forced AM frequency based on how it is done in the baofeng_uv17Pro.py driver.

What I have doesn't pass the unit tests. I don't get what the errors are trying to tell me. If I comment out the 'elif: section in

check_set_memory_immutable_policy()

then it passes all tests.

So whenever it is convenient for you, I would be grateful if you could take a look at this and help me fix it.

Thanks,
Jim

chirp/drivers/tdh8.py
```diff
@@ -1024,6 +1025,22 @@ class TDH8(chirp_common.CloneModeRadio):
                  chirp_common.PowerLevel("Mid",  watts=4.00),
                  chirp_common.PowerLevel("High", watts=8.00)]
 
+    def check_set_memory_immutable_policy(self, existing, new):
```
kk7ds, line 1028: Does this need to be on the base class or can it be only on the RT730 subclass? The failure is because you didn't add it to the whitelist in `test_chirp_common` but if we add it on the base class here, we have to add *all* the subclasses to the whitelist as well, which would be less than ideal, unless it's really required.

KC9HI: Actually it should probably be in the RT730 and any other model that supports airband (the TD-H3). I actually had put them there but switched it back to where it is now since I was unable to get it to work.

kk7ds: Okay, so it doesn't work for some reason when put on the RT730 itself? It really should...

KC9HI: I guess not if it isn't in the whitelist. I will add it to the whitelist. Should I just do it for the RT730 and don't worry about the TD-H3 unless someone complains about it?

kk7ds: I dunno, maybe just do it for the RT730 now so we can get this merged. So you're going to move this to the RT730 subclass and add just that to the whitelist right?

Not sure if it's unclear, but the whitelist only affects the test passing, not the runtime operation...

KC9HI: Moving to RT730 and adding to whitelist passes the tests.

kk7ds:
The driver test was failing before, along with the unit test, so I think something else must be going on. Lemme have a look.

KC9HI:
OK. Thanks.

kk7ds:
So, the test it's failing is actually trying to set a memory that is outside the range supported by the radio, and it's failing on the immutable thing before it has even checked that the frequency is valid for the radio. The error is actually in import_logic and I'm having a hard time understanding how that went unnoticed for so long...

KC9HI:
OK. Just leave a comment letting me know what I need to do, if anything, and I will give it a go tomorrow.

PR #1072, opened 2024-06-22
test radio driver only

KC9HI:
# CHIRP PR Guidelines


KC9HI:
@kk7ds When it is convenient, would you take a look at this to see if you can help me figure out what is wrong. It won't pass the 'drivers' test. It appears to have problems with the Baofeng UV-17 and Baofeng UV-13Pro radios that I don't think are even a part of the driver that I am editing.

Thanks,
Jim

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -521,6 +575,35 @@ def get_settings_pro_dtmf(self, dtmfe, _mem):
                                                     val]))
         dtmfe.append(rs)
 
+    def get_settings_f8hp_pro_dtmf(self, dtmfe, _mem):
```
kk7ds, line 578: This should be defined on the F8HP class only since it's only used there.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -968,6 +1112,8 @@ def get_settings(self):
         dtmfe = RadioSettingGroup("dtmfe", "DTMF Encode Settings")
         self.get_settings_common_dtmf(dtmfe, _mem)
         self.get_settings_pro_dtmf(dtmfe, _mem)
+        if self.MODEL == "BF-F8HP_Pro":
+            self.get_settings_f8hp_pro_dtmf(dtmfe, _mem)
```
kk7ds, line 1116: It would be much better to just override `get_settings_f8hp_pro_dtmf()` in the F8HP subclass and add the extra stuff there. I know this is the pattern you're used to, but in this case, this stuff is starting to get a bit messy as a result, so it'd be better to move that there. Using `super()` will let you call the parent implementation and then add on more so it's all localized down there.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -775,6 +862,58 @@ def apply_Key2short(setting, obj):
                           RadioSettingValueBoolean(_mem.settings.fmenable))
         basic.append(rs)
 
+    def get_settings_f8hp_pro_basic(self, basic, _mem):
```
kk7ds, line 865: This too.

kk7ds (commented):
The problem stems from the fact that the UV17 driver (and all its descendants) inherit from the UV17Pro. That's why the errors show that the UV17 models are failing, and are running your new pro code changes (around the parameterized `MEM_FORMAT`. 

I'll push a fix for those issues on top in a sec. I'd normally like that to be in the patch itself, but I'll keep it separate in order to make it clear what I've changed, so please do look at that. Also, I'm going to add another commit on top to clean up the extra settings methods you're adding here and move them to the F8HP subclass. If you were to look at a python model browser, you'd see that all the subclasses of UV17Pro, UV17, etc all get those f8hp methods, even though they're not really used. That can be quite confusing to people, and especially code inspection tools. The right way to do that is to override the method in the subclass, call the parent's version first (which is what `super()` does) and then add on the extra stuff you want. Hope that makes sense when you see it, but if not, let me know and I'll help clarify.

I've been working on the GA510v2 driver, which is based on the UV17 one, so the lineage from the UV17Pro driver is getting really deep at this point and thus we really need to be careful about cleanliness here. A lot of what you've changed here breaks my patches (not your fault, you haven't seen them yet) but I'll try to get those rebased on top of this and hopefully it won't be too bad.

So I'll push those revisions in a sec, please confirm they look okay to you before we merge this.

Thanks!

kk7ds:
This pretty much rebased on current master without any trouble. It's been a while since I made the changes so I'm not sure if that's surprising or not. I know I had done some work on this already. If you can test it and report what doesn't work I can try to address it.

KC9HI:
@kk7ds Sorry for the delay. I finally got a chance to test. All that I can see that I needed to change was the 'if self.MODEL == "BF-F8HP_Pro"' to use "BF-F8HP-PRO" and all seems to be working OK.

kk7ds:
Ah, yep, sorry. When I read this a bit ago, I thought you maybe were going to push up the fix, but I can, if you're willing to test again.

And... what are you thinking about time-wise for merging this?

PR #1081, opened 2024-07-09
ksun_m6: Fix initialization of new memories

wavexx:
Explicitly set a few extra bits to zero to initialize correctly new memory slots.

_unk3/_unk4 seem to be part of compander/hopping respectively, even though only a single value for both is accepted.

Setting _unk3 to 1 (compander:0b10) makes the radio ignore the channel completely. The original software also refuses to decode the slot and fail with a generic error.

Setting _unk4 to 1 (hopping:0b10) works on the radio, but still causes the original software to fail with an error.

### DIFF
--- a/chirp/drivers/ksun_m6.py
+++ b/chirp/drivers/ksun_m6.py
@@ -55,10 +55,8 @@ struct {
       skip: 1,
       tx_tone: 2,
       tx_code: 12;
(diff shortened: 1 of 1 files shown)


kk7ds:
Thanks, sorry this took so long to be merged!

PR #1082, opened 2024-07-28
IC-R6 Initial Driver

jbradsha:
Icom IC-R6 Initial version

# CHIRP PR Guidelines


kk7ds:
Your commit is titled something that won't make sense for people when it gets sent to the mailing list. Please change that to something like "Add Icom IC-R6 driver" per item 5 above. Also, please include a test image per item 6.

Thanks!

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+
+# TODO:
+# Channel: Frequency setting
+# Channel: TSQL freq or DTCS code with polarity
+# Channel: Skip scan settings
+# Channel: Comment (not sure if supported in ICF)
```
kk7ds, line 22: This is handled internally by CHIRP (in the metadata). Unless the radio can store it (I doubt it) you should just let CHIRP's base infrastructure do it for you.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+#    RadioSettingValueInteger, \
+
+mem_format = """
+// Channel memories: 1300x 16-byte blocks
+// 0x000 to 0x513F inclusive
+struct {
+  u8 freq0;               // Freq: low byte
+  u8 freq1;               // Freq: mid byte
+  u8 freq_flags:6,        // Related to multiplier (step size)
+     freq2:2;             // Freq: high bits - 18 bits total = step count
```
kk7ds, line 48: I'd really prefer that you use a u24 here instead of all the bit-shifting. I think you can just mask out the flag bits yourself once you grab the value. I'll do some thinking on how I could make this define-able in the bitwise here...

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+  u8 af_filter:1,         // AF Filter: 0=Off, 1=On
+     attenuator:1,        // Attenuator: 0=Off, 1=On
+     mode:2,              // Modulation: Index to "MODES"
+     tuning_step:4;       // Freq tuning steps: Index to "STEPS"
+  u8 unknown4ab:2,
+     duplex:2,            // 0 = None, 1 = Minus, 2 = Plus
+     unknown4e:1,
+     tmode:3;             // TSQL/DTCS Setting: Index to "TONE_MODES"
+  u8 offset_l;            // Offset - low value byte
+  u8 offset_h;            // Offset - high value byte
```
kk7ds, line 58: Can you change this to a u16 please? The point of bitwise is to minimize the assembly and masking of bits in the code as much as possible.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+  u8 unknown7:2,
+     ctone:6;             // TSQL Index. Valid range: 0 to 49
+  u8 unknown8;
+  u8 canceller_freq_h;    // Canceller training freq index - high 8 bits
+  u8 canceller_freq_l:1,  //                               - LSB
```
kk7ds, line 63: Just make this a u16 and then you won't need to combine `freq_h` with `freq_l` in your code.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+     unknown10m:4,
+     vsc:1,               // Voice Squelch Control: 0=Off, 1=On
+     canceller:2;         // USA-only Canceller option: Index to "CANCELLER"
+  u8 name0;
+  u8 name1;
+  u8 name2;
+  u8 name3;
+  u8 name4;
```
kk7ds, line 71: This should be a `u8 name[5]`

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+} memory[1300];
+
+// Unknown: 0x5140 to 0x5F7F inclusive
+struct {
```
kk7ds, line 75: Theres no need to do this sort of thing and it just makes this longer and less readable. Use `#seek nbytes` or `#seekto addr` to skip a section you're not going to handle.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+  u8 unknown03;  // 6bd3
+  u8 unknown04;  // 6bd4
+  u8 unknown05;  // 6bd5
+  u8 unknown06;  // 6bd6
+  u8 unknown07;  // 6bd7
+  u8 unknown08;  // 6bd8
+  u8 unknown09;  // 6bd9
+  u8 unknown10;  // 6bda
+  u8 unknown11;  // 6bdb
+  u8 unknown12;  // 6bdc
```
kk7ds, line 192: It'd be a lot more readable if you'd just change this to `u8 unknown[13]` and that's definitely the convention.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+struct { // Start: 6D00, End: 6D0F
+  char comment[16];
+} device_comment;
+
+// 22x ASCII-coded bank names
+// 0x6D10 to 0x6DBF inclusive
+struct {
+  char name[6];
+  u8 unknown0; // Padding?
+  u8 unknown1; // Padding?
```
kk7ds, line 289: `u8 padding[2]` will save a line in things like this.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+#    "TQQL-R" (Tone Squelch Reverse),
+#    "DTCS-R" (DTCS Reverse)”,
+#    "OFF"
+TONES = (67.0,  69.3,  71.9,  74.4,  77.0,  79.7,  82.5,  85.4,
+         88.5,  91.5,  94.8,  97.4, 100.0, 103.5, 107.2, 110.9,
+         114.8, 118.8, 123.0, 127.3, 131.8, 136.5, 141.3, 146.2,
+         151.4, 156.7, 159.8, 162.2, 165.5, 167.9, 171.3, 173.8,
+         177.3, 179.9, 183.5, 186.2, 189.9, 192.8, 196.6, 199.5,
+         203.5, 206.5, 210.7, 218.1, 225.7, 229.1, 233.6, 241.8,
+         250.3, 254.1)
```
kk7ds, line 323: Is this different than one of the sets in chirp_common? If so, it'd be nice to copy and modify it here so it's easier to tell what the delta is. Like, if it's missing 100.0 or something, then:
```
TONES = list(chirp_common.TONES)
TONES.remove(100.0)
```

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+                                    _mem.freq1 * 256 +
+                                    int(_mem.freq0))
+            mem.offset = 8333.3333 * ((_mem.offset_h * 256) + _mem.offset_l)
+        elif _mem.freq_flags == 60:
+            mem.freq = 9000 * (_mem.freq2 * 256 * 256 +
+                               _mem.freq1 * 256 +
+                               int(_mem.freq0))
+            mem.offset = 9000 * ((_mem.offset_h * 256) + _mem.offset_l)
+        else:
+            print(f"Unknown freq multiplier: {_mem.freq_flags}")
```
kk7ds, line 483: This needs to be a `LOG` call.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+            mem.freq = 1234567890
+            mem.offset = 0
```
kk7ds, line 485: It would be better to raise an exception here. If anyone ever hits this, they're just going to file a bug saying "I put in 146.52 and the radio shows 123.45". Raise an exception explaining the problem so they'll see it.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+            mem.skip = ""      # None
+        elif _flag.skip == 1:
+            mem.skip = "S"     # memscan skip (aka "Skip")
+        elif _flag.skip == 3:
+            mem.skip = "P"     # 3 = mem&vfo ("Pskip")
+
+        # Channel names are encoded in 6x 6-bit groups spanning 5 bytes.
+        # Mask only the needed bits and then lookup character for each group
+        # Mem format: 0000AAAA AABBBBBB CCCCCCDD DDDDEEEE EEFFFFFF
+        mem.name = CODED_CHRS[((_mem.name0 & 0x0F) << 2)
```
kk7ds, line 506: Please use `icf.warp_byte_size()` which handles this in both directions for you instead of all the bit math.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+            rset = RadioSetting("pgmscanedge/%d.name" % kx,
+                                f"Program Scan {kx} Name", rx)
+            # rset.set_apply_callback(myset_psnam, _pses, kx, "name", 6)
+            edges.append(rset)
+
+            # Freq (Hz) is 1/3 the raw value (expect this allows N x 8.333kHz)
+            flow = 1.0 * (_pses[kx].start3 * 65536 * 256 +
+                          _pses[kx].start2 * 65536 +
+                          _pses[kx].start1 * 256 +
+                          _pses[kx].start0) / 3e6
```
kk7ds, line 554: This almost certainly means you should be using a u32. Why not?

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+            fhigh = 1.0 * (_pses[kx].end3 * 65536 * 256 +
+                           _pses[kx].end2 * 65536 +
+                           _pses[kx].end1 * 256 +
+                           _pses[kx].end0) / 3e6
+
+            if (flow > 0) and (flow >= fhigh):
+                flow, fhigh = fhigh, flow
+
+            print(f"EDGE: {flow} -> {fhigh}")
```
kk7ds, line 563: Remove this before we merge.

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+            #                    "-- Scan %d Freq Step" % kx, rx)
+            # rset.set_apply_callback(myset_tsopt, _pses, kx, "tstp", bxnd)
+            # edges.append(rset)
+
+            # rx = RadioSettingValueList(mdopt, mdopt[ndxm])
+            # rset = RadioSetting("pgmscanedge/%d.mode" % kx,
+            #                    "-- Scan %d Mode" % kx, rx)
+            # rset.set_apply_callback(myset_mdopt, _pses, kx, "mode", bxnd)
+            # edges.append(rset)
+        # End for kx (Edges)
```
kk7ds, line 631: I'd really rather us not merge with all this commented-out code. I assume this is TODO work. Before we merge can you strip this out into a separate commit you can iterate on later?

chirp/drivers/icr6.py
```diff
@@ -0,0 +1,794 @@
...
+        return group
+
+    def set_memory(self, mem):
+        _mem = self._memobj.memory[mem.number]
+        _flag = self._memobj.flags[mem.number]
+
+        if mem.empty:
+            _flag.hide_channel = 1
+            return
+        _flag.hide_channel = 0
```
kk7ds, line 692: This is a nit, but above the if, just do this:
```
_flag.hide_channel = mem.empty
```
and it'll do the right thing :)

kk7ds (changes requested):
I assume based on the mailing list thread that you've figured this out, but your test image needs to go in `tests/images` :)

Thanks for all your work on this!

jbradsha:
Closed - will create a new PR once most of the above is fixed (some queries around endian - radio uses both big and small endian in different places).

kk7ds:
Please don't open a new one, just re-open this and push to it. That keeps the discussion history in tact and makes it easier to catch back up when you have another version.

Thanks!

jbradsha:
Thanks Dan. Will upload fresh copy shortly.. just running it through tox again.  Changed not all done (the different endians still need looked at. Not sure why Icom even did that, even if they used legacy firmware from another IC-Rx model would have expected it to be consistent) and the warp_bytes for names not working as expected so not added to the PR code.

chirp/drivers/icr6.py
jbradsha: Not working for me. Will debug locally and hopefully get it into a future PR.

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: I need to check the endian, the radio has both types (see offset for example)

chirp/drivers/icr6.py
jbradsha: I need to check the endian, the radio has both types (see canceller_freq for example)

chirp/drivers/icr6.py
jbradsha: Not visible in CS-R6 either so have removed all those comments.

chirp/drivers/icr6.py
jbradsha: On the to-do list, I need to figure out the endians, radio uses both.

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: Fixed... same list.

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: Need to figure out the endian, kept on the "to do" list for now.

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: Done

chirp/drivers/icr6.py
jbradsha: Fixed

chirp/drivers/icr6.py
jbradsha: IC-R6 packs 6x 6-bit characters into 40 bits, the first 4 bits are unused/padding
Looking at the id31 and id5100.py drivers which icf.warp_byte_size() they seem to have 8x 7-bit characters packed into 7 bytes so no need for padding. Think those leading 4 bits is breaking it.

kk7ds: This should let you pad/skip the bits: https://github.com/kk7ds/chirp/pull/1084

jbradsha: Worked a treat, driver updated to use it both ways

helmick43270:
Hello, just dropping in after reading the Service Manual and instruction book. Was toying with the idea of trying my frankensteined Kenwood level converter on my new (to me) IC-R6 and having a hexadecimal conversation with it until I read this.  Hope everyone is doing okay. This just seems like a lot of trouble for something not many people are requesting, which is exactly why it’s interesting to me.

jbradsha:
Hi,

There's a working "test module" attached to https://chirpmyradio.com/issues/2795 . It'd be great to get some feedback, had a spike of interest earlier last year but nobody provided any feedback on the module.  Code should be fairly similar to this pull request. Even getting a bit of activity on that feature request ticket would be great, may encourage others to test and feedback.

Most of the device memory is mapped out, including the auto-scan channels (which are the same as the first 1300 channels), I haven't raised a PR for a while so that's only in the local version.

Bit of tidy up needed as the IC-R6 uses different endianness in different places which has complicated the Python code a little.  I've a bit more work to do on the GUI for the device settings too.

Regards,
John Mi0SYN

PR #1085, opened 2024-08-11
Bump wxPython version to support newer python versions

lekv:
I noticed that wxPython couldn't be installed as a dependency with Python 3.12 on MacOS. It failed with the following error message:

ModuleNotFoundError: No module named 'attrdict'

After consulting https://github.com/wxWidgets/Phoenix/issues/2296, I found out that this was an issue with wxPython and confirmed that 4.2.1 works. I tested this locally.

# CHIRP PR Guidelines

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 wxPython>=4.0,<4.2.0 ; platform_system=="Linux" # See https://github.com/wxWidgets/Phoenix/issues/2225
-wxPython==4.2.0 ; platform_system!="Linux"
+wxPython==4.2.1 ; platform_system!="Linux"
 pyserial
 requests
 pywin32; platform_system=="Windows"


lekv:
@kk7ds — please let me know if you'd prefer I open a new bug. The issue and fix are related to https://chirpmyradio.com/issues/10577 but one could argue the are different things.

kk7ds:
I'm haven't bumped yet because of this: https://github.com/wxWidgets/Phoenix/issues/2499 .. Hoping for a newer release with a fix at some point. However, right now the macOS builds are done with 3.10 and 4.2.0 and I want to keep these requirements in sync. The window-opens-behind bug is incredibly annoying.

But yeah, if you prefer you can just manually install 4.2.1 locally, obviously, if the python version is more important to you.

rhaist:
Some data for @kk7ds on the [Phoenix#2499](https://github.com/wxWidgets/Phoenix/issues/2499) blocker, since I was looking at this PR:

**Environment:**
- macOS 26.4.1 (Apple Silicon, arm64)
- iTerm2 3.6.9
- Python 3.12.13 in isolated uv venvs
- Tested `wxPython==4.2.1` (`wxWidgets 3.2.2.1`) and `wxPython==4.2.4` (`wxWidgets 3.2.8`) side-by-side.

**Tests:**
1. A minimal `wx.App` + `wx.Frame` reproducer launched from the terminal.
2. The full `chirpwx.py` from master.

**Result:** On both 4.2.1 and 4.2.4, the CHIRP window opens in the foreground and receives focus on first launch. I cannot reproduce the window-opens-behind regression on either version here.

This obviously doesn't prove 4.2.4 fixes it — the bug clearly reproduces reliably for you — but it does suggest the trigger depends on some environmental factor we haven't pinned down. Would you mind sharing your repro conditions? Specifically:
- macOS version and chip (Intel/Apple Silicon)
- Terminal application (Terminal.app / iTerm2 / other)
- Whether it reproduces on first launch after reboot, or every launch
- Whether the terminal window is in the foreground at the moment `chirpwx.py` runs

With that I can try to replicate on my end, and if 4.2.4 does fix it we could revisit the version bump.

Also worth noting: Phoenix#2499 is still open with no "fixed in" comments, and the 4.2.2 / 4.2.3 / 4.2.4 changelogs don't mention it — so even if it's silently fixed, upstream hasn't confirmed it.

PR #1087, opened 2024-08-15
Bug/11423 draft alternatives for the uvk5 99

ei2081:
# CHIRP PR Guidelines


ei2081:
TOX help pls - looks like unit-test suggests this model i added is part of an alias ? (... # Make sure the alias model is NOT in the directory...)
Perhaps it ought to be simple an addition to the supported firmware codes ?

Will look into the support matrix fun after that

ty!

pbarrette:
Looks to me like you're adding a new model called "UV-K5(99)" but it already exists as an alias that's mapped to "UV-K5" in the "/chirp/share/model_alias_map.yaml" file.

chirp/drivers/uvk5.py
```diff
@@ -2083,3 +2064,19 @@ class RA79Radio(UVK5Radio):
...
+# I'm not 100% sure it's legit Quansheng!
+
+
+@directory.register
+class UVK5_99_Radio(UVK5Radio):
+    MODEL = "UV-K5(99)"
+
+    @classmethod
+    def k5_approve_firmware(cls, firmware):
+        approved_prefixes = ('OSFW-bd90ca3')
```
kk7ds, line 2081: If there is no difference between the models, there is no need for a separate subclass, just add this to the list in the base class. Given that it's already in the alias list as such, that's what people are already expecting.

Also, it looks like the suffix (after the dash) is a git hash, which means it's likely to change. So probably just add `OSFW` to that list. Unfortunately, that probably also means it's going to be unstable in the future, but that seems to be the destiny of this model :/

chirp/drivers/uvk5.py
```diff
@@ -1,29 +1,30 @@
...
+    # (at your option) any later version.
+    #
+    # This program is distributed in the hope that it will be useful,
+    # but WITHOUT ANY WARRANTY; without even the implied warranty of
+    # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    # GNU General Public License for more details.
+    #
+    # You should have received a copy of the GNU General Public License
+    # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+    """
```
kk7ds, line 27: This is random whitespace damage and must be removed/restored.

chirp/drivers/uvk5.py
```diff
@@ -272,29 +273,9 @@
...
-    205, 212, 223, 225, 226, 243, 244, 245, 246, 251, 252,
-    255, 261, 263, 265, 266, 271, 274, 306, 311, 315, 325,
-    331, 332, 343, 346, 351, 356, 364, 365, 371, 411, 412,
-    413, 423, 431, 432, 445, 446, 452, 454, 455, 462, 464,
-    465, 466, 503, 506, 516, 523, 526, 532, 546, 565, 606,
-    612, 624, 627, 631, 632, 654, 662, 664, 703, 712, 723,
-    731, 732, 734, 743, 754
-]
+CTCSS_TONES = chirp_common.TONES        # tone list was exact match to common.
+DTCS_CODES = chirp_common.DTCS_CODES    # tone list was exact match to common.
```
kk7ds, line 277: This really should be a separate commit from the one adding the prefix.

kk7ds (changes requested):

chirp/drivers/uvk5.py
pbarrette: Be aware that the hardware is the same, but some vendors are likely now shipping radios with a specific build of fagci-mod firmware.

It's documented here by armel (F4HWN): https://www.youtube.com/watch?v=mhgHB2H6n34

TLDW: The OSFW bit comes from https://github.com/fagci/uv-k5-firmware-fagci-mod
The bd90ca3 bit comes from the hash for a specific commit here: https://github.com/fagci/uv-k5-firmware-fagci-mod/tree/bd90ca3f1621204ca2c17f21f0d1fb6ef5caea8d

So it may not be directly compatible with the original UV-K5.

kk7ds: Yep, I'm aware, which is why I added this firmware checking and detection system for the UV-K5. I'm assuming the submitter has tested the OSFW stuff with the current driver since they submitted it with no delta from the base. If there is a difference, then we definitely need a subclass, but one that makes changes from the base class, like `uvk5_egzumer.py` does. And, it needs to be `@detected_by` the base class and not just `@register`'d so it doesn't show up as its own model.

ei2081: i will do some reading, some more testing, and resubmit. Fab resources ty.

PR #1088, opened 2024-08-15
RT490: Add supported charset

pbarrette:
Added additional characters supported by the RT-490 display. Tested against MMLradio JC-8629.

Fixes issues 11020 and 11309.

### DIFF
--- a/chirp/drivers/radtel_rt490.py
+++ b/chirp/drivers/radtel_rt490.py
@@ -421,6 +421,9 @@ class RT490Radio(chirp_common.CloneModeRadio):
                (0x3FC0, 0x4000)
               ]
 
+    _valid_chars = chirp_common.CHARSET_ALPHANUMERIC + \
+        "`~!@#$%^&*()-=_+[]\\{}|;':\",./<>?"
+
     if RT490_EXPERIMENTAL:
         # Experimental driver (already heavily tested)
         _ranges = [(0x0000, 0x2400), (0x3400, 0x3C40), (0x3F80, 0x4000)]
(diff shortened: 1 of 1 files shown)


chirp/drivers/radtel_rt490.py
```diff
@@ -421,6 +421,9 @@ class RT490Radio(chirp_common.CloneModeRadio):
                (0x3FC0, 0x4000)
               ]
 
+    _valid_chars = chirp_common.CHARSET_ALPHANUMERIC + \
+        "`~!@#$%^&*()-=_+[]\\{}|;':\",./<>?"
+
```
kk7ds, line 426: This isn't used anywhere so this won't have any effect. I assume you want to set `valid_characters` in `get_features()` ?

pbarrette: And yet it works, so I don't know what to tell you.

I took the example directly from mml_jc8810.py

kk7ds: It really doesn't. I changed it to just `'a'` and I can still put in names like "FOO". With and without your patch, I can only put in uppercase characters.

The `mmp_jc8810.py` actually references it exactly where I said, which is why it works there and does not work here:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/mml_jc8810.py#L470

pbarrette: Well, crap. I clearly missed that while I was trying to not re-contaminate with the NEEDS_COMPAT_SERIAL that still existed in my original file.

kk7ds (commented):
Can you squash these into a single commit (per the PR guidelines template)? The delta goes to the mailing list on every build and there's no need to tell 2000 people about both changes, just the fix :)

Thanks!

kk7ds:
And also please give that commit a useful commit message so the recipients know what you fixed. Please also (per the guidelines) put `Fixes #11020` and `Fixes #11309` in the commit message so it links the bugs as well.

pbarrette:
Done.

PR #1089, opened 2024-08-16
RT490: Add supported charset - Fixes #11020 - Fixes #11309

pbarrette:
Added additional characters supported by the RT-490 display.


kk7ds:
Looks good, I'll merge this in the next couple days for the next build.

Note for the future, you can just do this in-place and `git push -f` to update the PR so you don't have to create a new one.

Thanks!

pbarrette:
Thanks for the help.

I'm using Github Desktop at the moment. It doesn't always sync the way you'd expect and I didn't trust a force push, so it was just easier and cleaner to rebase.

PR #1092, opened 2024-08-16
driver - UVk5 - ctss & dtcs codes match common, reference common instead of redefining.

ei2081:
This is a minor update to re-use chirp_common code ctcss/dtcs definitions rather than duplicating. 
I assume this is preferred.


kk7ds:
Yep, thanks, but please squash this into one commit with a good commit message first line for the build announcement. Use `git rebase -i`, squash them all, update the commit message, then `git push -f` to update this PR.

Thanks!

ei2081:
think i got it with a little further help from [so](https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together) but out of my usual comfort zone with git there :D

PR #1095, opened 2024-08-24
Fix typos

dforsi:
This PR fixes some user-visible strings and some comments.


chirp/wxui/bugreport.py
```diff
@@ -168,7 +168,7 @@ def _build(self, vbox):
                 self,
                 _('The debug log file is not available when CHIRP is run '
                   'interactively from the command-line. Thus, this tool will '
-                  'not upload what you expect. It is recomended that you '
```
kk7ds, line 171: Ouch, that was mine and recently :/

PR #1122, opened 2024-09-16
ft8100: Changed upload and downlod prompt

N7XSD:
Added classmethod get_prompts to give upload and download instructions specific to the FT-8100.


kk7ds:
You marked the strings for translation, which means the catalog needs to be regenerated. I can just do this for you if you're not able, but:
```
$ make -C chirp/locale clean all
```
will do it.

Also, next time `git push -f` will re-use the same PR, which is preferable :)

Thanks!

N7XSD:
Thanks.

Followed both of your suggestions. I'm still learning git.

kk7ds:
Aren't we all :)

Thanks!

kk7ds:
I just rebased and fixed a typo in the first commit subject (since that gets emailed to about 2000 people when the next build happens).

PR #1141, opened 2024-10-15
New device: TYT TH-UV98

Augis154:
Added support for TYT TH-UV89 handheld radio
Solves #9842


kk7ds:
Thanks, looks mostly good. The bug reference must be in the commit message, not the PR note and it must be exactly in the format `Fixes: #9842`. I just pushed that update for you, but just FYI for the future.

PR #1145, opened 2024-10-17
Add Baofeng UV-5RH

kala13x:
Added support for Baofeng UV-5RH 10W handheld radio

### DIFF
--- a/chirp/drivers/baofeng_uv17Pro.py
+++ b/chirp/drivers/baofeng_uv17Pro.py
@@ -1554,3 +1554,20 @@ class F8HPPro(UV17Pro):
                               False, CHARSET_GB2312))
         rs.set_apply_callback(apply_stationid, _nameobj)
         basic.append(rs)
+
+
+@directory.register
+class UV5RH(UV17Pro):
+    VENDOR = "Baofeng"
+    MODEL = "UV-5RH"
+
+    VALID_BANDS = [UV17Pro._airband, UV17Pro._vhf_range, UV17Pro._vhf2_range,
+                   UV17Pro._uhf_range, UV17Pro._uhf2_range]
+    POWER_LEVELS = [chirp_common.PowerLevel("High", watts=10.00),
+                    chirp_common.PowerLevel("Low", watts=2.00),
+                    chirp_common.PowerLevel("Medium", watts=5.00)]
+    SCODE_LIST = ["%s" % x for x in range(1, 16)]
+    SQUELCH_LIST = ["Off"] + list("123456789")
+    LIST_PW_SAVEMODE = ["Off", "1:1", "1:2", "1:4"]
+    MODES = UV17Pro.MODES + ['AM']
+    _has_workmode_support = True


chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -1554,3 +1554,20 @@ def apply_stationid(setting, obj):
                               False, CHARSET_GB2312))
         rs.set_apply_callback(apply_stationid, _nameobj)
         basic.append(rs)
+
+
+@directory.register
+class UV5RH(UV17Pro):
+    VENDOR = "Baofeng"
+    MODEL = "UV-5RH"
```
kk7ds, line 1562: I haven't gone to look, but this is different than the "UV-5RH Pro Max" as listed on the front page of the site?

kala13x: According to the Baofeng docs "UV-5RH Pro Max" has max output power of 14.5W while "UV-5RH" has only 10W, so looks like they are different.

kala13x: @kk7ds I just hit the update branch button on this PR and one test failed with an error "Please do not include merge commits in your PR", should I close this one and make another?

kk7ds: I guess I'm missing how they're (substantially) different:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/baofeng_uv17Pro.py#L1387-L1395

Same bands, same (number of) power levels, power safe, modes. The squelch list is longer here, but that doesn't mean the other list is right for all the sub-variants.

The actual power values in the POWER_LEVELS don't make much difference. They don't control the actual output of the radio, they're just used when chirp is importing data from a dissimilar radio to try to pick the closest available power level.

Anyway, we normally try to avoid tons and tons of basic duplication just to get the model right in the display. I usually try to err on the side of substantial difference or *especially* if the clone handshake differs (like the F8HP PRO radio in this file). This probably sounds silly, but I assume you've tested that this is identical enough that it doesn't need any of those definitions and that the 5RM driver technically works with your 5RH right?

As far as the merge, yeah please rebase instead of merge, but I can do it for you before I actually merge this.

kala13x: @kk7ds 
I didn't know if the power levels and squelch settings were so insignificant. 
Yeah, I tested the 5RM driver and this new change, which I use almost every day. 

P.S. I just did a rebase.

kk7ds: Okay, I was going to say you need to remove this from the alias map, but I guess the UV-5RH is not specifically in the alias map today. The UV-5RHP is, but that seems to be based on something else, and we do support the GM-5RH, but I guess that's quite different with a different magic string. What a mess these companies make of this stuff :)

vmari:
This commit bricked my radio (uv-5rh). After writing a band plan using this model, password protection was enabled, making radio useless. I'm aware that I'm 100% responsible of taking that risk. Just commenting here to let everyone know that it's not working as expected.

It seems some kind of corruption because original software is not able to write a new plan back. 

If you know how to solve it would be nice to convert my new souvenir back into a working radio 😓

kk7ds:
Please open a bug on the site and attach the image you wrote to the radio, along with the backups chirp makes. Also, please avoid using the word "bricked" until and unless we determine it's unrecoverable. I'll also say for other readers here that the UV-5RH is under *heavy* use by lots of people and this is the first I've heard of such a problem. This commit didn't actually add anything other than a listing of the UV-5RH to the download box as people have been using it successfully by selecting one of the other models.

PR #1152, opened 2024-10-29
fixes #11360 by initializing bell and favorite to 0 when channel is blanked

jacobcalvert:
# Description
This PR adds favorite and bell settings to the extra settings. 
Fixes #11360

--------------------

# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/alinco_dr735t.py
+++ b/chirp/drivers/alinco_dr735t.py
@@ -333,6 +333,18 @@ class AlincoDR735T(chirp_common.CloneModeRadio):
                            ))
         bcl.set_doc("Busy Channel Lockout")
 
+        fav = RadioSetting("fav", "Favorite",
+                           RadioSettingValueBoolean(
+                               False
+                           ))
+        fav.set_doc("Favorite Channel")
+
+        bell = RadioSetting("bell", "Bell",
+                            RadioSettingValueBoolean(
+                                False
+                            ))
+        bell.set_doc("Bell Alert")
+
         stby_screen = RadioSetting("stby_screen", "Standby Screen Color",
                                    RadioSettingValueList(
                                        self.SCREEN_COLOR_MAP,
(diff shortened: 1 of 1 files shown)


chirp/drivers/alinco_dr735t.py
```diff
@@ -283,6 +283,10 @@ def find_key_in(d: dict, target_val):
             _mem.power_index = self.POWER_MAP.index(
                 mem.power) if mem.power in self.POWER_MAP else 0
             _mem.skip = 0x01 if mem.skip == "S" else 0x00
+            # set favorite, bell to 1 if it's ANYTHING not the defaults or 0
+            _mem.favorite = 0x01 if _mem.favorite not in [0x00, 0xFF] else 0x00
```
kk7ds, line 287: Are you sure this works? `_mem.favorite` is a `dataElement` instance not an integer so I'm not sure this will ever be true.

kk7ds:
Can you please give this a reasonable commit message? What you have is not very meaningful for the git history and the build announcement. Perhaps:
```
dr735: Fix defaulting bell and favorite values

Fixes #11360
```

chirp/drivers/alinco_dr735t.py
kk7ds: Also, would you prefer to add these to the extras so you can control them? Would also fix the defaulting... Not sure how important they are, but it's not hard to do and I can show you how if you want.

jacobcalvert: I would be happy to add them to extras since I already plumbed for some of the extras in the original driver PR. I'll work on that. Do you prefer I close this PR and open a new one with this bugfix and the extras or just update this one?

kk7ds: No, please just force push to this one, thanks.

LMK if you need help, but there should be some examples to copy in that driver already.

kk7ds (commented):
Sorry, forgot to click the button.

PR #1157, opened 2024-10-31
Added Wouxun kg-uvr5 support

gavazquez:
# CHIRP PR Guidelines


kk7ds:
Please don't copy driver files. Subclass the existing implementation and only change what is necessary. That is why the PR checks test is failing. Thanks!

gavazquez:
Will do, thanks :)

gavazquez:
@kk7ds I think most of the style issues should be solved now.
Let me know if there's something more that needs to be fixed and I'll do it :)

gavazquez:
@kk7ds I'm not used to program on python but let me know if there's something left to do :)

kk7ds:
> The original 920 driver lacked several memory sections. It didn't include some fields and some internal sections like TX power, band limits, possibility to change the step size, etc. That's why you see so many changes compared to the original 920pa driver

This is why the big code lift is bad, because I couldn't see the changes.

> If you want I can make it as a new firmware totally independent on the old driver but it will of course have many similarities

Your subclass can definitely have a different `MEM_FORMAT` from the base class if it makes sense, no problem there if that's what you mean. But no, I don't want it to be totally independent - I asked to subclass, I just don't want you to (effectively) delete and re-add the original driver in the process.

> Bear in mind that memory sections for all the different 920 models are the same, only the configuration values that they have change

That means that only really get_settings needs to be different? In that case, all you need to do is unify all those differences into a helper and call it from `get_settings()` and then override that in your subclass.

gavazquez:
Ok I might need to check out how to do the rebase as I totally messed it up :(

Regarding the code I'll try to implement what you say and upload it here and if it looks right to you I might just made a new push request

kk7ds:
Please keep it to this PR instead of opening a new one (you haven't yet). Looks like you rebased on something much older and pushed old commits from other people in here. Just fix it and `push --force` to keep it all here.

PR #1159, opened 2024-11-03
Add Tidradio TD-H3  settings valid as of 240416 firmware

johnsonm:
I have reviewed the settings in my TD-H3 running the HAM_240416 firmware, and added most of what was missing. If the `52.200TX` setting is enabled, it also allows 1.25m frequencies in the standard 220Mhz-250Mhz range to be programmed in duplex mode. 

Among some of the work that could have been done based on my settings audit but which I didn't do
* I didn't add Scramble because that mode shouldn't be in a ham radio...
* I didn't add anything for VFO settings like `47.Dec. code`
* I have exposed the `53.350TX` and `54.500TX` settings but not enabled any bands for upload because I don't know what those bands would legally be in any locale. However, the pattern I created would support adding that if there is a legitimate use for those frequencies on these radios.
* I may have missed some settings by mistake.

Some settings appear incompatible with older firmware versions, and I have found and implemented a guard using the `56.SCAN BAND` which is in a location that is not used by some older firmware versions; those firmwares have values in that location that are undefined on newer firmware versions to the best of my knowledge.

Additionally, I changed the TD-H3 power settings to the currently documented 2W/5W.

These changes are relative to `master` as of today, and pass the full test suite on my machine.

I consider these sufficiently minor that I didn't first open a separate CHIRP issue asking for work I intended to do myself.

As discussed on the mailing list, I have not replaced the TD-H3 test image. However, I have made a collection of images with various settings changes and a google sheet that I used to organize this work at https://drive.google.com/drive/folders/1BqRjX_4OElZVpYGFCnKVLLRwjUkWoMh8?usp=sharing and intend but do not promise to leave it around indefinitely.

I have not yet tested writing 1.25m frequencies to the radio, and would appreciate testing for any additional features that should be hidden behind the `56.SCAN BAND`  guard on older firmware versions.

Thanks for all your work on chirp over many years!

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -297,15 +297,28 @@ struct{
   u8 lsidekey4;
 } press;
 
-#seekto 0x0CA8;
+#seekto 0x0CA0;
 struct {
+  u8 unknown21:7,
+     dtmfdecode:1;
+  u8 unknown22:6,
+     dtmfautorst:2;
+  u8 unknown23:6,
+     dtmfdecoderesp:2;
+  u8 unknown24:5,
+     dtmfspeed:3;
+  u8 unknown25:4,
+     scanband:4;
+  u8 unknown26:8;
+  u8 unknown27:8;
+  u8 unknown28:8;
   u8 txled:1,
      rxled:1,
      unused11:1,
      ham:1,
      gmrs:1,
      unused14:1,
-     unused15:1,
+     dtmfst:1,
      pritx:1;
   u8 scanmode:2,
      unused16:1,
@@ -340,7 +353,10 @@ struct {
   u8 squelch;
   u8 tot;
   u8 rogerprompt:2,
-     unused11_4:4,
+     unused11_4:1,
+     tx220:1,
+     tx350:1,
+     tx500:1,
      lang:1,
      unused11_1:1;
   u8 save;
@@ -348,7 +364,8 @@ struct {
   u8 voxdelay;
   u8 onlychmode:1,
      breathled:3,
-     unused:3,
+     unused:2,
+     amband:1,
      alarm:1;
 } settings;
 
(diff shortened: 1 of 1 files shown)


johnsonm:
I found and fixed one bug I introduced while editing that prevented writing 1.25m channels. I have now successfully written a configuration including a 1.25m repeater to my radio.

johnsonm:
However, changing any settings with a 1.25m channel in place creates an error, so I need to fix that.

johnsonm:
The existing frequency guards are not very DRY. If that's OK, then it works now to modify configurations with those frequencies selected. But it might make sense to refactor those to use the new `get_tx_bands()` method instead of replicating the frequency lists.

johnsonm:
I followed an existing ON/OFF list pattern, but now realize that for binary choices I should have used checkboxes. I have removed all of them, including those that already existed, for a more consistent UX.

johnsonm:
It still disables duplex for 1.25m bands, so I have at least one more fix to do before it's ready.

kk7ds:
> I followed an existing ON/OFF list pattern, but now realize that for binary choices I should have used checkboxes. I have removed all of them, including those that already existed, for a more consistent UX.

This is likely because it was written by someone trying to mimic the factory software/interface which did everything with drop-downs. So, yep, better.

kk7ds:
> It still disables duplex for 1.25m bands, so I have at least one more fix to do before it's ready.

This works fine for me:

image

If I put in 223MHz, it auto-selects +1.6MHz per the band plan. If I put in 220MHz, then I can either select + duplex, type 1.6 and get it, or just do 1.6 in the offset field and then select + when asked.

johnsonm:
The problem was with the overlapping bands. The `get_memory()` and `set_memory()` functions didn't handle overlapping rx and tx bands. I took what seemed to me like the easiest, simplest approach:

```diff
@@ -1252,7 +1274,7 @@ class TDH8(chirp_common.CloneModeRadio):
                     FREQHOP_VALUES, current_index=_mem.freqhop))
             mem.extra.append(rs)
 
-        if chirp_common.in_range(mem.freq, self._rxbands):
+        if chirp_common.in_range(mem.freq, self._rxbands) and not chirp_common.in_range(mem.freq, self.get_tx_bands()):
             mem.duplex = 'off'
         if chirp_common.in_range(mem.freq, [AIRBAND]):
             mem.mode = 'AM'
@@ -1321,7 +1343,7 @@ class TDH8(chirp_common.CloneModeRadio):
         else:
             _mem.txfreq = mem.freq / 10
 
-        if chirp_common.in_range(mem.freq, self._rxbands):
+        if chirp_common.in_range(mem.freq, self._rxbands) and not chirp_common.in_range(mem.freq, self.get_tx_bands()):
             _mem.txfreq.fill_raw(b'\xFF')
 
         _mem.rxfreq = mem.freq / 10
```

I'm happy to squash the ON/OFF checkboxes down into the main commit if you'd find it cleaner. I have just rebased on current master. It passes tests for me, and I've also successfully chirped my radio with 1.25m offsets and duplex modes.

I'm not aware of anything else to fix, so I'll take this out of draft. Happy to have you tell me anything you would like changed, of course!

Thanks again for all the help this weekend, and for all your work through the years on this.

chirp/drivers/tdh8.py
```diff
@@ -1540,6 +1556,65 @@ def _get_settings(self):
                                 current_index=_press.lsidekey4))
             basic.append(rs)
 
+        if self.MODEL in H3_LIST:
```
kk7ds, line 1559: This is really not the best way to do this, as I'm sure you know, but unfortunately it's a prevalent pattern throughout the code base. No need to change, just complaining.

kk7ds (commented):
Nope, prefer the boolean refactors as a separate commit from fixes/features, thanks.

Just need to pass the style test target to be able to merge :)

chirp/drivers/tdh8.py
johnsonm: I tend to be a code chameleon — what pattern would you have preferred?

johnsonm:
> Just need to pass the style test target to be able to merge :)

When I click on details, the error just looks empty.

 I found `tox -e style` and fixed the ones I introduced. I didn't touch the failing ./chirp/drivers/alinco_dr735t.py that I didn't touch.

I've again rebased on latest master.

PR #1160, opened 2024-11-03
id31/51 fixes and rich help

kk7ds:
- **id51: Fix loading ICF files**
- **id31/51: Fix 6.25kHz offsets values**
- **id31/51: Fix TSQL/DTCS "error" on CS-31**
- **Add richer errors for specific driver issues**
- **Make baofeng_common radios suggest K1 help**
- **Collect system information for bug reports**
- **Fix bugreport on Linux (or older wx)**
- **Retry file uploads during bug report**


kk7ds:
FYI @Darieb .. not done yet, but this is close. You can test it by doing a download from the CHIRP-FakeClone radio and see the specials included in the banks tab.

PR #1167, opened 2024-11-14
add Baofeng UV-21Pro v2

maximeguillaud:
Add support for Baofeng UV-21 Pro V2


maximeguillaud:
Not sure if this fixes https://chirpmyradio.com/issues/10581 as I only have access to the v2 hardware.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -1575,3 +1575,19 @@ class UV5RH(UV17Pro):
     LIST_PW_SAVEMODE = ["Off", "1:1", "1:2", "1:4"]
     MODES = UV17Pro.MODES + ['AM']
     _has_workmode_support = True
+
+
+@directory.register
+class UV21ProV2(UV17Pro):
+    VENDOR = "Baofeng"
+    MODEL = "UV-21Pro-v2"
```
kk7ds, line 1583: I think we should just say `UV-21Pro` here. Since the bug you reference doesn't mention V2 but says that the existing UV-17Pro driver works, I think your concern over V1/V2 probably means that your driver will work for both. So yes, please also add the Fixes: tag to link/close that bug.

Also, I haven't eye-diffed this, but is this actually different from any of the above models? If so, then this looks okay, but if not, I think I'd prefer to just add an entry to the model_alias_map. There are so many total clones of the UV-17 that this file is starting to get littered with them :)

Thanks!

maximeguillaud: Hard for me to say if this differs from other models as I don't have access to all of them to test. In https://chirpmyradio.com/issues/10581 it is suggested to use the Baofeng UV-17Pro profile with the UV-21, however the UV-17Pro profile doesn't seem to support airband.
An alternative path to fix this would be to add airband to the UV-17Pro profile (which seems to support airband according to the specs I found online) and see whether it works with both UV-17Pro and UV-21; however in that case I would not be able to test with the UV-17Pro.

kk7ds: Okay so this looks identical to the above 5RH model with the exception of the power levels. Does the UV21 really only have hi/lo?

At the very least, please subclass that and just override the `POWER_LEVELS` bit so it's clear that it's otherwise identical.

kk7ds: Also looks maybe identical to the 17RPlus?

maximeguillaud: Indeed, the UV-17R-Plus profile seems to do the trick.

kk7ds: So should we replace this with just an update to `chirp/share/model_alias_map.yaml` ?

PR #1171, opened 2024-11-22
bug/11700/clone hang

kk7ds:
- **Fix clone dialog hang if serial open error**
- **Add developer serial-open-error test fixture**


kk7ds:
Unit failures are due to repeaterbook being down:

https://www.chirpmyradio.com/issues/11692

PR #1172, opened 2024-11-25
query sources: add przemienniki.eu

SP5BDW:
Related to #11430

* renamed `przemienniki` to `przemienniki_net`
* added `przemienniki_eu`
* implemented the same filtering as webpage [https://przemienniki.eu/](https://przemienniki.eu/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ assets
 .coverage
 chirpwx.spec
 chirp/locale/.files
+# for pycharm IDE
+.idea
--- a/chirp/locale/bg_BG.po
+++ b/chirp/locale/bg_BG.po
@@ -16,7 +16,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-10 10:50-0800\n"
+"POT-Creation-Date: 2024-11-27 12:36+0100\n"
 "PO-Revision-Date: 2024-09-05 02:01-0700\n"
 "Last-Translator: Стоян <stoyanster от гмаил>\n"
 "Language-Team: Bulgarian <dict@ludost.net>\n"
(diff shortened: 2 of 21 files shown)


chirp/wxui/config.py
```diff
@@ -158,6 +158,16 @@ def get_bool(self, key, section=None, default=False):
...
 
+    def get_str_list(self, key, section=None, default=[]):
+        val = self.get(key, section)
+        if val is None:
+            return default
+        else:
+            return val.split(',')
+
+    def set_str_list(self, key, value, section=None):
+        self.set(key, ','.join(value), section)
```
kk7ds, line 169: You need to make sure the value has no commas in it here, else it will not store/load the same value.

kk7ds (commented):
This doesn't look right for me on macos:

image

I'd prefer you stick to the same sort of simple layout (name: field) as the other query dialogs for consistency. This style also lends itself to working on all the platforms with reasonable success.

Also, while poking around, I tried selecting nothing in any of the select boxes (band, mode, etc) and querying. That worked, but the next time I tried a query, the dialog does not open and I get this stack trace:
```
  File "/Users/dan/Documents/chirp.git/chirp/wxui/main.py", line 1946, in _menu_query_przeu
    self._do_network_query(query_sources.PrzemiennikiEuQueryDialog)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/main.py", line 1925, in _do_network_query
    d = query_cls(self, title=_('Query %s') % query_cls.NAME)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/query_sources.py", line 191, in __init__
    vbox = self.build()
  File "/Users/dan/Documents/chirp.git/chirp/wxui/query_sources.py", line 721, in build
    self._mode.SetCheckedStrings(prev)
  File "/Users/dan/Library/Application Support/pipx/venvs/chirp/lib/python3.10/site-packages/wx/core.py", line 2694, in _CheckListBox_SetCheckedStrings
    assert s in self.GetStrings(), "String ('%s') not found" % s
AssertionError: String ('') not found
```

Do we need to keep both of these? I thought the original one was down (appears up now) but that the owner was not being very cooperative. If the new one is going to supplant the old, can/should we just replace the old one with this one?

Thanks!

kk7ds:
> Changed to the same layout as the old one.

Thanks, looks better now.
 
> As for keeping both - unfortunately both lists are populated by the community and differ a bit. Also:
> 
> * przemienniki.net has repeaters from other countries, not only PL. But as you said, the owner isn't much cooperative, the page might go down at any time
> * przemienniki.eu is still under development and right now lists only polish repeaters.
> 
> I would keep both for now, maybe i should squash them into a single dialog with a `wx.Choice` to select between them? I don't know if making a dynamic dialog like that would be a good idea.

The RadioReference query is a little bit like that because CA/NA are different. However, it's too complex, IMHO. It's fine to keep both if they're substantially different, but if the old one dies out, we should remove it to reduce clutter I think.

I updated this to fix the style errors and change the comment in the commit message so it will close the referenced bug on merge. The other PR check failure is a false positive comparing this to other drivers (which it shouldn't be doing) so it can be ignored.

Thanks!

PR #1173, opened 2024-11-26
Fix typos

dforsi:
This PR fixes some minor spelling errors.


chirp/drivers/tk760g.py
```diff
@@ -1035,8 +1035,8 @@ def set_memory(self, mem):
         # frequency
         _mem.rxfreq = mem.freq / 10
 
-        # this are a mystery yet, but so falr there is no impact
```
kk7ds, line 1038: I think "mystery" is correct here.

dforsi: you are right, I removed that

PR #1174, opened 2024-11-26
Do a case insensitive sort of available translations for CHIRP

dforsi:
The list of languages available in the View menu is sorted in a case sensitive way. This PR applies a case insensitive sort that makes the list easier to read.

This PR also avoids an exception caused by `GetAvailableTranslations()` returning also `messages` as an available language because `chirp/locale/` contains the file `messages.po` (this doens't happen on Linux when CHIRP is installed because that file isn't installed). A better fix would be to skip any object that doesn't have a `DescriptionNative` property but it would make the code more complicated for an edge case.

| Before (case sensitive) | After (case insensitive) |
| ------------- | ------------- |
| CHIRP language case senstitive | CHIRP language case insenstitive |


kk7ds:
👍

PR #1175, opened 2024-11-27
Add Radtel RT-470X Plus BT signature, separate class and test file

OK2MOP:
The updated driver to solve "New Model" [issue #11695](https://chirpmyradio.com/issues/11695) created on request of Radtel (as I don't have the radio myself) and tested by them. A new signature has been added, however PTTID setting had to be removed as invalid value is read from radio at that position. 

In the end I have decided to create a new class in case there are additional changes (e.g. Bluetooth setting), but this is unlikely as all models share same CPS according to Radtel.

1. Newest master has been used
1. Single commit
1. Related to #11695
1. Commit message indicates the RT-470X Plus BT  model
1. Test image Radtel_RT-470X_BT.img is included (and was one of the reasons to create a separate class)


kk7ds:
I rebased this and also added `Fixes: #11695` to the commit message for proper linkage to the bug. The comments here in the PR are only seen in github, not in git history, which means the bug tracker doesn't see them.

PR #1176, opened 2024-11-27
Add Radtel RT-620 basic driver and test image based on UV-17 driver

OK2MOP:
I have create a driver referenced in  "New Model" [issue #11699](https://chirpmyradio.com/issues/11699) for this radio, which shares memory organisation  with UV-17 model (with different settings positions in many cases but same memory organisation and separate MEMORY_DEFS needed). As the number of differences was  quite high in the end, I put it in separate file. Only GPS model was accessible to me, but the non-GPS would probably just miss the GPS settings. The work was based on stock CPS, however some less used features (like scan ranges) were not implemented as I got info from Radtel that this model will probably be grandfathered soon. The same is true for some parameters of VOX (frequency, codes) which I was not able to figure out but they are mostly set in radio directly anyway. Many of them are in memory definitions but the driver code to process them in RadioSetting groups is missing.

Channel banks are not in memory map as far as I can tell and were causing errors mentioned in issue, so they have been disabled. The issue also contains a text file with description of remaining unimplemented (rarely changed) features.

1. Newest master has been used
1. Single commit
1. Related to #11699
1. Commit message indicates the  RT-620  model (maybe GPS should have been added explicitly?)
1. Test image Radtel_RT-620.img is included (contains extra local repeater not in stock factory setting to demonstrate some settings missing there, hopefully this is OK)


OK2MOP:
OK, I just found that I have kept the original copyright header from source file, fixed that

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,545 @@
...
+    RadioSettingValueBoolean, RadioSettingValueList
+from chirp.drivers import baofeng_uv17Pro, baofeng_uv17
+
+LOG = logging.getLogger(__name__)
+LIST_DTMFST = ["Off", "DT-ST", "ANI-ST", "DT+ANI"]
+LIST_SAVE = ["Off", "1:1", "1:2", "1:3", "1:4"]
+
+
+def _download_cal(radio):
+    if hasattr(radio, 'CAL_BLOCK') and \
```
kk7ds, line 28: It would be better to use `isinstance(radio, RT620)` here. However, these download/upload functions are only referenced from your class, so I think you can/should just get rid of this check.

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,545 @@
...
+        if _mem.rxtone == 0:
+            _mem.rxtone = 0xFFFF
+        if _mem.txtone == 0:
+            _mem.txtone = 0xFFFF
+
+    def get_features(self):
+        rf = super().get_features()
+        rf.has_tuning_step = True
+        rf.valid_tuning_steps = self.STEPS
+        rf.valid_modes = self.MODES
```
kk7ds, line 541: This is already done on the base class so you should not need to do it again here.

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,545 @@
...
+        _mem.wide = self.MODES.index(mem.mode)
+        if _mem.rxtone == 0:
+            _mem.rxtone = 0xFFFF
+        if _mem.txtone == 0:
+            _mem.txtone = 0xFFFF
+
+    def get_features(self):
+        rf = super().get_features()
+        rf.has_tuning_step = True
+        rf.valid_tuning_steps = self.STEPS
```
kk7ds, line 540: Can you just convert the base class to do this the same way it does `valid_modes` below? Then you can just define `STEPS` on your class (as you have done here) and be good.

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,545 @@
...
+        if number == -1:
+            return self._memobj.vfo.b
+        elif number == -2:
+            return self._memobj.vfo.a
+        else:
+            return super()._get_raw_memory(number)
+
+    def get_memory(self, number):
+        #        if isinstance(number, str):
+        #            number = self._get_specials()[number]
```
kk7ds, line 507: Please remove all this commented-out code for the final version. Without syntax highlighting it can be very difficult to distinguish commented-out code from actual code.

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,545 @@
...
+    struct proglocks locks;
+    #seekto 0x8000;
+    struct pttid pttid[15];
+    struct ani ani;
+    #seekto 0x8061;
+    struct dtmfen dtmf;
+    """
+    MEM_FORMAT = MEM_DEFS + MEM_LAYOUT
+    REMOVE_SETTINGS = ['settings.scanmode']
+    REMOVE_GROUPS = []
```
kk7ds, line 232: This is always empty, so what is the point of it?

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,545 @@
...
+@directory.register
+class RT620(baofeng_uv17.UV17):
+    """Radtel RT-620"""
+    VENDOR = "Radtel"
+    MODEL = "RT-620"
+    _radio_memsize = 0x2efff
+    _has_workmode_support = True
+    _has_gps = True
+    _vfoscan = True
+    _has_gps = True
```
kk7ds, line 50: This is redefined from two lines above

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,545 @@
...
+        # Settings menu 30
+        if _mem.settings.sk2long >= len(self.LIST_KEYS):
+            val = 0x00
+        else:
+            val = _mem.settings.sk2long
+        rs = RadioSetting("settings.sk2long", "Side Key 2 Long Press",
+                          RadioSettingValueList(
+                              self.LIST_KEYS, current_index=val))
+        extra.append(rs)
+        if self._has_top_key:
```
kk7ds, line 389: This is always `True` so you don't need this. Are you planning or aware of some other radio that will subclass this and set this to be `False`?

kk7ds (changes requested):
Please put your `Fixes: #xxx` in the *commit message* so it's seen by the bug system.

Thanks!

chirp/drivers/radtel_rt620.py
OK2MOP: These were relicts when I was trying to put VFO1 and VFO2 into channel list but when I did not succeed I have left it there if somebody wanted to continue. It will be removed.

chirp/drivers/radtel_rt620.py
OK2MOP: The idea was to make it possible to download and back up the calibration block if present but not upload it back again. If the Calibration block in (potential) inherited class is set to None/False it will be ignored, otherwise it will be applied for the base class for download and removed for upload.

chirp/drivers/radtel_rt620.py
kk7ds: Thanks. You're right to want to put those as special channels instead of the "work mode" approach that so many drivers have taken. I might take a shot at fixing that in the base UV-17 driver and maybe I'll tag you to check my work applies for your driver as well, if you're willing.

chirp/drivers/radtel_rt620.py
kk7ds: 👍

chirp/drivers/radtel_rt620.py
OK2MOP: I originally did not want to touch the other source files, so that I do not break something somewhere in inheritance (as this was a standalone test driver not touching anything else) but it is probably a good idea. Only I am not sure where else it is inherited, will have to check.

chirp/drivers/radtel_rt620.py
OK2MOP: This is actually always `False`, but the official CPS for this radio sets top keys. However, the model of the radio is not equipped with the key. I was suspecting the non-GPS version or some other radio with the same CPS has it, so   I implemented it.

chirp/drivers/radtel_rt620.py
OK2MOP: As you are mentioning it for `rf.valid_modes = self.MODES` , it's true, the base class should use my values. Is it possible the base class was refactored in last month or so and originally both STEPS and MODES were defined as global variables in the original class (with STEPS staying that way for now)? Because everything which was in base class I have tried to use directly (but I could have overlooked this).

chirp/drivers/radtel_rt620.py
OK2MOP: I am willing, there is probably no urgency to have this driver merged right now, better to do everything right and the amount of code here could be significantly reduced (whole `get_settings_workmode` could be removed in that case). 

Only I now remember that when I was experimenting with it it seemed the TX and RX frequency was stored differently in VFOs, but that may be actually just two ways of storing them which are interchangeable.

kk7ds: Up to you, I'd rather just get it in and make incremental changes. We can also just remove the workmode stuff for this merge and then follow up. The fact that they're stored differently for the VFOs isn't a problem, just needs a slightly different approach for `get_memory()`

chirp/drivers/radtel_rt620.py
OK2MOP: This is actually the GPS version (should this be added to model name then?), I don't have access to non-GPS one unfortunately.

kk7ds: Nah, let's just leave it as it is and deal with that later if needed.

chirp/drivers/radtel_rt620.py
kk7ds: No, not preferred at all. The drivers I've written do it the proper way (or exclude the VFOs because I see very little point in letting those be programmed in advance by PC). I think the OEM software for a lot of these chinese radios implement them as separate settings and that pattern has worked its way into CHIRP, even though it's less ideal.

No rush, thanks very much for your work on this. I will plan to push the pending changes tomorrow for a build on Friday. If we can include this by then, then it will be in that release, but if not, the following week.

OK2MOP:
I will review it tomorrow as I am really tired and although it seems to work, I may have made a mistake somewhere. Plus I did not expect the push to actually auto-propagate here (but I should have). Plus for some reason the rebase in my code did not happen.

kk7ds:
Yep, local rebase and force push is the way to go.

chirp/drivers/radtel_rt620.py
kk7ds: Sorry, I missed this reply.

> The idea was to make it possible to download and back up the calibration block if present but not upload it back again. If the Calibration block in (potential) inherited class is set to None/False it will be ignored, otherwise it will be applied for the base class for download and removed for upload.

Okay, but the `hasattr()` will always be true on this and subclasses - you won't really be able to "delete" the class attribute from subclasses. So in this case, I think just remove the `hasattr` and keep the rest of what you've got (aside from the comment below).

I would prefer if we do not upload the calibration data without an explicit opt-in. Too many users upload the image from one radio into another without knowing the problems it may bring. We used to always download and upload everything on radios where we grabbed this data and we had a lot of problems. One way is to have a "virtual" setting in the settings dialog that will enable upload of the calibration data, where we can warn the user before they enable it.

By my reading of this, you're currently always downloading and never uploading this, right?

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,437 @@
...
+    RadioSettingValueBoolean, RadioSettingValueList
+from chirp.drivers import baofeng_uv17Pro, baofeng_uv17
+
+LOG = logging.getLogger(__name__)
+
+
+def _download_cal(radio):
+    if hasattr(radio, 'CAL_BLOCK') and \
+            radio.CAL_BLOCK not in radio.BLOCK_ORDER:
+        radio.BLOCK_ORDER.append(radio.CAL_BLOCK)
```
kk7ds, line 28: This is not doing what (I think) you think it's doing. This is modifying the effectively-global definition of `UV17.BLOCK_ORDER`, as is the `.remove()` below(). That means if I do the following:

1. Download from an RT620 (`BLOCK_ORDER` gets modified)
2. Download from a UV17 in a new tab (modified `BLOCK_ORDER` gets used, may fail?)
3. Upload step 2 back to the UV17 (modified `BLOCK_ORDER` gets used, maybe cause damage?)

Things don't behave properly.

In order to make this correct, you need to store something on the object *instance* and not modify the mutable copy attached to the *class* like you're doing. The instance of the list on the UV17 is the same instance for it and all subclasses that do not override it.

Copying it and re-setting it on the radio instance here would be one way, but that's obscure and confusing. Probably better would be to make the `baofeng_uv17._download()` method take a parameter of `extra_blocks=...` or something.

It sounds like this is not actually implemented currently and we're never uploading the cal block right? If so, can we just remove that from this patch for the time being and work on it in a follow-on?

kk7ds (commented):
This looks better, except that I'd like to have the revisions squashed into the base patch and tagged with the proper bug reference there. I think it's also worth splitting the uv17pro base refactor out into its own patch. However, I can do that git-fu for you before I merge. I'll leave this here for your final word and comment on the cal block thing but should be able to get this in tomorrow I hope.

chirp/drivers/radtel_rt620.py
OK2MOP: You are probably right that it could generate issues for possible child classes. Again it was workaround for a single file driver, now I can patch it correctly. In the meantime I have found out the UV17 driver suffers same issue with Banks tab, so it needs to be patched anyway. So I will create 2 more PRs (one for UV-17Pro driver, one for UV-17 driver) and then rebase this PR on them

OK2MOP: It seems that my local copy of BLOCK_ORDER which should have prevented that got deleted some time during the edits and only the comment remained at that place. But it would be better to fix it by other means.

kk7ds: This is a common thing in python that trips people up. Assignment is by reference. Consider this example:
```
class Foo:
    mylist = [1, 2, 3]

fooA = Foo()
fooB = Foo()

fooA.mylist.append(4)
print(fooB.mylist)
```
If you run that, it prints:
```
[1, 2, 3, 4]
```
Because there is only one copy of `mylist` and it's stored on the _class_, not copied for each instance. That's okay for global non-mutable stuff, but you have to be careful. This is why I say it would be better to just pass the block list to the download/upload function. But, if you're going your `BLOCK_O_READ` route, you need to be careful to copy it on the local object instance.

OK2MOP: I know it is a class variable, actually I am planning on using 
`BLOCK_O_READ = baofeng_uv17.UV17.BLOCK_O_READ + [2] # back up calibration`
This should be OK as it creates a new instance, right?

kk7ds: Yes, that will do it :)

OK2MOP:
Hello, I have tested read/write several times on my radio and re-read updated configuration in stock CPS without an error. 

What is not fixed (not fixable?):
- Inherited error message from UV-17 driver when switching to Settings tab:
`ERROR: Request for struct element name not in dict_keys(['code'])`
I was not able to locate in UV-17 driver what is the root cause there as there is no line number to find. I suspect some part of in-memory structure description is missing in both radios. But it does not seem to influence anything regarding the settings I have tried.
- Sometimes the download or upload fails if the display turned off on the radio in the meantime, second try is OK. I suppose this is the behaviour of radio in general (maybe audio in K connector generates some garbage on serial port).

chirp/drivers/radtel_rt620.py
```diff
@@ -0,0 +1,426 @@
...
+
+    def remove_extra_items(self, group):
+        tmp = RadioSettingGroup(group.get_name(), group.get_shortname())
+        for element in group:
+            name = element.get_name()
+            if not isinstance(element, RadioSetting):
+                tmp.append(self.remove_extra_items(element))
+            elif name not in self.REMOVE_SETTINGS:
+                tmp.append(element)
+        return tmp
```
kk7ds, line 234: I meant to mention this when I commented on the missing group thing, but.. this is a lot of work to just exclude one setting. Can we just add another class flag (like `_has_gps`, etc) and skip this setting in the base class instead of having to filter this out? If so, we can do a cleanup after the fact.

OK2MOP: Sorry, I was away from computer over the weekend until now. Of course, as I said, it was adapted from the code for UV-110D, and at the moment when I did not want to touch the original driver. I as you have already merged this, I will look into what can be done and make a PR later during the first December week

PR #1177, opened 2024-11-27
Issue 1067 Initial ICOM IC-M710 driver

kg7hf:
This is the initial driver to fix issue: [#1067](https://chirpmyradio.com/issues/1067) for the ICOM IC-M710 HF Marine SSB radio. 

This is the basic driver to program the radio channels.  
 - global settings are not supported
 - special memory banks are not full supported 
 - channel power settings is not supported


chirp/drivers/icf.py
```diff
@@ -505,6 +505,9 @@ def _clone_to_radio(radio):
             LOG.debug("Waiting for clone result...")
             time.sleep(0.5)
 
+    if radio.ignore_clone_result() == True:
+        return bytes(b'\x00')
```
kk7ds, line 509: This method returns a boolean, so please don't break that here. Just `return True` here please.

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,227 @@
...
+#seekto 0xE020;
+struct {
+  bbcd rx_freqL[4];   
+  u8 modeL;
+  char nameL[7]; 
+  bbcd tx_freqL[4];  
+  bbcd rx_freqH[4];   
+  u8 modeH;
+  char nameH[7]; 
+  bbcd tx_freqH[4];
```
kk7ds, line 58: This looks to me like you're combining odd and even memories instead of just making this align with the radio itself. Instead of this and this `if (number % 2) == 0` thing below, can't you just do this?
```
struct {
  bbcd rx_freq[4];
  u8 mode;
  char name[7];
  bbcd tx_freq[4];
} memory[466];
```
Then you just have `memory[number]` as the actual object for each given memory...

Perhaps a leftover vestige of trying to process the radio frames from the radio that can be optimized?

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,227 @@
...
+    _model = "\x162\x00\x01"  # 4-byte model string
+    _endframe = ''
+    _ignore_clone_ok = True
+
+    _memsize = 0xEFC0 
+    _ranges = [(0xE000, 0xEFC0, 32)]                 # Ranges of the mmap to send to the radio
+
+    _num_banks = 10              # Most simple Icoms have 10 banks, A-J
+    _bank_index_bounds = (0, 99)
+ #   _bank_class = IcomBank
```
kk7ds, line 83: I know this is a draft, but please do remove commented-out code from the version we expect to merge.

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,227 @@
...
+            mem.mode = MODULATION_MODES[int(_mem.modeL)]
+            #mem.power = POWER_LEVELS[int(2)]     # TODO, have to find where this is in the record!
+            mem.offset = int(_mem.tx_freqL)      # if the frequency is above 30MHz (166666665 or FF,FF,FF,FF), it's simplex
+            if int(_mem.tx_freqL) < 30000000:
+                 mem.duplex = 'split'
+                 mem.comment = 'TX: ' + str((int(_mem.tx_freqL)/1000000)) + ' MHz'
+            else:
+                mem.duplex = 'off'      
+                mem.comment = ''
+                _mem.tx_freqL.set_raw('\xFF\xFF\xFF\xFF')
```
kk7ds, line 171: There is lots of leftover bad practice from the python2 days in the tree, but please pass `bytes` to `set_raw()` and not strings.

kk7ds (commented):
Instructions for a development environment, including how to run tests is here:

https://www.chirpmyradio.com/projects/chirp/wiki/DevelopersPython3Environment

Probably best to try to get them all happy instead of just relying on github doing it :)

chirp/drivers/icf.py
kg7hf: Done

chirp/drivers/icm710.py
kg7hf: Done

chirp/drivers/icm710.py
kg7hf: Done

chirp/drivers/icm710.py
kg7hf: Done

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,178 @@
...
+        mem.mode = MODULATION_MODES[0]
+        if (_mem.mode > -1) and (_mem.mode < 7):
+            mem.mode = MODULATION_MODES[int(_mem.mode)]
+
+        # if the frequency is above 30MHz
+        # (166666665 or FF,FF,FF,FF), it's simplex
+        mem.offset = int(_mem.tx_freq)
+        if int(_mem.tx_freq) < 30000000:
+            mem.duplex = 'split'
+            mem.comment = 'TX: ' + \
```
kk7ds, line 145: The comment is for the user, you're not allowed to coerce it to some value you want it to be. Some drivers can store the comment themselves, but since you can't chirp will store the comment in the metadata when it saves a `.img` file. So you should just remove this.

chirp/drivers/icf.py
```diff
@@ -505,6 +505,9 @@ def _clone_to_radio(radio):
             LOG.debug("Waiting for clone result...")
             time.sleep(0.5)
 
+    if radio.ignore_clone_result() is True:
```
kk7ds, line 508: No need for the `is True`.. Some linters will complain if you do this, and some only if you compare with `==` but either way, please just test the truthiness, don't require it to be exactly `True`.

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,178 @@
...
+    # Store details about a high-level memory to the memory map
+    # This is called when a user edits a memory in the UI
+    def set_memory(self, mem):
+        # Get a low-level memory object mapped to the image
+        _mem = self._memobj.memory[int(mem.number)]
+
+        _mem.name = mem.name.ljust(7)[:7]
+
+        if mem.empty == 1:
+            _mem.rx_freq.set_raw(bytes(b'\xFF')*4)
```
kk7ds, line 169: a `b'foo'` is already a `bytes`. Doing this just wastes cpu cycles. Just `b'\xFF' * 4` is good. Also, please use spaces around operators.

But otherwise, thanks for converting these to `bytes` 👍

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,178 @@
...
+        rf.has_offset = False
+        rf.has_name = True
+        rf.has_bank = False
+        rf.has_bank_names = False
+        rf.has_tuning_step = False
+        rf.has_ctone = False
+        rf.has_cross = True
+        rf.has_infinite_number = False
+        rf.has_nostep_tuning = True
+        rf.has_comment = True
```
kk7ds, line 81: This should be `False` since you're not able to store a user's comment. Setting it to `False` means chirp will just handle it for you at the file layer.

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,178 @@
...
+
+MEM_FORMAT = """
+
+#seekto 0xE020;
+struct {
+  bbcd rx_freq[4];
+  u8 mode;
+  char name[7];
+  bbcd tx_freq[4];
+}memory[232];
```
kk7ds, line 39: Looks better thanks :)

chirp/drivers/icm710.py
kg7hf: I see. Is there a better way to show what the TX frequency is when the radio is in split mode? Maybe I should explore VFO A/B as a way?

kk7ds: As I mentioned on the ML, `split` duplex stores the TX frequency in the offset field. So if `tx_freq != rx_freq`, you need to set `mem.offset = tx_freq` and `duplex='split'`. Also, it looks like you're only supporting `duplex='split'` and `duplex='off'` but not 'duplex=''`. The latter means "simplex" (i.e. same TX and RX frequency). The test framework assumes simplex for its test channel, which is one reason you're failing a bunch of tests. Any user familiar with CHIRP will assume they can set memories in simplex mode.

kg7hf: Yeah, I've been banging my head on how to get the unit tests to pass.  I think the radio works like this:
if RX frequency is set and TX is FFFFFF, it's receive only, like WWV
if RX = TX, it's simplex
if RX != TX but valid tx, it's duplex.  In reality, all channels are duplex, but I think we can hide that from the user.  In the ICOM software you explicitly have to set both RX and TX frequencies.  Further, some of the global settings will prevent certain memory channels to be forced simplex depending on the setting.  I don't quite know how that works yet. 
So if duplex ='' then set TX = RX; if duplex = 'split' set the offset, and if duplex = 'off' set the field raw to FFFFFFFF

kk7ds: Right, this is a very common arrangement in lots of radios. The job of the driver is to adapt between CHIRP's model and the radio one. So in this case, you _must_ act like the radio has a specific simplex mode by just setting the tx and rx freq the same.

kk7ds:
Getting closer! You can ignore the "% same" errors in the PR checks - that's just because your driver is so short.

kg7hf:
> Getting closer! You can ignore the "% same" errors in the PR checks - that's just because your driver is so short.

I'm pretty stuck on the test_sync_out error at this point. Maybe time to put it down and have a food coma. Happy Thanksgiving

kk7ds:
It's because your `_ranges` excludes most of the start of the file, which means the test string is never uploaded to the radio. I'll modify the test to consider that possibility.

chirp/drivers/icm710.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # Store details about a high-level memory to the memory map
+    # This is called when a user edits a memory in the UI
+    def set_memory(self, mem):
+        # Get a low-level memory object mapped to the image
+        _mem = self._memobj.memory[int(mem.number)]
+
+        _mem.name = mem.name.ljust(7)[:7]
+
+        if mem.empty == 1:
```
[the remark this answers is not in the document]
kk7ds: Ah, `1 == True` is true. Either way, best to make this a truthiness test like all the other code. I will fix.

kk7ds:
I'm planning to merge this when the test run finishes, which I assume you're okay with. If not, put down the turkey and speak up :)

kg7hf:
> I'm planning to merge this when the test run finishes, which I assume you're okay with. If not, put down the turkey and speak up :)

I just did a quick test on the radio, it appears to function, good clones.  Merge! Thanks for the help.

PR #1178, opened 2024-11-28
Move STEPS in baofeng_uv17Pro to UV17Pro class, fixes #11716

OK2MOP:
As we have discussed in https://github.com/kk7ds/chirp/pull/1176#pullrequestreview-2466055160, this should be separate

2. Newest master has been used
3. Single commit
4. Added a feature in tracker with ID #11716 to have it referenced here


OK2MOP:
It makes me also wonder if `ga510.py`  driver has the steps incorrectly defined as it lists "STEPS" and "LIST_STEPS" on global level but does not reference them anywhere. But there are multiple drivers in this file and it is not clear with which one it should be associated.

OK2MOP:
I will force-push updated code with the Dan Smith's suggestion of adding

`def get_mapping_models(self):
    return []
`

PR #1179, opened 2024-11-28
Issue #11717 mitigation in UV17 driver, extra blocks for read now possible in subclass

OK2MOP:
As we have discussed in https://github.com/kk7ds/chirp/pull/1176#discussion_r1861323606, this should be solved in superclass, i.e. here instead in the RT620 driver:
- The has_bank mitigation has to be done here as the driver is also affected (Bug #11717 mitigation)
- Added extra class variable BLOCK_READ to allow additional block to be read and stored in image (for future calibration upload from backup with big warning on inherited drivers if necessary)

1.   tox style and fast-driver tests passed
1.   Newest master has been used
1.   Single commit
1.   Mitigates bug in tracker with [ID #11717](https://chirpmyradio.com/issues/11717) added to have it referenced here

### DIFF
--- a/chirp/drivers/baofeng_uv17.py
+++ b/chirp/drivers/baofeng_uv17.py
@@ -75,7 +75,7 @@ def _download(radio):
     status.msg = "Cloning from radio..."
     radio.status_fn(status)
 
-    for block_number in radio.BLOCK_ORDER:
+    for block_number in radio.BLOCK_O_READ:
         if block_number not in memory_map:
             # Memory block not found.
             LOG.error('Block %i (0x%x) not in memory map: %s',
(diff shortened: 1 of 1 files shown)


chirp/drivers/baofeng_uv17.py
```diff
@@ -158,6 +158,10 @@ class UV17(baofeng_uv17Pro.UV17Pro):
 
     MODES = ["FM", "NFM"]
     BLOCK_ORDER = [16, 17, 18, 19,  24, 25, 26, 4, 6]
+
+    # Add extra read blocks (e.g. calibration block 2) to the end here:
+    BLOCK_O_READ = BLOCK_ORDER
```
kk7ds, line 163: This does not make a copy. If I modify `BLOCK_ORDER` I'm also modifying `BLOCK_O_READ`. You need to do something like this:
```
BLOCK_O_READ = list(BLOCK_ORDER)
```

kk7ds:
Mentioning the bug number in the commit won't close it, you need to use exactly `Fixes: #11717` for it to work.

I'll push up a fix for these issues and the method override I mentioned in the bug so we can get this merged quickly for your other stack.

OK2MOP:
OK, thanks and sorry for the complications. I was also not sure I should write Fixes as it was mitigation, Next time I will be more careful. So "fixes #11716" is also not correct and I should update the text in the second issue to make it correct?

I have the code modified for RT-620 driver to force push, but I will need ~2-3h to get home and test it with the radio besides updating the source that everything works correctly. Ideally I will do it after the merge

PR #1181, opened 2024-11-28
Add TH-350US version. Fixes #10887 and #11487.

gwholtha1:
TYT [US software version](https://www.twowayradioforum.com/t/help-programming-a-tyt-350th-analogue-ht/8916/11) sets up the TH-350 for channel names that are 8 characters long. Current CHIRP TH-350 driver expects 5 characters for channel names, which results in garbled names when programming. This update creates a TH-350US option to properly program US-configured radios.

### DIFF
--- a/chirp/drivers/th350.py
+++ b/chirp/drivers/th350.py
@@ -260,6 +260,8 @@ class Th350Radio(BaofengUVB5):
         "VFO3": -1,
     }
 
+    NAME_LENGTH = 5
+
     _memsize = 0x1000
 
     @classmethod
@@ -284,7 +286,7 @@ class Th350Radio(BaofengUVB5):
         rf.can_odd_split = True
         rf.valid_skips = ["", "S"]
         rf.valid_characters = CHARSET
-        rf.valid_name_length = 5
+        rf.valid_name_length = self.NAME_LENGTH
         rf.valid_bands = [(130000000, 175000000),
                           (220000000, 269000000),
                           (400000000, 520000000)]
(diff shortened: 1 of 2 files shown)


chirp/drivers/th350.py
```diff
@@ -393,3 +393,210 @@ def _get_memobjs(self, number):
...
+      u8 uhflopwr6;
+      u8 uhflopwr7;
+    } test;
+    """
+
+    def process_mmap(self):
+        self._memobj = bitwise.parse(self.MEM_FORMAT, self._mmap)
+
+    # Override uvb5 set_memory to allow for 8 character channel names
+    def set_memory(self, mem):
```
kk7ds, line 559: Instead of reproducing `set_memory()` entirely here, can you please modify the one in the base class to use a class variable? Something like:
```
class Th350Radio:
    NAME_LENGTH = 5
    def get_settings(self):
        ...
        rf.name_length = self.NAME_LENGTH
    def set_memory(self, mem):
        ...
        # Use self.NAME_LENGTH instead of hard-coded 5

class Th350USRadio(Th350Radio):
    NAME_LENGTH = 8
    # The rest of your stuff, but exclude set_memory() and get_features()
```

kk7ds:
Is there any way to determine the US/non-US model from the clone procedure? If so, It would be best to detect the right model when the user selects TH-350, or at least prevent them from choosing the US model when they have the non-US, etc.

chirp/drivers/th350.py
```diff
@@ -393,3 +393,210 @@ def _get_memobjs(self, number):
...
                 len(filedata) == (cls._memsize + 0x30))
+
+# US version has channel names that are 8 characters long
+
+
+@directory.register
+class Th350USRadio(Th350Radio):
+    """TYT TH-350US"""
+    VENDOR = "TYT"
+    MODEL = "TH-350US"
```
kk7ds, line 404: Can you leave `MODEL` alone and use `VARIANT = 'US'` please?

gwholtha1:
Thanks for the review. I added a class variable for the channel name length. This required adding the class variable to the uvb5 driver in order to modify the set_memory function, but I don't have that radio to test it.

I also changed the th350 driver to use VARIANT instead of MODEL.

Great suggestion to determine US vs. non-US model from the clone procedure, but I'm not sure how to do that. I checked the identifier string at the start of the radio memory, but it's the same for both variants.

chirp/drivers/th350.py
```diff
@@ -393,3 +395,166 @@ def _get_memobjs(self, number):
...
         return (filedata.startswith(b"TRI350 Radio Program data") and
                 len(filedata) == (cls._memsize + 0x30))
+
+# US version has channel names that are 8 characters long
+
+
+@directory.register
+class Th350USRadio(Th350Radio):
+    """TYT TH-350US"""
+    VENDOR = "TYT"
```
kk7ds, line 405: Not necessary to have this here, but won't hurt either.

kk7ds:
I just pushed to:

1. Rebase
2. Squash the fixups into the main commit (per the PR guidelines)
3. Fix the bug linkage (Needs to be `Fixes #xxx`, per the PR guidelines) in the commit message
4. Snipped out the extraneous `VENDOR` while I was in there.

Thanks!

gwholtha1:
Thank you!

gwholtha1:
@kk7ds This also fixes #11422 and probably #9343.

kk7ds:
Can you comment on those? I'll close as appropriate, thanks!

gwholtha1:
Done. Thank you!

PR #1183, opened 2024-11-29
uv17: Fix missing setting warning message

kk7ds:
The UV17-based radios don't have a 'name' property for DTMF memories,
and the existing code had a try..except around the attempt to decode
it in the base pro driver. This is normally a good way to go, but
the bitwise infra logs an error, so this changes it to just check for
and skip if there's no name property.

Related to #11699


kk7ds:
As mentioned in #1176

PR #1195, opened 2024-12-13
Driver for Kenwood TK690, TK790, and TK890

Patronics:
See Issues #3363 and #1093.
Supports all channel-specific settings, bank arrangement, and customization of front-panel button functions. Tested on TK-690, and by comparison with files exported from KPG-44D for all 3 models. Based on code started by Pavel Milanes and Tom Hayward in 2016.


kk7ds (changes requested):
Thanks for picking this up. I haven't made it through all of this, but a lot of the comments here apply to more than just the quoted lines, so I'll stop here and iterate after changes. I know almost all of this is not your doing, but thanks for taking charge of it. I can help with some of these if you want/need.

Patronics:
Thanks for all the feedback Dan! 

I've resolved almost all of your suggestions, and things definitely are a bit easier to follow now! I assumed that I shouldn't touch what looked to be standard boilerplate logic that worked fine, but as you pointed out, I inherited code that was unnecessarily complicated (and not particularly standard functionality for chirp either, as I'd assumed it was), after implementing your suggestions I definitely understand more about many aspects of this driver's logic that I'd taken for granted without looking at them closely until now. 

I'll need to dig into the logic a bit more to figure out the intended purpose of the vaguely named variables in your remaining requested changes, as those are another aspect of the code that I was given that was already functional, albeit definitely not very clear what's going on for a lot of it. I probably won't have time for that tonight, but in the meantime I welcome any other suggestions for tweaks and improvements to this driver, whenever you get a chance to look over it again. Thanks!

chirp/drivers/tk690.py
```diff
@@ -0,0 +1,1340 @@
...
+        # building the data for the memmap
+        fdata = b""
+
+        for k, v in data.iteritems():
+            # posible bad data
+            if k == 0:
+                k = 1
+                raise errors.InvalidValueError(
+                    "Invalid bank value '%d', bad data in the image? \
+                    Triying to fix this, review your bank data!" % k)
```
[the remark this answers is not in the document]
Patronics: I'm looking through this _prep_data function, as far as I can find it's not actually invoked by the code at all? It also had some clearly nonsensical aspects to it earlier on, trying to write some values to 0x7000 while the memory in this radio only goes up until 0x60A0. Unless you see any valuable functionality in it that I'm overlooking, making any of this logic worth keeping around somewhere, I'm thinking I'll just get rid of the whole function (then there's no need to worry about any confusing logic within it lol)

Patronics:
I've also done some more testing, and found that the way that the radios encode certain settings seems to be different from one model to another (specifically, the location of the bit that specifies wide/narrow bandwidth is different between the TK790 and the 890, and there's no option for setting the bandwidth at all in the 690). How would you recommend storing this in the MEM_FORMAT to minimize duplicated logic? Presumably it's not really optimal to duplicate the whole MEM_FORMAT when they're so close to identical overall?

chirp/drivers/tk690.py
```diff
@@ -0,0 +1,1340 @@
...
+            c = 1
+            for i in v:
+                fdata += bytes([k]) + bytes([c]) + bytes([k - 1]) + bytes([i])
+                c = c + 1
```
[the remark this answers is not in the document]
Patronics: See reply to above regarding the prep_data function, I might remove it altogether unless you think any of it is worth keeping around

chirp/drivers/tk690.py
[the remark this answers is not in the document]
kk7ds: Yeah, I don't see it called either. Frustrating :(

One pattern you'll see often is people copying a driver and modifying it to work for a different radio. That ends up with all kinds of orphaned crap, and this could be a similar situation. With hundreds of drivers to look at, it's quite hard to notice and reject those things so plenty slipped in over the years. There's a job in "PR checks" that tries to spot overly-similar drivers now, but it's not perfect of course.

So yes, please nuke it if it's not useful/used.

chirp/drivers/tk690.py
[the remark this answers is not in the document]
kk7ds: Yes please. Frustrating that I spent time trying to figure it out...

kk7ds:
There was a question about `MEM_FORMAT` differences among models which either got deleted or wiped away by a push. If that's still a question, please restate it here and I can offer some suggestions.

chirp/drivers/tk690.py
```diff
@@ -0,0 +1,1211 @@
...
+        for buttonName in ASSIGNABLE_BUTTONS:
+            _fullHeadWarning = ""
+            if buttonName in FULL_HEAD_ONLY_BUTTONS:
+                _fullHeadWarning = "  (If Equipped)"
+            rs = RadioSetting(
+              buttonName,
+              "Configured function for "+buttonName+" button"+_fullHeadWarning,
+              RadioSettingValueMap(
+                    BUTTON_FUNCTION_LIST,
+                    self._memobj.button_assignments[buttonName]))
```
kk7ds, line 1128: This should use the non-deprecated `current_index=` pattern instead of the implicit "current value" one it's using here. You should be getting a `FutureWarning` in the log for it.

The newer pattern is safer and allows the UI to handle cases where bad data is stored in the radio so we can highlight the broken setting instead of crashing right here because the value is out of range for the array and thus load no settings.

chirp/drivers/tk690.py
```diff
@@ -0,0 +1,1211 @@
...
+            button_assignments.append(rs)
+        return group
+
+    def set_settings(self, settings):
```
kk7ds, line 1132: This also looks like copy pasta from another driver, handling multiple overly-complex setting possibilities that aren't actually in the list returned from get_settings. I'm guessing L1161 is the only part of this that is actually used.

kk7ds (commented):
Looking closer, thanks.

chirp/drivers/tk690.py
Patronics: Hmm, can you give an example of using RadioSettingValueMap with a current_index pattern? I couldn't find any examples of it in any other drivers in the repo, and my attempt at it (the following code: 
```python
RadioSettingValueMap(
                    BUTTON_FUNCTION_LIST,
                    current_index=next((i for i, (_, value) in enumerate(BUTTON_FUNCTION_LIST) if value == self._memobj.button_assignments[buttonName]), None)))
```
Produced the following error: `TypeError: RadioSettingValueMap.__init__() got an unexpected keyword argument 'current_index'`
I tried using the same logic but with "user_option" instead of "current_index", and that works, but provides the same FutureWarning as my original logic did.

I tried to open a few of the test images for other radios that used RadioSettingValueMap for their settings to see if any of them handled it differently, (specifically, I've tested the "radtel rt490", the "th7800", and "puxing px888k"), and all three of them produced the same FutureWarning about using current_index instead of value for safety.

Patronics:
Okay, I've pushed adjustments to again resolve almost all of your requested changes, aside from the one I had a follow-up question on regarding the futureWarning with RadioSettingValueMap. 

Also regarding the deleted message about MEM_FORMAT, I realized the layout difference that I thought I'd observed was just due to me miscounting the bytes in a hex editor, so no need to worry about that 🙂

chirp/drivers/tk690.py
kk7ds: Sorry, I read this as `RadioSettingValueList` because of the pattern, saw the warning locally when I tried, and didn't realize it was the map. I need to make the same change to map. As you can (probably) see, map uses list as its parent, which is why it's complaining but there's no way to pass the index in.

That's fine, I'll fix that, this, and the others after merge.

chirp/drivers/tk690.py
```diff
@@ -0,0 +1,1100 @@
...
+    return struct.pack(">BBHB", ord(cmd), 0x8F, addr, RX_BLOCK_SIZE_H)
+
+
+def _handshake(radio, msg="", full=True):
+    """Make a full handshake"""
+    if full is True:
+        # send ACK
+        radio.pipe.write(ACK_CMD)
+
+    # receive ACK
```
kk7ds, line 231: I think you said you were going to work on more settings after this. When you do, would you mind removing all these pointless comments? Any comment that is meaningful (i.e. explaining why instead of what) should stay, but these "receive ACK", "check ACK", "send data" etc comments are just annoying.

kk7ds:
We need to squash this down to one commit since we don't need to persist the iterative fixes here. I'll do that and then merge this. the unit failure is because repeaterbook is broken now, so unrelated.

kk7ds:
(also rebased and fixed the commit message to properly close the referenced bugs)

PR #1199, opened 2024-12-15
uv5r: Expose rich "K1 connection" error

kk7ds:
These radios use a K1 connector and suffer from the usual fitment
issues. Expose the rich error with "more info" about this problem.

Related to #11738


kk7ds:
Unit failure is repeaterbook (still broken)

PR #1200, opened 2024-12-15
Fix missing EOL in FRS/GMRS stock config

kk7ds:


kk7ds:
Repeaterbook is broken

PR #1201, opened 2024-12-15
RSValueMap lazy init

kk7ds:
- **Fix missing EOL in FRS/GMRS stock config**
- **Fix RadioSettingValueMap deprecated current-value**
- **tk690: Remove stray print() debug**
- **Make RadioSettingValueMap support lazy init**


kk7ds:
Repeaterbook still broken

PR #1202, opened 2024-12-16
fix/enable selecting banks with spacebar

Patronics:
I noticed that while in the bank editor you can visually select checkboxes with the spacebar (as well as navigating with the arrow keys), the change from selecting in this way doesn't actually take effect in the drivers. There's logic for handling it, but only in the context of setting bank indexes in the few drivers that support that. Now checkbox selection works as expected, whether clicked or selected via the spacebar. 🙂


chirp/wxui/bankedit.py
```diff
@@ -222,6 +222,12 @@ def _index_changed(self, event):
         value = event.GetString()
         if isinstance(self._col_defs[col], ChirpBankIndexColumn):
             self._change_memory_index(self.row2mem(row), int(value))
+        # if activating a checkbox with spacebar, treat it like a click
+        elif isinstance(self._col_defs[col], ChirpBankToggleColumn):
```
kk7ds, line 226: The name of this method (`_index_changed`) is no longer correct with this change as it's handling more than just a changing index. Can you rename it so it's more obvious?

Patronics:
I just realized, with this change we can effectively remove the _memory_changed callback, instead pointing it to the same function. I'll test that a bit more to make sure there's no unexpected side effects, then open another pull request with that adjustment

PR #1204, opened 2024-12-16
bankedit.py: remove redundant memory_changed function

Patronics:
Now that the _cell_changing method handles both cases, memory_changed is redundant.

### DIFF
--- a/chirp/wxui/bankedit.py
+++ b/chirp/wxui/bankedit.py
@@ -111,7 +111,7 @@ class ChirpBankEdit(common.ChirpEditor):
         self._memory_cache = {}
 
         self._grid.Bind(wx.grid.EVT_GRID_CELL_CHANGING, self._cell_changing)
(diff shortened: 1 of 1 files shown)


kk7ds:
This fails on a radio with an actual index:
```
ERROR: <function ChirpBankEdit._cell_changing at 0x1223ab370> raised unexpected exception
  File "/Users/dan/Documents/chirp.git/chirp/wxui/common.py", line 642, in run_safe
    return fn(*args, **kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/bankedit.py", line 225, in _cell_changing
    self._change_memory_index(self.row2mem(row), int(value))
ValueError: invalid literal for int() with base 10: ''
```
You can use the `tests/images/Icom_IC-2820H.img` to test. Haven't looked at why, but should be trivial.

PR #1206, opened 2024-12-17
Update wxPython to >=4.2.1 - fixes #10577

socram8888:
wxPython 4.2.1 fixes the dependency on the broken attrdict3 package, as per their issue tracker: https://github.com/wxWidgets/Phoenix/issues/2225#issuecomment-1596190592


kk7ds:
We can't bump the min requirement on Linux above 4.0 because major still-supported distros still only provide 4.0.x. Also, 4.2.1 is broken on macos and I want to keep the formal windows/mac builds synchronized (see #1085 for more info).

socram8888:
I see. For you information, though, current requirements.txt will fail to install on Windows 10 with Python 3.12 with the dreaded missing attrdict dependency :/

PR #1207, opened 2024-12-19
Add basic support for Baofeng BF-T20D and W31D - closes #11671

socram8888:
WIP attempt to add support for those two walkies. More MIGHT actually work but those are the ones I can personally try.

Missing:
 - Setting walkie talkie settings (already identified the location and settings, just not done yet)
 - Round TX/RX frequency based on firmware version. Some older are only able to use frequencies rounded to the nearest 5KHz, instead of 2.5KHz and 6.25KHz like the v1.05 I have.

Working:
 - Edit all channel settings


chirp/drivers/baofeng_digital.py
```diff
@@ -0,0 +1,492 @@
...
+        for step, cmd in enumerate(preamble, 1):
+            self.pipe.write(cmd[0])
+            if self.pipe.read(len(cmd[1])) != cmd[1]:
+                raise errors.RadioError(
+                    f'Radio replied incorrectly to preamble step {step}')
+
+    def _clean_buffer(self):
+        self.pipe.timeout = 0.001
+        self.pipe.read(256)
+        self.pipe.timeout = 0.1
```
kk7ds, line 206: Have you tested this on Windows? On windows, setting the timeout flushes the buffer, which probably doesn't matter here but it sometimes surprises people when they do it in the middle of a conversation with the radio.

chirp/drivers/baofeng_digital.py
```diff
@@ -0,0 +1,492 @@
...
+        if value in (0x0000, 0xFFFF):
+            return ('', None, 'N')
+
+        # Parse lower 14 bit as BCD
+        bcd_value = (
+            1000 * (value >> 12 & 0x3) +
+            100 * (value >> 8 & 0xF) +
+            10 * (value >> 4 & 0xF) +
+            (value & 0xF)
+        )
```
kk7ds, line 287: Can you re-use `bitwise.bcd_to_int()`?

kk7ds: ...and I assume you can't use a bcd bitwise type for some reason? I figured you were setting it as a non-BCD value somewhere  but I don't see it.

chirp/drivers/baofeng_digital.py
```diff
@@ -0,0 +1,492 @@
...
+                elif rx_tmode == 'DTCS':
+                    mem.rx_dtcs = rx_tvalue
+
+                if tx_tmode == 'Tone' and rx_tmode == '':
+                    mem.tmode = 'Tone'
+                elif (
+                    tx_tmode and
+                    tx_tmode == rx_tmode and
+                    tx_tvalue == rx_tvalue
+                ):
```
kk7ds, line 350: Please don't waste the two lines here. I know some people like black formatting, but I hate it :)

chirp/drivers/baofeng_digital.py
```diff
@@ -0,0 +1,492 @@
...
+
+    def _pack_dtcs(self, code, polarity):
+        return (
+            0x8000 |
+            (polarity == 'R' and 0x4000) |
+            0x1000 * (code // 1000 % 10) |
+            0x100 * (code // 100 % 10) |
+            0x10 * (code // 10 % 10) |
+            code % 10
+        )
```
kk7ds, line 399: Same for these, surely you can re-use the bitwise functions and still or-in your flag right?

chirp/drivers/baofeng_digital.py
```diff
@@ -0,0 +1,492 @@
...
+    POWER_LEVELS = [
+        chirp_common.PowerLevel("Low", watts=0.5),
+        chirp_common.PowerLevel("High", watts=2.00),
+    ]
+    _serial_id = 0x04
+    _proto = PROTO_D
+
+
+# Disabled because UNTESTED, though it should work
+# @directory.register
```
kk7ds, line 482: Let's not put this into the tree like this and instead attach a .py file to the (or a new) tracker issue  and let people test it as a module.

chirp/drivers/baofeng_digital.py
```diff
@@ -0,0 +1,492 @@
...
+        rf.has_rx_dtcs = True
+        rf.has_bank = False
+        rf.has_name = False
+        rf.has_tuning_step = False
+        rf.can_odd_split = True
+        rf.has_cross = True
+        rf.memory_bounds = (1, self._proto.chan_count)
+        rf.valid_bands = self.VALID_BANDS
+        rf.valid_duplexes = ["", "+", "-", "split"]
+        rf.valid_modes = ['FM', 'NFM', 'DIG', 'DN']
```
kk7ds, line 146: `DN` means Yaesu system fusion, so unless this radio implements that, you may not use `DN` here. Also `DIG` is reserved for the digital setting in HF radios. It's not a good use of the mode, but that's what it has meant in CHIRP in the past, so it needs to remain.

CHIRP needs a consistent view of the world in terms of what modes mean, so that you copy a memory from one radio to another, we can select the proper mode *or* reject it if it's not compatible. For example, if you open an FT3DR image and copy a `DN` memory out of it and paste into this radio, CHIRP should reject that paste because your radio doesn't support actual YSF DN mode (assuming the above).

What digital mode is this actually?

chirp/drivers/baofeng_digital.py
socram8888: Yep, I'm using Windows in fact for all development. No issues.

chirp/drivers/baofeng_digital.py
socram8888: I have absolutely no idea. Their sales pitch says it's GMSK, and that is all the information I have.

It supports 25KHz and 12.5KHz using the same modulation, and I don't know any  standard mode that uses that modulation with variable width?

kk7ds: Yeah there has to be more than just GMSK. Also I doubt it's 25kHz or 12.5kHz wide, but rather those are the channel spacing recommendations, but who knows for sure I guess.

How about we just support FM,NFM and make the digital mode a .extra toggle for now (i.e. FM+extra['digital'] means "25kHz wide in digital mode" ? Sort of like we do for radios with the "scrambler" modules - we don't consider that a different mode, just a flag on an existing one. If we figure out that it is actually something else (or a second radio starts supporting it) we can add it into the mode list?

kk7ds:
Other than the DIG/DN thing, this looks good now I think. What was the answer on the commented-out second model? Did you get it tested or was I not clear about what I was suggesting?

For the module, if you leave it out of this file and we commit it, then you can attach a copy of the file with the second radio definition (or with only that) to an issue and people can use Help->Load from issue to try it out. If you're not sure that it works, the module route would be safer.

socram8888:
Not sure if you'd agree but regarding the mode, I think it would make sense to actually add a new one, as there's already a myriad of other Baofeng walkies using it (see that linked Wiki page where they've certified seven, and they actually had even more listed in the CPS source code which might be for either domestic use or non-US).

I'll try to figure out looking through the documentation if they've given it any particular name, other than a generic "GMSK mode". I might also try contacting the email on the back of the manual (`wangjianhui(a)baofengradio.com`) to see if they can clarify a bit.

Regarding the BF-T20D I thought you meant I should create a new issue and did it: https://chirpmyradio.com/issues/11742

socram8888:
The page I'm mentioning is https://www.sigidwiki.com/wiki/Baofeng_GMSK, which also lists all the FCC reports for the different models that share the same digital mode.

I'll replace it for now with a digital flag, but I really think that should have its own mode given they're different models that are interoperable between them.

kk7ds:
> The page I'm mentioning is https://www.sigidwiki.com/wiki/Baofeng_GMSK, which also lists all the FCC reports for the different models that share the same digital mode.

Ack, thanks, I was just reading the FCC report on the W31D. Even that wiki page says it's not clear what exactly is going on.
 
> I'll replace it for now with a digital flag, but I really think that should have its own mode given they're different models that are interoperable between them.

Thanks, I'll do some thinking about how we might make it easier to support some bespoke modes like this without having to enum everything. I'm sure the future will bring more of this than less :/

socram8888:
Just to be 100% sure - you'd prefer to wait until there's a better way to add new modes, yes? Or add that digital switch in the extras?

kk7ds:
Sorry, no, I'd rather add it with the digital extra toggle now and ponder the other mode thing after we merge. I've spent a little time thinking about it so I have some ideas but with the holidays I haven't tried to implement anything. We should get this merged with just the extra toggle if you're willing though.

I was going to push to the build queue tonight for a build tomorrow, FYI.

kk7ds:
Okay, cool. Also you still have the other BaofengBFT20D model in there. If you're not sure about it, I can snip that out before I merge and attach as a module to the other bug to avoid advertising that it's fully supported. Should I remove it?

kk7ds:
Okay, well, I assume that's the plan for the moment anyway. So I pushed to snip that off and rebase so I can merge this for tomorrow.

kokroo:
@socram8888 I can help you test the Baofeng T20D.
Is there any way I can read the data from the Radio and send it to you?
I checked the FCC reports and it seems like the internals on the T20D and W31D are probably the same.

kk7ds:
@kokroo Instructions and a module to try here: https://www.chirpmyradio.com/issues/11742

PR #1212, opened 2024-12-30
Added support for Kenwood TS-790E. Possibly fixes #10215

danielegozzi:
# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/kenwood_live.py
+++ b/chirp/drivers/kenwood_live.py
@@ -47,6 +47,7 @@ THF6_MODES = ["FM", "WFM", "AM", "LSB", "USB", "CW"]
 
 
 RADIO_IDS = {
+    "ID007": "TS-790E",
     "ID019": "TS-2000",
     "ID009": "TS-850",
     "ID020": "TS-480_LiveMode",
--- a/chirp/drivers/ts790e.py
+++ b/chirp/drivers/ts790e.py
@@ -1,0 +1,264 @@
+import logging
+import threading
+
+from chirp import chirp_common, directory, errors
+
+from chirp.drivers.kenwood_live import KenwoodLiveRadio, NOCACHE, RADIO_IDS
+
+LOG = logging.getLogger(__name__)
+
+TS790E_DUPLEX = ["", "split", "+", "-"]
+TS790E_TMODES = ["", "Tone"]
+TS790E_SKIP = ["", "S"]
+
+TS790E_MODES = {
+    "LSB": "1",
+    "USB": "2",
+    "CW":  "3",
+    "FM":  "4",
+    "CWN": "7",
+}
+TS790E_MODES_REV = {val: mode for mode, val in TS790E_MODES.items()}
+
+TS790E_TONES = list(chirp_common.OLD_TONES)
+TS790E_TONES.remove(69.3)
+TS790E_TONES.remove(97.4)
+TS790E_OFFSETS = ["", "+", "-"]
+
+TS790E_BANDS = [
+    (135000000, 173999900),
+    (300000000, 949999900),
+    (1200000000, 1304957000),
+]
+
+TS790E_SHIFTS = [600000, 1600000, 35000000]
+
+TS790E_SPECIALS = ["P1", "P2", "P3", "A1", "A2", "A3", "C1", "C2", "C3"]
+
+
+def get_band(freq):
+    i = 0
+    for band in TS790E_BANDS:
+        if freq >= band[0] and freq <= band[1]:
+            return i
+        i += 1
+    return None
+
+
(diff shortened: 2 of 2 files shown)


danielegozzi:
A couple of words on the _why_.
This_ couldn't work as a simple edit to kenwood_live.py because the radio hangs with the spurious commands when the driver is trying to determine LAST_DELIMITER and LAST_BAUD. The radio needs to be powered down to reset serial communications after receiving the mangled detection commands.

Indeed this started as a copy of ts850.py. I know this is not the way new drivers are meant to be implemented, but most of the globals are different and a lot of the implementation is different. The 850 driver would need a lot of conditionals to work for both the 790E and the 850 models.
I have no objections in replicating the ts850.py copright heading if needed, although the current file has no heading because that's what I understood from the contribution guidelines.

As I have the actual TS-790E (yay for me!) I could test a substantial rewrite with a kenwood_live.py that somehow better resets the serial communication, but I don't know how you would coordinate the tests covering the other Kenwood models.

kk7ds:
If we were to just move the globals into the class so you could override them then the delta would be small right? That's a more common pattern in the the other drivers. What conditionals would be needed in the base driver?

```
class TS790:
    ...
    MODES = {
        "LSB": "1",
    ... etc
```

The issue is not one of license but of code duplication. Many of the drivers in CHIRP started as copies of another instead of proper re-use. When it came time for a small number of us to refactor them all for python3, it became clear how much duplication there was in the tree (requiring duplicated effort to fix).

danielegozzi:
If you think it's beneficial I could try the refactoring in kenwood_live.py (slight risk of breaking other drivers).
The duplication detected by the automated tests, though, is between the new driver and the ts850.py one. I don't know if the refactoring can help with that. You see, the 850 is quite similar but with subtle differencies in part of the binary messages and the radio capabilities. So the common part is mostly boilerplate, comments and such.

danielegozzi:
Ok, after some polishing (meaningful, not pretextual) even the duplication test from the ts850.py driver is passing.

I could delete the command() function further reducing the new code by applying the refactoring that kenwood_live.py was suggesting in `TODO: This global use of LAST_DELIMITER breaks reentrancy`. I explored this path in [my temporary branch](https://github.com/danielegozzi/chirp/tree/kenwood_live_refactor) currently excluded from the PR, as it has a different scope. If you think that is a good idea, I could open a PR for merging the kenwood_live.py refactoring, then we can come back to the ts790e.py driver.

kk7ds:
> Ok, after some polishing (meaningful, not pretextual) even the duplication test from the ts850.py driver is passing.

I'm a bit confused.. if this is very similar to the ts850, why not refactor _that_ to allow swapping out the common bits and reduce this down to just a very small delta?

That said, these are really not that big so especially with the small dedup you did which gets this under the line, I won't hold you hostage further.

> I could delete the command() function further reducing the new code by applying the refactoring that kenwood_live.py was suggesting in `TODO: This global use of LAST_DELIMITER breaks reentrancy`. I explored this path in [my temporary branch](https://github.com/danielegozzi/chirp/tree/kenwood_live_refactor) currently excluded from the PR, as it has a different scope. If you think that is a good idea, I could open a PR for merging the kenwood_live.py refactoring, then we can come back to the ts790e.py driver.

I haven't looked at your branch (yet), but fixing that would be great. I don't have any kenwood_live HF radios but I have plenty of VHF ones to test with so between the two of us we should be able to say something authoritative about a fix.

Up to you whether that goes first or this does. I'm okay with merging this now if you want (I'll wait until you answer), but if you'd rather get that done first, that's fine.

danielegozzi:
> I'm a bit confused.. if this is very similar to the ts850, why not refactor _that_ to allow swapping out the common bits and reduce this down to just a very small delta?

That's because the common parts are now mostly function declarations (same names, same parameters), comments, decorations and non-functional stuff. The different part is most of the actual logic. However, the test is passing now, so no big deal.

> I haven't looked at your branch (yet), but fixing that would be great. I don't have any kenwood_live HF radios but I have plenty of VHF ones to test with so between the two of us we should be able to say something authoritative about a fix.
> 
> Up to you whether that goes first or this does. I'm okay with merging this now if you want (I'll wait until you answer), but if you'd rather get that done first, that's fine.

I opened PR #1213 for the kenwood_live.py refactoring, so we can fix that before including this new driver for the TS-790E - I think that the 790E around have probably already been programmed in the mid-nineties, so there is no hurry.

danielegozzi:
Rebased on master after merging #1213 , thus deleting the duplicated code for the command() instance method. This was tested with my TS-790E and still works.

kk7ds:
Thanks!

PR #1218, opened 2025-01-05
icf520: Add support for Icom IC-F520, IC-F521, IC-F620, and IC-F621

N7XSD:
Fixes #11490
Fixes #6505

# CHIRP PR Guidelines


chirp/drivers/icf520.py
```diff
@@ -0,0 +1,1643 @@
...
+    # FIXME: BEGIN Ugly Hack
+    # Force driver to use one bank that contains all memories.
+    # This allows us to work on other parts of the driver before
+    # the bank model is complete.
+
+    def save_mmap(self, filename):
+        self._memobj.bank[0].capacity = 256
+        for i in range(1, self._num_banks):
+            self._memobj.bank[i].capacity = 0
+        super().save_mmap(filename)
```
kk7ds, line 714: Can you do this in `sync_in()` or maybe in `process_mmap()`? Doing it as part of save feels wrong. If this is needed for your code to work, doesn't this require a save and re-open for this to take effect anyway?

Also, doing it in `process_mmap()` will mean that it runs on load, so that if someone opens a newer image with an older chirp someday, that chirp will flatten things the way it needs.

Also-also, you may or may not want to implement this kind of radio as banks. The bank model and editor really doesn't work the way it needs to for these kind of commercial radios with dynamic zones or tables. In the TK-8180 driver (which has a similar dynamic split) I decided to make it expose sub-devices for each of the zones, which I think works better.

chirp/drivers/icf520.py
```diff
@@ -0,0 +1,1643 @@
...
+
+@directory.register
+class ICF520Radio(icf.IcomCloneModeRadio):
+    """
+    Icom IC-F520 VHF Land Mobile Radio
+    """
+
+    VENDOR = "Icom"
+    MODEL = "IC-F520"
+    ALIASES = [ICF621Radio, ICF620Radio, ICF521Radio]
```
kk7ds, line 676: Please don't do it this way. This `ALIASES=` mechanism is from long ago, before we had the metadata trailer on the files and thus couldn't tell these apart. Now these should just be registered subclasses.

Also, if these really all use the same model info, is there some other way we can tell the difference between them? From the look of your code I can download any of these radios using any of the aliases. We should try to thwart that, even if we have to do something nasty like checking in `process_mmap()` and raising an error.

I'm guessing that also means that `valid_bands` covers all the union of all the variants instead of properly enforcing the limits for a given variant, right? That should be properly set on each subclass and enforced so you can't put VHF channels into a UHF radio, etc.

kk7ds (commented):
I don't have time for an in-depth review of this, but here are a few things from a skim. Thanks! I think I have an F621 somewhere, I might be able to try this.

chirp/drivers/icf520.py
N7XSD: Yes.  I moved the code to `process_mmap()`.

I'll have a look at the TK-8180 driver.  For now, treating the radio as having a single bank works.

I still hope to write a bank model... someday.

chirp/drivers/icf520.py
N7XSD: I'll remove the `ALIASES=` thing.

Frankly, I don't know if they use the same model info.  There are a lot of bits sent by the radio I haven't mapped out.

`decode_model()` from `icf.py` looks useful but the IC-F520, etc. have a different header (?) size.  With only two radios (IC-F521 and IC-F621-2), I'm not confident I could create a reliable test.

Yes, `valid_bands` covers the union of all the variants.  (Icom's CS-F500 software doesn't even do that much checking.)  I'm reluctant to base model checking on the frequencies in a download since a radio mis-programmed by CS-F500 could leave a radio un-programable by chirp.  For a while I was using `MODEL = "IC-F500 series"`.  But decided to see what would happen with aliases (which I didn't know were deprecated.)

When one of these radios is programmed with a bad frequency, the display flashes when that channel is selected.  The rest of the channels are still usable.  Not much of an error message but it's something.

I'll spend some more time looking at the header.

kk7ds: > Frankly, I don't know if they use the same model info. There are a lot of bits sent by the radio I haven't mapped out.

In that case let's please leave the others unavailable (we can put a module in an issue so someone can test and let us know). People get super mad when they see a radio is "supported", but it, and find out that it's never been tested :)

> `decode_model()` from `icf.py` looks useful but the IC-F520, etc. have a different header (?) size. With only two radios (IC-F521 and IC-F621-2), I'm not confident I could create a reliable test.

Okay, but those two use the same model_info? There must be other bits in the image that tell the software which actual model it is.

> Yes, `valid_bands` covers the union of all the variants. (Icom's CS-F500 software doesn't even do that much checking.) I'm reluctant to base model checking on the frequencies in a download since a radio mis-programmed by CS-F500 could leave a radio un-programable by chirp. For a while I was using `MODEL = "IC-F500 series"`. But decided to see what would happen with aliases (which I didn't know were deprecated.)
> 
> When one of these radios is programmed with a bad frequency, the display flashes when that channel is selected. The rest of the channels are still usable. Not much of an error message but it's something.

Okay, but CHIRP works differently from the factory software and we should have one model per variant with a consistent set of parameters. Even if the radio "handles" invalid frequencies, CHIRP should try its best to mirror what the radio actually supports.

N7XSD: Got it!  I'll drop all references to _should work_ variants and find a way to distinguish between the IC-F521 and the IC-F621-2.

Thanks for the explanation.  This helps me a lot.

kk7ds:
LMK when I should look at this again. It's got a bunch of unsquashed commits and merges and it's not based on current master so I assume you're still working on it :)

N7XSD:
Still not getting git and github.  I'll probably just delete everything (except icf520.py) and create a fresh copy of the repository... again.

Is there a work flow you recommend to keep me out of trouble in the future?

N7XSD:
Yes, please.  I decided to limit the driver to the IC-F621-2.  I can come back to the IC-F521 later.

The Real World is taking up a lot of my time lately.  This way, I can contribute _something_ useful instead of letting it sit all lonesome on my hard disk.  :-)

chirp/drivers/icf520.py
```diff
@@ -0,0 +1,1604 @@
...
+                self._get_trunk_key_group(),
+                self._get_display_group(),
+                self._get_set_mode_group(),
+                self._get_common_group()))
+
+    def set_settings(self, settings):
+        """Accepts the top-level RadioSettingGroup returned from get_settings()
+        and adjusts the values in the radio accordingly. This function expects
+        the entire RadioSettingGroup hierarchy returned from get_settings()."""
+        for element in settings:
```
kk7ds, line 1426: Also check out `MemSetting` which takes a symbolic path into the memory object and can set boolean, integer, string, and map types directly so you don't have to manually do everything in `set_settings()` yourself.

kk7ds (commented):
This is okay enough to include now. I'm going to snip out the extra log message and rebase/squash as mentioned. Thanks!

PR #1222, opened 2025-01-10
Added alias UV-5R 8W Tripower

ZombieEater:
`Related to #11763 ` Alias added to model BF-F8HP for UV-5R 8W Tripower as suggested by Baofeng Support.

This UV-5R model, 8 Watt version, Tripower matches with BF-F8HP as recommended by Baofeng Support to be used in Chirp.
I just added the alias to be shown in the models list.

### DIFF
--- a/chirp/drivers/uv5r.py
+++ b/chirp/drivers/uv5r.py
@@ -2025,12 +2025,16 @@ class TDUV5RRadio(chirp_common.Alias):
     MODEL = "TD-UV5R TriPower"
 
 
+class UV5R8WTri(chirp_common.Alias):
+    VENDOR = "Baofeng"
+    MODEL = "UV-5R 8W TriPower"
+
 @directory.register
 class BaofengBFF8HPRadio(BaofengUV5R):
     VENDOR = "Baofeng"
     MODEL = "BF-F8HP"
     ALIASES = [RT5_TPAlias, ROGA5SAlias, UV5XPAlias, TSTIF8Alias,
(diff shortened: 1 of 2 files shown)


ZombieEater (commented):
Ignore this commit, mine bad.

chirp/drivers/uv5r.py
```diff
@@ -2025,12 +2025,16 @@ class TDUV5RRadio(chirp_common.Alias):
...
+    VENDOR = "Baofeng"
+    MODEL = "UV-5R 8W TriPower"
+
 @directory.register
 class BaofengBFF8HPRadio(BaofengUV5R):
     VENDOR = "Baofeng"
     MODEL = "BF-F8HP"
     ALIASES = [RT5_TPAlias, ROGA5SAlias, UV5XPAlias, TSTIF8Alias,
-               TenwayUV5RPro, TSTST9Alias, TDUV5RRadio]
+               TenwayUV5RPro, TSTST9Alias, TDUV5RRadio, UV5R8WTri]
```
kk7ds, line 2037: This is the old way to do in-app aliases, which is not how they should be done anymore (I need to update the developer docs probably).

However, if this is just a straight "choose X instead of Y" as it seems from your posts on the matter, please do this in `chirp/share/model_alias_map.yaml` which affects the website list only.

kk7ds:
Also, please do not propose against the `build_queue` branch. Post propose against `master` which I push to `build_queue` when it's time to build.

PR #1226, opened 2025-01-13
uv17pro: Fix exposing static banks on some models

kk7ds:
Previously in commit:5d295a8b we added a static bank model to show
the static mappings between memories and "banks" in these radios.
However, only some of the more advanced models have this functionality,
not all the radios that inherit from UV17Pro. This fixes that to only
apply this model to the UV17ProGPS and BF-F8HP-PRO models.

Further, in commit:d1f8504e we fixed the bug where the UV17 models
inherited this behavior to disable it for them. Those overrides are
no longer necessary, so this removes them.

Related to #11642

### DIFF
--- a/chirp/drivers/baofeng_uv17.py
+++ b/chirp/drivers/baofeng_uv17.py
@@ -492,14 +492,6 @@ class UV17(baofeng_uv17Pro.UV17Pro):
 
         self.set_memory_common(mem, _mem)
 
-    def get_features(self):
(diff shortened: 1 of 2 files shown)


kk7ds:
Okay @KC9HI I think this cleans up my mess.

KC9HI:
Yes @kk7ds I think this is what was needed. Thanks!

PR #1237, opened 2025-01-21
TK690: Add test frequencies to settings, and fixes

Patronics:
Fixes power level handling, and fix band limits for the tk690(K3)

These changes all work, except that currently it fails one of the automated tests on the new settings pane for setting test frequencies, as the unused test frequencies default to the value in memory of `0xFFFFFFFF` (in the official UI, this is displayed as a blank box), which is otherwise outside of the RadioSettingsValueInteger's acceptable range (and, given the value's encoding in memory is formatted as `lbcd`, I suspect that sure if other aspects of validation or handling for that value would fail as well. The Developer's browser tab displays the value as `166666665`, not the `4294967295` one would expect from a pure binary handling)). 

Is there a recommended way to handle a default blank value like that? Should I just use RadioSettingsValue (or RadioSettingsValueString) and give it a validation callback to handle the default value? Or is there some other Settings datatype that would be more appropriate?

I suppose another option is to just remove the testing frequencies from settings for now, in order to get the other fixes merged in, and allow advanced users who need to adjust the testing frequencies to do so from the developer browser without a dedicated settings tab.

Thanks!

### DIFF
--- a/chirp/drivers/tk690.py
+++ b/chirp/drivers/tk690.py
@@ -150,6 +150,11 @@ struct {
   char soft_ver[5];
 } properties;
 
+struct {
+  lbcd rxfreq[4];       // 00-03
+  lbcd txfreq[4];       // 04-07
+} test_frequencies[16];
+
 #seekto 0x40A0;
 struct {
   char name[16];
@@ -544,8 +549,8 @@ class Kenwoodx90(chirp_common.CloneModeRadio, chirp_common.ExperimentalRadio):
     BAUD_RATE = 9600
     VARIANT = ""
     MODEL = ""
-    POWER_LEVELS = [chirp_common.PowerLevel("High", watts=45),
-                    chirp_common.PowerLevel("Low", watts=5)]
+    POWER_LEVELS = [chirp_common.PowerLevel("High", watts=110),
+                    chirp_common.PowerLevel("Low", watts=45)]
     MODES = ["NFM", "FM"]  # 12.5 / 25 Khz
     _name_chars = 8
     _group_name_chars = 8
@@ -877,7 +882,7 @@ class Kenwoodx90(chirp_common.CloneModeRadio, chirp_common.ExperimentalRadio):
         if mem.power is None:
             _mem.power = 0
         else:
-            self.POWER_LEVELS.index(mem.power)
+            _mem.power = self.POWER_LEVELS.index(mem.power)
 
         _mem.wide = self.MODES.index(mem.mode)
 
@@ -1044,7 +1049,10 @@ class Kenwoodx90(chirp_common.CloneModeRadio, chirp_common.ExperimentalRadio):
                                            "Basic Settings")
         button_assignments = RadioSettingGroup("button_assignments",
                                                "Configurable Button Functions")
-        group = RadioSettings(basic_settings, button_assignments)
+        test_frequencies = RadioSettingGroup("test_frequencies",
+                                             "Test Frequencies")
+        group = RadioSettings(basic_settings, button_assignments,
+                              test_frequencies)
 
         # Basic Settings:
         rs = RadioSetting("head_type",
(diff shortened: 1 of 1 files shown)


kk7ds:
> Fixes power level handling, and fix band limits for the tk690(K3)
> 
> These changes all work, except that currently it fails one of the automated tests on the new settings pane for setting test frequencies, as the unused test frequencies default to the value in memory of `0xFFFFFFFF` (in the official UI, this is displayed as a blank box), which is otherwise outside of the RadioSettingsValueInteger's acceptable range (and, given the value's encoding in memory is formatted as `lbcd`, I suspect that sure if other aspects of validation or handling for that value would fail as well. The Developer's browser tab displays the value as `166666665`, not the `4294967295` one would expect from a pure binary handling)).

In the actual WX UI these show up as yellow boxes, which is how we handle "out of range" things (when done properly). The tests just don't expect "broken" settings in a default test image. I don't really want to make the tests ignore that though because in almost all situations it means the driver author hasn't implemented something correctly, or a later change broke parsing an existing image.

> Is there a recommended way to handle a default blank value like that? Should I just use RadioSettingsValue (or RadioSettingsValueString) and give it a validation callback to handle the default value? Or is there some other Settings datatype that would be more appropriate?

You could just pick `(range[1] + range[0]) / 2` as the default if the thing isn't set and let chirp set those if they're not (or pick some default test frequency for each model).

> I suppose another option is to just remove the testing frequencies from settings for now, in order to get the other fixes merged in, and allow advanced users who need to adjust the testing frequencies to do so from the developer browser without a dedicated settings tab.

Yeah, I mean, that's easier to just not care about these. Without the full test mode implemented in chirp (please no :P) these aren't really that important anyway, and someone that is going to use them is pretty much going to need the OEM software anyway. You could also just leave out-of-range values out of the settings entirely and let people poke some default value into place to make it show up if necessary.

Patronics:
Thanks for the suggestions, I think I've found a pretty elegant solution by combining a couple of them. Now the settings pane uses volatile checkboxes to enable any inactive test frequencies, setting them to the default middle-of-range value you suggested.

It irks me a little bit that there's no way to reset the test channels to the disabled state with the default value in the UI, but that's a very minor nitpick overall, and of course there's no harm in just not using a given test frequency if you don't need it, or digging into the file with a hex editor if you really wanted to. Everything works, these changes should be ready for merging in now :)

kk7ds:
Hmm, yeah, I mean that works I guess, but it violates the goal of not having the structure really change at runtime. It used to be a hard requirement in the old UI, not as much in the new one, but moving to a very dynamic structure kinda rubs me the wrong way.

Do you have a strong desire/need for the test frequencies to be exposed in chirp? If so, what's the use-case? Another way to do it would be a separate sub-device for those as specials, which would support them being deleted or not just like regular memories. TBH, the Pavel-esque approach of implementing these group-based Kenwood commercial radios as all one address space with bank support doesn't feel right to me, which is why the drivers I've written (tk8180, tk8160, tk280) like this expose the groups as sub-devices. It's much more work, but feels more right-er, and that would also lend itself (potentially) to the test frequencies if you want.

But yeah, I guess I'm just interested to know what the strong need is for these in chirp, given the lack of test mode...

Patronics:
> But yeah, I guess I'm just interested to know what the strong need is for these in chirp, given the lack of test mode...

The radio has a built-in test mode when you hold PF1 when powering it on, which allows it to test transmission and receiving at the programmed test frequencies with the panel's UI. 

Honestly I don't know all the details of how the computer-connected test mode is supposed to be used, but the guide I've been following for retuning the tk-690h for use with ham frequencies suggests using the radio's test frequencies (measuring voltages at a test point while set to each frequency, and adjusting some potentiometers), and programming a few additional ones, hence my interest in adding support for managing those test frequencies in chirp, especially since I was already tweaking the driver a bit more to fix the band limits and power level handling anyway.

I'd say that while yes, the structure technically is more dynamic this way, it leads to very little change to the overall layout, and is equivalent to your suggestion of "let people poke a default value to make them accessible" except that this implementation makes that more accessible for users, without them needing to figure out the underlying data structure.

Also, regarding you suggestion of treating the testing frequencies as an additional sub-device of channels, while I suppose that theoretically would be doable, they are treated very differently than normal channels are, with no metadata associated with them, but instead simply a densely packed array of frequencies, so the normal method of deleting the channel wouldn't work with them, and we encounter the same issues with being unable to insert 0xFFFFFFFF into an lbcd data type.

kk7ds:
> The radio has a built-in test mode when you hold PF1 when powering it on, which allows it to test transmission and receiving at the programmed test frequencies with the panel's UI.
> 
> Honestly I don't know all the details of how the computer-connected test mode is supposed to be used, but the guide I've been following for retuning the tk-690h for use with ham frequencies suggests using the radio's test frequencies (measuring voltages at a test point while set to each frequency, and adjusting some potentiometers), and programming a few additional test frequencies for the new range, hence my interest in adding support for managing those test frequencies in chirp, especially since I was already tweaking the driver a bit more to fix the band limits and power level handling anyway.

Ah, okay, "panel test" mode I guess. Others have that too, I've just always used the OEM software for doing it since it's much easier. But, fair enough.

> I'd say that while yes, the structure technically is more dynamic this way, it leads to very little change to the overall layout, and is equivalent to your suggestion of "let people poke a default value to make them accessible" except that this implementation makes that more accessible for users, without them needing to figure out the underlying data structure.

What I meant was in the radio browser in developer mode ;)

> Also, regarding you suggestion of treating the testing frequencies as an additional sub-device of channels, while I suppose that theoretically would be doable, they are treated very differently than normal channels are, with no metadata associated with them, but instead simply a densely packed array of frequencies, so the normal method of deleting the channel wouldn't work with them, and we encounter the same issues with being unable to insert 0xFFFFFFFF into an lbcd data type.

Sure, just like the others. That sub device would just expose basically no features (no squelch modes, duplex, etc). Anyway, not that important I guess.

PR #1241, opened 2025-01-27
Fix sorting columns on linux

kk7ds:
Since we don't show the filter query on Linux, we need to also not
choke post-sort when we try to re-filter.

Fixes #11793


kk7ds:
Failures are from repeaterbook, unrelated.

PR #1248, opened 2025-02-01
fix retevis rb626 support

r2bln:
# Retevis RB626 fix

Retevis RB626 differs from RB26 in a way that RB626 has 32 memories total compared to 30 (GMRS) of RB26. This leads to a different offset of the settings part in the image blob and different constant values in a couple of places. Meaning the RB26 driver is not applicable "as-is".


chirp/drivers/retevis_rt21.py
```diff
@@ -196,6 +196,69 @@
...
+  u8 compander:1,      // Compander              C
+     unknown1:1,       //
+     highpower:1,      // Power Level
+     wide:1,           // Bandwidth
+     bcl:1,            // Busy Lock  OFF=0 ON=1
+     unknown2:3;       //
+  u8 reserved[3];      // Reserved               D-F
+} memory[32];
+
+#seekto 0x002D;
```
kk7ds, line 215: I know the pattern in this file is to just redefine all the memory maps, but.. from your description it sounds like the only differences here are `memory[32]` and the `#seekto` address, is that right?

If so, I think it would be better to parameterize the original and let each driver format it as appropriate. Here'd an example: [kk7ds/chirp@master/chirp/drivers/ft1d.py#L908](https://github.com/kk7ds/chirp/blob/master/chirp/drivers/ft1d.py#L908). That way any further changes don't have to be made in two places.

kk7ds (commented):
Thanks, just one suggestion inline. Could you also add a test image for this radio? I assume that it might have failed the settings tests and highlighted this earlier if we had one, and it would be good to have one as part of this change as well.

Also cc @kc9hi in case he has any concerns about this change.

chirp/drivers/retevis_rt21.py
r2bln: That is right. I will look into it.

kk7ds: Okay thanks. If you get stuck let me know and I will propose what I think you want and you can test to confirm.

kk7ds:
Nice, thanks. I'll leave this open a bit in case @KC9HI wants to comment, but looks good to me otherwise.

KC9HI:
@kk7ds, look OK to me. Thanks @r2bln.

kk7ds:
Sweet thanks!

PR #1250, opened 2025-02-03
icf520.py: Clarify display inhibit and tx inhibit

N7XSD:
For the user:
Display inhibit column name changed from "Inhibit" to "Mask", useful doc string was added, and enable choice changed from "Inhibit" to "Yes".

For the developer:
Variable "inhibit" changed to "disp_inhibit".  Similar changes to related variables.

Variable "inhibit_tx" changed to "tx_inhibit" to be consistent with "disp_inhibit" and "compander_inhibit".

Comment about using 0 MHz as an unused memory changed to reflect these changes.

### DIFF
--- a/chirp/drivers/icf520.py
+++ b/chirp/drivers/icf520.py
@@ -687,6 +687,7 @@ class ICF621_2Radio(icf.IcomCloneModeRadio):
         rf.valid_modes = list(MODES)
         rf.has_offset = True
         rf.has_tuning_step = False
+        rf.valid_tuning_steps = chirp_common.TUNING_STEPS
         rf.valid_duplexes = list(DUPLEX)
         rf.valid_bands = list(self._valid_bands)
         rf.can_odd_split = True
--- a/chirp/drivers/icf520.py
+++ b/chirp/drivers/icf520.py
@@ -232,9 +232,9 @@ struct {
 } dtmf_autodial[10];
 struct {
   char  name[10];
-  u32   inhibit:1,
+  u32   disp_inhibit:1,
         freq_rx:31;
-  u32   inhibit_tx:1,
+  u32   tx_inhibit:1,
         freq_tx:31;
   u16   rx_tone_off:1,
         rx_tone_digital:1,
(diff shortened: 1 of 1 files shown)


chirp/drivers/icf520.py
```diff
@@ -795,13 +795,17 @@ def get_memory(self, number):
...
             RadioSettingValueList(
-                INH_CHOICE, current_index=int(_mem.inhibit)))
-        _inhibit.set_doc("Inhibit")
-        memory.extra.append(_inhibit)
+                YES_CHOICE, current_index=int(_mem.disp_inhibit)))
+        _disp_inhibit.set_doc(
+            "Prevent the radio from displaying the channel.  This "
+            "prevents use of the channel but CHIRP will still download "
+            "the channel information.")
+        memory.extra.append(_disp_inhibit)
```
kk7ds, line 808: Thanks for the extra verbiage. Just so I understand, this is not intended to delete it from CHIRP, but just have it not show up as usable on the radio right? Most Yaesu radios have a "mask" bit that makes a channel disappear from the dial, and that's all they set when you "delete" something on the radio. It's frustrating because a factory reset doesn't even reset the contents of the memories, it just "masks" everything.

kk7ds (commented):
In playing with this using your test image, I realized you've got another bug. I can't cut and paste the first (for example) memory in the radio because it complains that the radio doesn't support 12.5kHz steps (but it must since the channel is stored). Can you confirm, and if so, just put 12.5 (and probably 6.25 at least?) in your `valid_steps`?

chirp/drivers/icf520.py
N7XSD: I'll create a ticket and pull request to fix valid_steps.

When the IC-F621 Tx frequency is set to 0, the memory is "deleted" and the radio does not display the channel.  Both `icf520.py` and CS-F500 agree on this behavior.  I don't remember what the radio will download of the old values from a "deleted" memory.  In `icf520.py`, I force the memory to default values for everything.

When a radio is "Masked" (inhibit in CS-F500), all memory settings are preserved by the radio but the channel will not be displayed (and can't be used).  The radio will download the channel and it is usable by `icf520.py` and CS-F500.

My choice of the word mask was mostly arbitrary.  I have a radio from the 1980's which allows a memory to be masked (hidden) or skipped (on scan).  Alas, that radio is not PC programmable.

N7XSD: Fixes #11811

I thought this should be a separate pull request but I didn't do it right.

kk7ds: Yeah, you really should not be using your master branch for PRs, but that's why. I just pushed a change to this to prefix the first commit with the driver module. Each of these changes gets emailed to the list when a build is made, and so driver-specific stuff really needs to be called out so people know it affects them.

PR #1258, opened 2025-02-09
Fix: windows wx requirements and remove of last __future__

Gornoka:
With these changes you can install chirp on windows in python3.13

also removes last import of __future__.

### DIFF
--- a/chirp/drivers/uvb5.py
+++ b/chirp/drivers/uvb5.py
@@ -13,8 +13,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-from __future__ import division
-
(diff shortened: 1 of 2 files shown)


kk7ds:
Yeah, thanks, but the the official builds are still on 3.10 across all the platforms for a variety of reasons. Since I'm the one that primarily works on the UI, I'm the one that has to deal with all the wxpython differences when we have a split. So you can install the newer version locally, but I want to keep the binary distributions consistent until I'm ready to bump them all. Here are some references to previous discussions:

https://chirpmyradio.com/issues/10577

#1085 
#1206

Happy to take the future commit separately though.

Gornoka:
Thanks for the clarification,
superseeded by #159.

PR #1263, opened 2025-02-15
update requirements.txt no issue wxPython on Linux 

Krusz-Beton:
there is no need to separate package versions between linux and windows anymore 
Issue https://github.com/wxWidgets/Phoenix/issues/2225 is closed now
**wxPython** in version **4.2.2** works fine on linux
I suggest you go back to the latest version of wxPython

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,4 @@
-wxPython>=4.0,<4.2.0 ; platform_system=="Linux" # See https://github.com/wxWidgets/Phoenix/issues/2225
-wxPython==4.2.0 ; platform_system!="Linux"
+wxPython
 pyserial
 requests
 pywin32; platform_system=="Windows"


kk7ds:
Please see https://www.chirpmyradio.com/issues/10577

wxPython 4.2.2 is still broken on macOS, and 4.0.x is still the distro version on a ton of supported linux systems. 4.2.0 is what is in the binary builds and so this needs to match those until that changes.

Krusz-Beton:
@kk7ds ok, you're right if **wxPython 4.2.2** still not working on  macOS, this PR makes no sense
by the way I had a case on Debian 11

- wxPython 4.1.x could not compile (slib error)
- wxPython 4.2.0 could not compile (as you know)
- wxPython 4.2 **.2** it compiled beautifully

kk7ds:
Yeah, FWIW, it's far better to use the distro-supplied wxpython package as there have been several cases where people have compiled wx and wxpython against their GTK and found issues, but if they use the one matched/tested by the distro the problems go away. Use the instructions on the site to install with pipx (which doesn't care about `requirements.txt` anyway) and you should be good without any need to compile anything.

I'll close this, as noted, thanks.

PR #1264, opened 2025-02-16
query sources: add mapy73.pl

Krusz-Beton:
added new source mapy73.pl

mapy73.pl has repeaters from several European countries (based on the old database from przemienniki.net). The database structure has been slightly improved and is ready for new types of repeaters. The update of FM Polska repeaters was successful, so we continue further development work on the new website.

Prepared several data exports from mapy73.pl especially for the chirp program.

In this Pull Request chip can import prepared data directly from the website (export data cached, changed after database update)

### DIFF
--- a/chirp/wxui/query_sources.py
+++ b/chirp/wxui/query_sources.py
@@ -30,6 +30,7 @@ from chirp.sources import radioreference
 from chirp.sources import repeaterbook
 from chirp.sources import przemienniki_net
 from chirp.sources import przemienniki_eu
+from chirp.sources import mapy73pl
 from chirp.wxui import common
 from chirp.wxui import config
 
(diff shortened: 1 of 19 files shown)


kk7ds:
I thought the new replacement for the old site was this: https://www.chirpmyradio.com/issues/11430

Do we need to keep that new one, the old one and this third one or will one encompass all that data?

Krusz-Beton:
I suggest you keep these 3 databases (the final data will be different, and the approach to chirp files is also different), but it's your project Dan and your decisions, I'm not imposing anything.

https://www.chirpmyradio.com/issues/11430 - I only have repeaters from Poland
przemenniki.net - the site is down, apparently someone is saving it, but I've already lost hope

Krusz-Beton:
Hello @kk7ds, do you need any update from me?  Is this code right for you? When do you plan release it?

chirp/wxui/query_sources.py
```diff
@@ -931,6 +932,89 @@ def get_params(self):
...
+        'si-fm',  # 'Slovenia FM'
+        'is-fm'  # 'Iceland FM'
+    ]
+    _options_text = [
+        'FM-Poland',
+        'Poland FM-LINK',
+        'Poland DMR',
+        'Poland C4FM',
+        'Poland DSTAR',
+        'Poland FM',
```
kk7ds, line 962: Why is there a "Poland FM" and a "FM-Poland" in this list? The UI makes it hard to know which to choose (perhaps it's obvious for people in the know, but...).

chirp/wxui/query_sources.py
```diff
@@ -931,6 +932,89 @@ def get_params(self):
...
+        self.SetSizer(vbox)
+        panel = wx.Panel(self)
+        vbox.Add(panel, 1, flag=wx.EXPAND | wx.ALL, border=20)
+        grid = wx.FlexGridSizer(2, 5, 0)
+        grid.AddGrowableCol(1)
+        panel.SetSizer(grid)
+
+        # Option
+        self._option_text = wx.Choice(panel, choices=self._options_text)
+        self._option_text.SetStringSelection(self._options_text[0])
```
kk7ds, line 991: It seems a bit weird to have a query that just dumps the whole database for each of these regions (or sub-databases?)... All our other sources allow for filtering at least band or purpose. What's the thinking behind just providing this single option and dumping everything all the time?

kk7ds (commented):
Please rebase and squash out your incremental "test" commits, per the guidelines. This should be landed as a single commit in my repo.

Krusz-Beton:
@kk7ds   thank you for the code review

1. you are right, this "FM-Poland" is misleading. I propose "FM-Poland network" this is the  local successor of echolink, because "Poland FM-Poland" doesn't sound good

2. Our database has a few more extra objects, related to repeters. But in a way you are right, this is an export specifically for chirp, fact is static file and generated every time when the database was changed (server optimization, database changes every few days, user queries every few minutes). Filters have been omitted here, just like the iPhone has only a few options that work, compared to the multitude of options in Android. The emphasis here is on simplicity for less initiated users (FM-Link filtering for 2m, 4m, 6m, 10m bands makes no sense, adding 2m repeaters on SSB requires changing filter chirp sourcecode). Most transceivers supports simultaneously 2m and 70cm band, so I give it to them in all within the mode.  Those radios that support only one band, e.g. baofeng 888s has only 8 channelsso, the user anyway has to select manually  a few repeters lines ctr+c and paste them into the radio ctr+v, which is very easy in chirp. 

3. rebase and squash, done. I hope everything will be ok, sorry for my shortcomings

chirp/wxui/query_sources.py
Krusz-Beton: 1. you are right, this "FM-Poland" is misleading. I propose "FM-Poland network" this is the local successor of echolink, because "Poland FM-Poland" doesn't sound good

chirp/wxui/query_sources.py
Krusz-Beton: 2. Our database has a few more extra objects, related to repeters. But in a way you are right, this is an export specifically for chirp, fact is static file and generated every time when the database was changed (server optimization, database changes every few days, user queries every few minutes). Filters have been omitted here, just like the iPhone has only a few options that work, compared to the multitude of options in Android. The emphasis here is on simplicity for less initiated users (FM-Link filtering for 2m, 4m, 6m, 10m bands makes no sense, adding 2m repeaters on SSB requires changing filter chirp sourcecode). Most transceivers supports simultaneously 2m and 70cm band, so I give it to them in all within the mode. Those radios that support only one band, e.g. baofeng 888s has only 8 channelsso, the user anyway has to select manually a few repeters lines ctr+c and paste them into the radio ctr+v, which is very easy in chirp.

PR #1270, opened 2025-02-24
Add Radtel RT-493 Driver

paulober:
rt-493: Added Driver

This PR adds a driver for the Radtel RT-493.
I implemented it as fake `CloneModeRadio` as it wants a request per two memory channels when reading but it's simpler to just dump all at once due to the layout of settings (the official app does it the same way).

I'm currently not sure what to put into the `valid_bands` because the European version is 430 MHz to 440 MHz but the device can do 400 MHz to 470 MHz.


chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+    """Raw read from the radio device"""
+    data = bytearray()
+    try:
+        while len(data) < amount:
+            chunk = radio.pipe.read(amount - len(data))
+            if not chunk:
+                LOG.debug("No response from radio")
+                raise errors.RadioNoContactLikelyK1()
+            data.extend(chunk)
+            time.sleep(0.002)  # 2ms delay
```
kk7ds, line 60: Why? This makes no sense if it's not delaying some transmission to the radio that it's not ready for. In general drivers that have made extensive use of timing have been the most problematic, especially across platforms, and in almost all cases I've removed it to the betterment of everything. Can you please explain why you think this is necessary especially with a long timeout set.

These "`rawrecv`" type wrappers are not very good practice IMHO, so I'm wondering if maybe you just copied this from another driver that has a bunch of this stuff in it, thinking it was a good idea? If so, let's please remove it. With a long timeout (which is fine) a regular `pipe.read(N)` will read N bytes if available or timeout, and there's really no reason at low speeds like this to loop back for multiple trips.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+LOG = logging.getLogger(__name__)
+
+COMMAND_ACCEPT = b"\x06"
+CHARSET_HEX = "0123456789ABCDEFabcdef"
+
+
+def _clean_buffer(radio):
+    radio.pipe.reset_input_buffer()
+    radio.pipe.reset_output_buffer()
+    time.sleep(0.002)
```
kk7ds, line 40: You'll need to explain why you think these resets are necessary, especially with the following junk read (which is fine). These sorts of things behave very differently on different platforms (mainly between windows and the other POSIX platforms) and these tend to be problematic.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+    data = bytearray()
+    try:
+        while len(data) < amount:
+            chunk = radio.pipe.read(amount - len(data))
+            if not chunk:
+                LOG.debug("No response from radio")
+                raise errors.RadioNoContactLikelyK1()
+            data.extend(chunk)
+            time.sleep(0.002)  # 2ms delay
+    except Exception:
```
kk7ds, line 61: You're raising `RadioNoContactLikelyK1` and then swallowing it here to not show it to the user. If this radio has a (potentially problematic) K1 connector then that's good to have, but you should allow it to actually make to the UI, otherwise it's pointless. You should also `LOG.error` or `LOG.exception` the actual failure in order to be useful for debugging.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+        raise errors.RadioError("Invalid response from the radio")
+
+    return bytes(data)
+
+
+def _rawsend(radio, data):
+    """Raw send to the radio device"""
+    try:
+        radio.pipe.write(data)
+        time.sleep(0.002)
```
kk7ds, line 77: Same here, this makes no sense.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+    except Exception:
```
kk7ds, line 78: And same need to log the actual problem.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+    """Get data from the radio """
+    hdr = _rawrecv(radio, 4)  # Read 4 byte header
+    data = _rawrecv(radio, length)  # Read data
+
+    # DEBUG
+    LOG.info("Response:")
+    LOG.debug(util.hexprint(hdr + data))
+
+    c, a, resp_length = struct.unpack(">BHB", hdr)
+    if a != int.from_bytes(addr, byteorder="big") \
```
kk7ds, line 92: Why are you doing this? If the byte order isn't actually big, use the little-endian `struct.unpack` format so you can compare directly.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+    LOG.debug("Sent magic sequence")
+    ack = _rawrecv(radio, 2)
+    LOG.debug("Received magic sequence response")
+    if len(ack) != 2 or ack[1:2] != COMMAND_ACCEPT:
+        if ack:
+            LOG.error("Received: Len=%i Data=%s"
+                      % (len(ack), util.hexprint(ack)))
+        raise errors.RadioError("Radio refused to enter programming mode")
+
+    radio.pipe.reset_input_buffer()
```
kk7ds, line 114: Same here, please don't do this unless there's some really (really) good and well-tested reasoning.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+
+def _get_memory_address(channel_id):
+    block = (channel_id - 1) // 16           # Determines high byte
+    offset = ((channel_id - 1) % 16) * 0x10  # Determines low byte
+
+    return (block, offset)
+
+
+def _read_block(radio, channel_id):
+    address = bytes(_get_memory_address(channel_id))
```
kk7ds, line 149: This makes me think you're pulling small chunks of memory and assembling them into an image instead of downloading all of memory. The way this should work is CHIRP should download the entire memory image of the radio and store it sequentially in the image file as it is stored in memory. If you need to only write back portions of the image on upload, then that's fine, but we should capture a full image from the start to end of memory during download. Without that, we'll have really disastrous problems once we realize we need to read another section in the middle to support some new feature.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+
+    try:
+        # 52=read | 20 = hex(32 bytes) = length of 2 channels
+        # + settings without header
+        cmd = b"\x52" + address + b"\x20"
```
kk7ds, line 154: `struct.pack` please

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+
+        if address != b"\x00\x00":
+            _rawsend(radio, COMMAND_ACCEPT)
+            response = _rawrecv(radio, 1)
+            if response != COMMAND_ACCEPT:
+                raise errors.RadioError("Radio refused to read block at %04s"
+                                        % util.hexprint(address))
+
+        return data
+    except Exception:
```
kk7ds, line 166: Again, you're swallowing your specific error at L162 and re-raising here, as well as any of the specific things that `_rawsend` and `_rawrecv` are raising, like L119 which would be quite important.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+                                % util.hexprint(address))
+
+
+def _encode_frequency(freq_mhz):
+    """
+    Encodes a frequency in MHz into 4 bytes.
+    Converts from:
+        MHz → Byte3 Byte2 . Byte1 Byte0
+    Example:
+        431.350 → 00 50 13 43
```
kk7ds, line 193: This looks like scaled BCD to me, like many other drivers, and you've got it defined as such in your memory map. Why are you hand-encoding it and using `set_raw` to force it into memory?

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+    :param byte: The byte to decode (integer)
+    :return: Dictionary with settings
+    """
+    return {
+        # (1=Wide, 0=Narrow)
+        "W/N Mode": "Wide" if ord(byte) & 0x10 else "Narrow",
+        "Compander": bool(ord(byte) & 0x80),  # (1=ON, 0=OFF)
+        "Scramble Mode": ord(byte) & 0x07,  # (0=OFF, 1-7=Scramble levels)
+        "Special QT/DQT": "A" if ord(byte) & 0x20 else
+        ("B" if ord(byte) & 0x40 else "OFF"),
```
kk7ds, line 243: Please use bitwise bitfields.. the purpose of that is to avoid this hand-masking which tends to be error-prone and is much harder to read.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+
+        # block = radio._cache_block(i, in_prog_mode=True)
+        result = _read_block(radio, i)
+        # block1, block2 = result[:16], result[16:]
+        data.extend(result)
+
+        LOG.debug("Downloaded memory channel %i" % i)
+        # LOG.debug(block)
+
+    _exit_programming_mode(radio)
```
kk7ds, line 306: This should be in a `finally` block to make sure it runs even if there is an error. Easiest way to do that is in your `sync_in()` and `sync_out()` methods.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1152 @@
...
+    status.msg = "Downloading from radio"
+
+    status.cur = 0
+    status.max = radio._upper
+    radio.status_fn(status)
+
+    data = bytearray()
+
+    _enter_programming_mode(radio)
+    for i in range(1, radio.MEM_ROWS, 2):
```
kk7ds, line 294: Yeah, as above, please do not try to download selections of the memory like this. Download the entire thing always.

kk7ds (changes requested):
Hi, thanks for working on this model. I didn't do a thorough review of this because I think there are a bunch of structural things that need to change in this driver to get it close. Probably in order:

1. Download the entire memory of the radio as an image, upload specific sections if necessary (not common)
2. Use bitwise for the native formats, don't do all the hand-encoding of things and forcing it raw into the data structures
3. The communication timing stuff and the multiple wraps of `read()` and `write()` are anti-patterns that should be avoided.

paulober:
@kk7ds Thanks for the feedback i'll look into it. For now, how can i build a windows installer so people can test it?

kk7ds:
Just attach the `radtel_rt493.py` file to an issue and they can load it directly using this procedure: https://www.chirpmyradio.com/projects/chirp/wiki/LoadingTestModules ... no need to build a full binary install.

chirp/drivers/radtel_rt493.py
paulober: What do you mean by this?
There is no command to download everything at once (at least not one that I know of) therefore I have to send a read command per 2 memory channels. And I put it in a radio.MEM_ROWS so this could be overwriten by alias driver classes.

kk7ds: There's no command to "read a memory" or "read two memories" either. You're just reading memory in 0x20-byte chunks, you just happen to be starting at the address of the first memory. You should start at zero and read the full size of memory. This radio appears to behave just like 80% of all the other chinese radios that copied kenwood's model, so you can look at those for inspiration.

Here's an example that is probably almost exactly what you need to do: https://github.com/kk7ds/chirp/blob/master/chirp/drivers/radtel_rt490.py#L364 and TBH, you might even be able to just subclass that radio. In fact the memory structure of that radio looks quite similar to what you have here, so it might be best to re-use a lot more of that.

paulober:
You mentioned that maybe uploading only stuff that changed would be a good idea (if i remember correclty). 
I just found out that if you upload for example only the channels and not the aditional memory rows coming afterwards with some settings you will lose their content. In the current version i fogot to update the limit in the uplaod function causing los of every memory line after the latest uploaded one.

kk7ds:
Not that it's a good idea, that it's the way to handle the case where you should only upload part of the memory. Some radios store calibration settings in the main area of memory which should NOT be uploaded from one radio to another. If that's the case here, you want to download everything, but only upload the regions that should be actually sent back to the radio. This is how OEM software works and is the case for many of the radio in-tree. Better-designed radios store calibration settings in a different section of memory and expect the whole image to be uploaded every time.

But yes, what you do not want to do is try to pick-and-choose what you download and upload from the radio unless you definitely know what you're doing because it's easy to get some settings not updated that are dependent on other changes you make (like memory content and scan flags stored separately). It is by far much better to download and upload the whole image of the radio unless you have a specific need not to, which is why I want you to (at least) download the whole thing.

chirp/drivers/radtel_rt493.py
paulober: I'm not sure what you mean. I convert the addr to int with byteorder big because it is of type bytes.

chirp/drivers/radtel_rt493.py
paulober: I just tried with 0x40 instead of 0x20 at the end of the address like they did in the RT-490 driver but this just crashes the communication and you get ff and 1c infinelty.

chirp/drivers/radtel_rt493.py
paulober: I tried it with `0x40` at the end (like rt490 by default) but that just brakes the communication.

kk7ds: Yeah, that's fine, use 0x20 if that's the block size the radio wants. You just want to read out all of memory in whatever the radio's native block size is.

paulober:
Is there an option how I can provide a default image for the radio so users have an option like "Reset To > EU Default".
This image would contain the default channel setup when you receive the radio brand new.

PR #1279, opened 2025-03-04
IC-T10: Adding Icom IC-T10 driver. Fixes #10181

karvanitis:
Adding support for Icom IC-T10 driver, including all memories and settings.

Fixes #10181


chirp/drivers/ict10.py
```diff
@@ -0,0 +1,619 @@
...
+        mem.mode = MODES[_mem.mode]
+        mem.duplex = DUPLEXES[_mem.duplex]
+        mem.dtcs_polarity = DTCS_POLARITY[_mem.dtcs_polarity]
+        mem.tmode = TMODES[_mem.tmode]
+        mem.power = POWER_LEVELS[_mem.power]
+
+        mem.extra = self._get_extra(_mem)
+
+        if _skp is not None:
+            mem.skip = (_psk & bit and "P") or (_skp & bit and "S") or ""
```
kk7ds, line 506: Can you make this a bit array instead of masking out the bits yourself please?

chirp/drivers/ict10.py
```diff
@@ -0,0 +1,619 @@
...
+
+        return mem
+
+    def get_memory(self, number):
+        if not self._mmap:
+            self.sync_in()
```
kk7ds, line 512: This is not okay to do here. I know it's in the other two drivers, but it shouldn't be. CHIRP opens and closes the port in the UI, this is not likely to ever be open here. I have had a todo personally to clean these up and try to unify these drivers, but I don't have radios to test the result with. Can you please help posterity here by cleaning up the others while you do this, and ideally making them all share a common core? Aside from maybe settings, they all appear mostly the same, modulo the channel count and memory format.

chirp/drivers/ict10.py
```diff
@@ -0,0 +1,619 @@
...
+
+        assert (self._mmap)
```
kk7ds, line 514: These asserts are compiled out at runtime. Also, there's no need to assert that `_mmap` exists. If we're calling `get_memory()` we expect the driver has been initialized.

chirp/drivers/ict10.py
```diff
@@ -0,0 +1,619 @@
...
+            except KeyError:
+                raise errors.InvalidMemoryLocation(
+                    "Unknown channel %s" % number)
+
+        return self._get_memory(number, extd_number)
+
+    def _fill_memory(self, number):
+        _mem = self._memobj.memory[number]
+
+        assert (_mem)
```
kk7ds, line 530: Same here - this will never not be true, and it also isn't run in the built versions.

chirp/drivers/ict10.py
```diff
@@ -0,0 +1,619 @@
...
+
+        # zero-fill
```
kk7ds, line 532: This isn't actually a zero fill (because rtone/ctone). However, it would be a lot better to do `_mem.fill_raw(b'\x00')` and then set those two values accordingly.

chirp/drivers/ict10.py
```diff
@@ -0,0 +1,619 @@
...
+            elif mem.skip == "P":
+                _skp &= ~bit
+                _psk |= bit
+            else:
+                _skp &= ~bit
+                _psk &= ~bit
+
+    def set_memory(self, mem):
+        if not self._mmap:
+            self.sync_in()
```
kk7ds, line 611: Same.

chirp/drivers/ict10.py
```diff
@@ -0,0 +1,619 @@
...
+from chirp import chirp_common, bitwise, errors, directory
+from chirp.settings import RadioSetting, RadioSettingGroup, \
+    RadioSettingValueInteger, RadioSettingValueList, \
+    RadioSettingValueBoolean, RadioSettingValueString, \
+    RadioSettings
+
+LOG = logging.getLogger(__name__)
+
+
+ICV86_MEM_FORMAT = """
```
kk7ds, line 28: You're failing the PR checks because this driver is nearly a copy of the `icv86` driver, and here's the proof :)

Please subclass that driver and just override what you need here (or better yet, refactor the base to something common and have both inherit from it. In the py2->py3 conversion, I (and others) spent *hours* converting the same copied code in dozens of drivers, instead of fixing things in one place. The new PR check is there to enforce that people stop just copying a driver to a new file and changing a few things, which is very hard for a human reviewer to notice. The comments below would apply to three drivers if the v80, v86, and this new driver all shared a common base.

chirp/drivers/ict10.py
karvanitis: Removed sync_in, will address remainder of comment below.

chirp/drivers/ict10.py
karvanitis: There are a few ICOM radios that all share similar memory format, features, etc..  as a result depending on the author of a particular driver and their individual choices there will be similarities between implementations.  For this reasons, there also exist drivers that while sharing a similar memory footprint pass the PR checks.

A common core is a fine request, however I would prefer to do this in a subsequent commit, as I no longer have access to this particular radio (it belong to a friend and need some time to retrieve it).

In theory most ICOM ICF based HTs from a certain generation (ICTxx, ICVxx, etc..) could be subclassed however this may be mitigated by access to these physical radios.

Please share your preference on how to proceed, and assuming you are ok with performing the work in a subsequent commit I can create the ticket.

kk7ds: I'd really rather not put this in and then have the refactor die on the vine because the pressure to combine is off. Are we talking like a week or a year?

Surely you can at least subclass the V86 for this one and override what you need without changing the base driver right? Also, for really minor stuff I would think the driver tests should be good enough (i.e. to ensure that get and set still work).

karvanitis: Either way I plan to have access to the radio this week, one a short term basis as it is in use. I own the other radios. Incremental changes will help mitigate any potential breaks, imo, as we will be starting with a known working base.  Thats my preference at least.

kk7ds: Okay, so I have your word you'll be back in a weekish to refactor right? :)

This needs a rebase and I'll override the PR check for duplication.

karvanitis: Im retired, so ill do my best ;)

Do you need me to rebase?

kk7ds: I thought you would, but I can do it for you

karvanitis: Much appreciated.

Issue [11902](https://chirpmyradio.com/issues/11902) created to address concerns.

PR #1281, opened 2025-03-07
Paste fixes

kk7ds:
- **Revert "th_uv88: Add Dual Wait settings for RA89,P2,etc"**
- **Add "copy portable" to memedit right-click menu**
- **Fix TSV paste bringing in a ghost memory**


kk7ds:
@KC9HI This adds "copy portable" and also fixes the ghost memory, I was just being stupid and was able to repro it.

KC9HI:
@kk7ds I am glad that you were able to reproduce it and fix it. Thanks for both this and the right-click menu.

PR #1284, opened 2025-03-13
Fixes bug #11882 in tdh8.py

TrimbleSoftware:
# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -703,6 +703,7 @@ HOP_LIST = ["A", "B", "C", "D"]
 SYNC_LIST = ["ON", "OFF"]
 LANG_LIST = ["Chinese", "English"]
 BTV_SAVER_LIST = ["OFF", "1:1", "1:2", "1:3", "1:4"]
+H3_BTV_SAVER_LIST = ["OFF", "1:1", "1:2", "1:3", "1:4", "1:8"]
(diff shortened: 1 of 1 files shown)


chirp/drivers/tdh8.py
```diff
@@ -1477,12 +1478,17 @@ def _get_settings(self):
...
             basic.append(rs)
 
-            rs = RadioSetting("save", "Battery Save",
+            rs = RadioSetting("save",
+                              "Power Save" if self.MODEL in H3_LIST
+                              else "Battery Save",
                               RadioSettingValueList(
-                                  BTV_SAVER_LIST,
+                                  H3_BTV_SAVER_LIST if self.MODEL in H3_LIST
+                                  else BTV_SAVER_LIST,
```
kk7ds, line 1489: This is pretty dense, so it'd be better if you just set the list on the class (once for H8, override for H3) and reference `self.SAVER_LIST` here or something. Anyway, not a huge deal so I'll let it slide, but if you want to follow up with a cleanup that would be appreciated :)

kk7ds:
In the future, please put "Fixes #11882" exactly in the commit message. I'm not sure if it will pick up yours with "bug" in the middle or not (we'll see).

Thanks!

TrimbleSoftware:
Per your suggestion I refactored the code and submitted it as PR #1285. Thx, Fred

PR #1285, opened 2025-03-14
Related to #11882 Code cleanup in tdh8.py

TrimbleSoftware:
Cleaned up code per your suggestions noted on PR #1284. Overall this was a good idea and will make the driver code more readable and maintainable in the future. Thanks for the suggestions.


kk7ds:
This needed to be rebased and also had a style error in it (trailing whitespace). I fixed both, thanks for the follow up!

PR #1287, opened 2025-03-14
Related to #11882 Added Roger Beep list for TD-H3

TrimbleSoftware:
Found that the driver module tdh8.py was not allowing the selection of different Roger Beep settings for the TD-H3. Added a Roger Beep list for the TD-H3 for Off, TONE1 and TONE2 to allow different roger beep tones. Added logic to allow the TD-H8 to still be a boolean setting with a checkbox.

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -2641,7 +2641,7 @@ class TDH8_GMRS(TDH8):
         if 31 <= mem.number <= 54:
             if mem.freq not in GMRS_FREQS:
                 msgs.append(chirp_common.ValidationError(
-                    "The frequency in channels 31-54 must be between"
+                    "The frequency in channels 31-54 must be between "
                     "462.55000-462.72500 in 0.025 increments."))
             if mem.duplex not in ('', '+', 'off') or (
                     mem.duplex == '+' and mem.offset != 5000000):
--- a/chirp/bandplan.py
+++ b/chirp/bandplan.py
@@ -69,7 +69,7 @@ class Band(object):
             return self
         limits = (self.limits[0] + self.offset, self.limits[1] + self.offset)
         offset = -1 * self.offset
-        if self.duplex in '+-':
+        if self.duplex and self.duplex in '+-':
             duplex = '+' if self.duplex == '-' else '-'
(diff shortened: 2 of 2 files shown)


TrimbleSoftware:
Sorry, I did the git pull in the wrong order. Can you rebase this on your end to remove the merge commit? Thx, Fred

chirp/drivers/tdh8.py
```diff
@@ -1519,13 +1519,23 @@ def _get_settings(self):
         basic.append(rs)
 
         if self.MODEL != "RT-730":
```
kk7ds, line 1521: This line becomes unnecessary now right? For cleanliness this can just be the `elif` condition instead of the new `else` right?

chirp/drivers/tdh8.py
```diff
@@ -1519,13 +1519,23 @@ def _get_settings(self):
...
+                rs = RadioSetting("rogerprompt", "Roger",
+                                  RadioSettingValueBoolean(
+                                      _settings.rogerprompt))
+
             basic.append(rs)
 
         rs = RadioSetting("txled", "Disp Lcd(TX)",
-
                           RadioSettingValueBoolean(_settings.txled))
+
```
kk7ds, line 1538: Please try to avoid adding or removing random whitespace like this. It makes the lines look changed in the git history when they weren't, which makes it hard to figure out what/why something was changed years later.

kk7ds (commented):
You can rebase it after the fact and just `git push -f` to update this. But yes, I can do it this time if you need.

Also, it's really best if you don't do these in your `master` branch, but create a new topic branch for a change, which also makes rebasing easier. I can push to your master branch to fix stuff for you, but I have to do extra steps which I don't have to do if you use a different branch.

Otherwise, thanks for these changes!

chirp/drivers/tdh8.py
TrimbleSoftware: if the radio is not a RT-730 it could either still be an H8 or an H3 (or eventually an H3 Plus). The H8 and the H3 have different roger beep options. This logic keeps then separate I believe...

Have you heard much of anything on adding the H3 Plus to CHIRP. Should be pretty easy if TIDRADIO would share a memory dump map.

PR #1289, opened 2025-03-15
Related to #11882 Add roger beep list for TD-H3

TrimbleSoftware:
Replaces faulty PR #1287

Found that the driver module tdh8.py was not allowing the selection of different Roger Beep settings for the TD-H3. Added a Roger Beep list for the TD-H3 for Off, TONE1 and TONE2 to allow different roger beep tones. Added logic to allow the TD-H8 to still be a boolean setting with a checkbox.

Hope I can finally get a clean PR... LOL!

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -1519,12 +1519,20 @@ class TDH8(chirp_common.CloneModeRadio):
         basic.append(rs)
 
         if self.MODEL != "RT-730":
-            rs = RadioSetting("rogerprompt", "Roger",
-                              RadioSettingValueBoolean(_settings.rogerprompt))
+            if self.MODEL in H3_LIST:
+                # H3 uses roger-beep list
+                rs = RadioSetting("rogerprompt", "Roger",
+                                  RadioSettingValueList(
+                                      self._roger_list,
+                                      current_index=_settings.rogerprompt))
+            else:
+                # H8 uses roger-beep bool
+                rs = RadioSetting("rogerprompt", "Roger",
+                                  RadioSettingValueBoolean(
+                                      _settings.rogerprompt))
             basic.append(rs)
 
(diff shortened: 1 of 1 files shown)


chirp/drivers/tdh8.py
```diff
@@ -1519,12 +1519,20 @@ def _get_settings(self):
         basic.append(rs)
 
         if self.MODEL != "RT-730":
```
kk7ds, line 1521: This will allow through H3 and H8

chirp/drivers/tdh8.py
```diff
@@ -1519,12 +1519,20 @@ def _get_settings(self):
...
-            rs = RadioSetting("rogerprompt", "Roger",
-                              RadioSettingValueBoolean(_settings.rogerprompt))
+            if self.MODEL in H3_LIST:
```
kk7ds, line 1522: This will allow H3, not H8

chirp/drivers/tdh8.py
```diff
@@ -1519,12 +1519,20 @@ def _get_settings(self):
...
+                # H3 uses roger-beep list
+                rs = RadioSetting("rogerprompt", "Roger",
+                                  RadioSettingValueList(
+                                      self._roger_list,
+                                      current_index=_settings.rogerprompt))
+            else:
```
kk7ds, line 1528: This will allow H8, not H3 because they took the path above. RT-370 will not make it here because of the first if.

If you remove the top if and change this `else` to `elif self.MODEL != 'RT-370'` then I think it's the same without the extra nesting. Further, you could use something like `elif self.MODEL.startswith('TD-H8')` if you want it a little more future proof or something.

Anyway, not a big deal, I just don't really like nested `if`s when they look like this

kk7ds (commented):
I closed the bug manually since it didn't auto-close in the previous one (as I suspected due to the extra word in between). This will still relate to that though for posterity.

PR #1293, opened 2025-03-18
add UV21ProV2 and UV21ProGPS

sweelim20:
# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/baofeng_uv17Pro.py
+++ b/chirp/drivers/baofeng_uv17Pro.py
@@ -1441,6 +1441,27 @@ class UV17Pro(bfc.BaofengCommonHT):
 
 
 @directory.register
+class UV21ProV2(UV17Pro):
+    VENDOR = "Baofeng"
+    MODEL = "UV-21ProV2"
+
+    _airband_rx = (108000000, 135999999)
+    _vhf_range = (136000000, 174000000)
+    _vhf2_range = (220000000, 260000000)
+    _uhf_rx1_range = (350000000, 390000000)
+    _uhf_range = (400000000, 479999999)
+    _uhf_rx2_range = (480000000, 520000000)
+
+    VALID_BANDS = [_airband_rx,
+                   _vhf_range,
+                   _vhf2_range,
+                   _uhf_rx1_range,
+                   _uhf_range,
+                   _uhf_rx2_range]
+    MODES = UV17Pro.MODES + ['AM']
+
+
+@directory.register
 class UV25(UV17Pro):
     VENDOR = "Baofeng"
     MODEL = "UV-25"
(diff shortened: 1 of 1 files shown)


chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -168,7 +168,7 @@ def _upload(radio):
     status = chirp_common.Status()
     status.cur = 0
     status.max = radio.MEM_TOTAL // radio.BLOCK_SIZE
-    status.msg = "Cloning from radio..."
+    status.msg = "Cloning to radio..."
```
kk7ds, line 171: This should be a separate commit since it's unrelated to the UV21 (but thanks for catching it)

kk7ds:
Per guideline 3 above can you please squash your fixes commit into the parent? Also split out the string fix please.

Thanks!

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -1386,6 +1386,27 @@ def set_memory(self, mem):
         self.set_memory_common(mem, _mem)
 
 
+@directory.register
+class UV21ProV2(UV17Pro):
+    VENDOR = "Baofeng"
+    MODEL = "UV-21ProV2"
```
kk7ds, line 1392: Also please see #1167. It seems like this doesn't need its own entry and we can just subclass the other or put it in `share/model_alias_map.yaml` right?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -1425,6 +1446,15 @@ def get_features(self):
         return rf
 
 
+@directory.register
+class UV21ProGPS(UV17ProGPS):
+    VENDOR = "Baofeng"
+    MODEL = "UV-21ProGPS"
+
+    VALID_BANDS = UV21ProV2.VALID_BANDS
+    MODES = UV21ProV2.MODES
```
kk7ds, line 1455: I haven't looked at what the base class uses but there are a lot of variants in here already that differ by very little if anything. Is this really a unique combination?

chirp/drivers/baofeng_uv17Pro.py
sweelim20: #1174 improves by using the VALID_BANDS from UV17Pro and added AM.

I was originally motivated because Airband AM support was missing from UV17Pro.
When I looked into it, I realized that the frequency bands for UV21ProV2 is not same as UV17Pro. Hence the new supported frequency bands. 

My suggestion is add the frequency bands above after #1174 merges.

chirp/drivers/baofeng_uv17Pro.py
sweelim20: see [#1296](https://github.com/kk7ds/chirp/pull/1296) for msg fix PR.

PR #1298, opened 2025-03-21
update upper frequency limit that is displayed in validation error msg

mightyohm:
Modify the upper frequency limit that is displayed in the validation error message to be consistent with the range of allowable GMRS frequencies for memory slots 31-54.

# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -2650,7 +2650,7 @@ class TDH8_GMRS(TDH8):
             if mem.freq not in GMRS_FREQS:
                 msgs.append(chirp_common.ValidationError(
                     "The frequency in channels 31-54 must be between "
-                    "462.55000-462.72500 in 0.025 increments."))
+                    "462.55000-467.71250 in 0.025 increments."))
             if mem.duplex not in ('', '+', 'off') or (
                     mem.duplex == '+' and mem.offset != 5000000):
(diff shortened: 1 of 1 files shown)


kk7ds:
Thanks!

In the future it's best to keep the first line of the commit message short (<80 chars) and mention the driver or model so people reading from the ML will know what has changed. Also good to reference a bug number (like `Fixes #123` or `Related to #123`) if applicable.

mightyohm:
Updated commit message. Thanks for the feedback!

kk7ds:
I already merged it, so too late. But also, the title of the PR here in github is distinct from the commit message (although it populates from it the first time if you don't change it). It's the commit message that matters as that is all that gets persisted into the tree. Anyway too late now and not a big deal. Thanks!

mightyohm:
Thanks for the info! I understand now. Changes to the branch after the merge don't matter. Still figuring out how all of this works. Thanks again!

mightyohm:
What determines when a new chirp-next is built? I thought it was a nightly build?

mightyohm:
Dan,

Excellent, thank you very much for the info. I was able to download the new build this morning and see the change. This is my first contribution to a project on Github, so even though the change was trivial it's neat to see it in the app!

Jeff

kk7ds:
Cool, I and the other users thank you :)

PR #1299, opened 2025-03-22
Add rt493 rt580g drivers

paulober:
rt-493 and rt-580g: Add drivers

Adds drivers for the Radtel RT-493 and RT-580G radios.


paulober:
I think there is a bug with the immutable system. If you add "extra" to the list of immutable values your `mem.extra` fields get disabled as they should. But if you then try to edit the name of this channel you get a warning that extra is set to immutable. I think the name field should not get counted towards the extra fields but if, then it should also be disabled in UI.

kk7ds:
> I think there is a bug with the immutable system. If you add "extra" to the list of immutable values your `mem.extra` fields get disabled as they should. But if you then try to edit the name of this channel you get a warning that extra is set to immutable. I think the name field should not get counted towards the extra fields but if, then it should also be disabled in UI.

Yep, probably, I'll have a look. Can you open a bug on the chirp site for this and I'll take it?

paulober:
> Yep, probably, I'll have a look. Can you open a bug on the chirp site for this and I'll take it?

https://chirpmyradio.com/issues/11904

paulober:
@kk7ds I want to finally get this driver merged so I was working on the last tox errors. One of them again concerns immutability: `chirp.chirp_common.ImmutableValueError: Field tmode is not mutable on this memory` I also get it for offset and probably for more if I were to remove tmode. So basically the policy is if a channel is in AM mode most settings for that channel are locked.
What is the origin of this error and how should I fix it?

kk7ds:
`ImmutableValueError` is raised when you try to modify something on a memory that is in the `immutable` list set on it.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+    data = radio.pipe.read(length)  # Read the data
+    if not data:
+        raise errors.RadioNoContactLikelyK1()
+
+    # DEBUG
+    LOG.info("Response:")
+    LOG.debug(util.hexprint(hdr + data))
+
+    c, a, resp_length = struct.unpack(">BHB", hdr)
+    if a != int.from_bytes(addr, byteorder="big") \
```
kk7ds, line 59: Please don't do this. You should be passing `addr` into this as an integer not a bytestring, so no such interpretation should be necessary. I try to discourage use of `int.from_bytes()` everywhere in the codebase and use `struct` where necessary to keep things consistent and explicit about lengths.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+    hdr = radio.pipe.read(4)  # Read 4 byte header
+    if not hdr:
+        raise errors.RadioNoContactLikelyK1()
+    data = radio.pipe.read(length)  # Read the data
+    if not data:
+        raise errors.RadioNoContactLikelyK1()
+
+    # DEBUG
+    LOG.info("Response:")
+    LOG.debug(util.hexprint(hdr + data))
```
kk7ds, line 56: No need to do this. We have automatic full tracing happening and this just inflates the logs (and makes them hard to read).

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+
+def _get_memory_address(channel_id):
+    block = (channel_id - 1) // 16           # Determines high byte
+    offset = ((channel_id - 1) % 16) * 0x10  # Determines low byte
+
+    return (block, offset)
+
+
+def _read_block(radio, channel_id):
+    address = bytes(_get_memory_address(channel_id))
```
kk7ds, line 120: This should be returning an integer and not passing around order-dependent bytestrings for something like this. Fixing this will also address what would be my comment on L131 below about the comparison. And of course, the manual formatting of each byte on L123...

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+    # 52=read | 20 = hex(32 bytes) = length of 2 channels
+    # + settings without header
+    cmd = struct.pack(">BBBB", 0x52, address[0], address[1], 0x20)
+
+    radio.pipe.write(cmd)
+    data = _recv(radio, address, 32)
+
+    # Sending an accept after reading the first line will crash
+    # the communication. The accept is sent when entering programming
+    # mdoe, so before reading the first block
```
kk7ds, line 130: "mode" is misspelled/.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+
+DTCS_BYTE_LOOKUP_HIGH = {
+    0x09: 64, 0x0A: 65, 0x0B: 66, 0x13: 67, 0x19: 68, 0x1A: 69,
+    0x25: 70, 0x26: 71, 0x2A: 72, 0x2C: 73, 0x2D: 74, 0x32: 75,
+    0x34: 76, 0x35: 77, 0x36: 78, 0x43: 79, 0x46: 80, 0x4E: 81,
+    0x53: 82, 0x56: 83, 0x5A: 84, 0x66: 85, 0x75: 86, 0x86: 87,
+    0x8A: 88, 0x94: 89, 0x97: 90, 0x99: 91, 0x9A: 92, 0xA5: 93,
+    0xAC: 94, 0xB2: 95, 0xB4: 96, 0xC3: 97, 0xCA: 98, 0xD3: 99,
+    0xD9: 100, 0xDA: 101, 0xDC: 102, 0xE3: 103, 0xEC: 104,
+}
```
kk7ds, line 232: Isn't this just a list of the codes in order? Why can't you use `chirp_common.DTCS_CODES`?

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+                raise ValueError("Invalid tone value: %i" % value)
+
+        # DTCS Normal Polarity (DxxxN) and DTCS Inverted Polarity (DxxxI)
+        dtcs_code_idx = _decode_dtcs_byte(byte1, byte2)
+        assert dtcs_code_idx != -1
+        dtcs_polarity = "N" if byte2 == 0x28 or byte2 == 0x29 else "R"
+        if 0 <= dtcs_code_idx <= len(self.DTCS_CODES) - 1:
+            return "DTCS", self.DTCS_CODES[dtcs_code_idx], dtcs_polarity
+        else:
+            return None, None, None
```
kk7ds, line 399: This looks like you're decoding each byte independently and doing your own math. I think you could use the `union` functionality in bitwise to avoid a lot of this.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+            False -> If the channel is excluded from the scan list.
+        """
+        if not 1 <= channel <= 199:
+            raise ValueError("Channel number must be between 1 and 199")
+
+        raw_bytes = self._memobj.scan_add_settings.get_raw()
+
+        # Each byte represents 8 channels, find the correct byte and bit
+        byte_index = (channel - 1) // 8
+        bit_index = (channel - 1) % 8
```
kk7ds, line 419: Looks like you should be using the `bit[N]` type in bitwise, which lets you index individual bits for flag arrays like this, and many of the things below.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+        mem.number = number
+
+        self.decode_mem_blob(_mem, mem)
+
+        return mem
+
+    def _encode_qt_dqt(self, tone_mode, tone_value=None, tone_polarity=None):
+        if tone_mode == "Tone" or tone_mode == "TSQL":
+            assert tone_value is not None
+            return int(tone_value * 10).to_bytes(2, "little")  # Encode as Hz
```
kk7ds, line 634: Please do not use `int.to_bytes()` like this. The bitwise field should handle the endianess and conversion. And please don't do your own formatting of low-level memory data except in places where it's entirely avoidable.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+
+            return bool(int.from_bytes(byte2, byteorder="big") & 0x01)
+        elif setting == "CPYC":
+            byte1 = self._memobj.memory[6-1].global3.get_raw()
+
+            return bool(int.from_bytes(byte1, byteorder="big") & 0x01)
+        elif setting == "VOXF":
+            byte1 = self._memobj.memory[4-1].global1.get_raw()
+
+            return bool(int.from_bytes(byte1, byteorder="big") & 0x01)
```
kk7ds, line 785: Please don't do this sort of thing. Identify the bits in the bitwise definition and get/set them ad objects instead of doing your own bit ops.

chirp/drivers/radtel_rt493.py
```diff
@@ -0,0 +1,1021 @@
...
+                self._set_bool(element.get_name(), element.value)
+            elif isinstance(element.value, RadioSettingValueList):
+                # options = self._get_setting_options(element.get_name())
+                options = self._SETTINGS_OPTIONS[element.get_name()]
+                self._set_int(element.get_name(),
+                              options.index(str(element.value)))
+            elif isinstance(element.value, RadioSettingValueInteger):
+                self._set_int(element.get_name(), int(element.value))
+            else:
+                LOG.error("Unknown setting type: %s" % element.value)
```
kk7ds, line 994: New drivers should ideally be using `MemSetting` which handles symbolic application of almost all the standard types directly to the memory object, eliminating the need for all this bespoke code.

chirp/drivers/radtel_common.py
```diff
@@ -0,0 +1,264 @@
...
+
+    MEM_ROWS: int
+
+    def get_mmap(self) -> bytearray: ...
+
+
+def rt_clean_buffer(radio: RadtelLikeRadio, LOG: Logger):
+    radio.pipe.timeout = 0.005
+    junk = radio.pipe.read(256)
+    radio.pipe.timeout = 5  # 5000ms
```
kk7ds, line 27: Please don't do this sort of thing. On Windows, setting the timeout value flushes the buffer and can do other things to the serial driver that don't happen on the other platforms. Switching the timeout in quick succession will give platform-dependent results and should be avoided. I imagine `Serial.reset_input_buffer()` is all you need.

chirp/drivers/radtel_common.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    if len(data) != 32 + 1:
+        raise ValueError("Data must be exactly 32 bytes long.")
+
+    # Replace last byte with 0 for calculation
+    calculated_checksum = _calculate_checksum(data[:-1])
+    return calculated_checksum == data[-1]
+
+
+def rt_read_block(
```
kk7ds, line 68: This and `rt_do_download()` don't appear to be used by at least the rt493 driver. TBH, this common file doesn't seem to be all that common and shared bits could easily be in one of the two drivers and imported from the other. Several of my comments in the r493 file that I started reviewing first apply here.

chirp/drivers/radtel_common.py
```diff
@@ -0,0 +1,264 @@
...
+from typing import Literal, Tuple, Protocol, Any, Callable
+
+from chirp import chirp_common, errors, util
+
+COMMAND_ACCEPT = b"\x06"
+CHARSET_HEX = "0123456789ABCDEFabcdef"
+BLOCK_SIZE = 0x10  # 16 bytes
+
+
+class RadtelLikeRadio(Protocol):
```
kk7ds, line 12: I don't see that you're using this anywhere? Also, please don't create new inheritance hierarchies for your drivers. If you want to have a common base class that inherits from chirp_common.Radio that you use for your drivers, that's cool, but this has no common ancestry.

kk7ds (commented):
Thanks for working on this. I left a bunch of comments. I haven't reviewed all of it because a bunch of things will change as a result of those. Also, are you sure that none of the existing drivers are re-usable or adaptable for either of these radios?

PR #1300, opened 2025-03-22
Disable immutable mem.extra in expanded memory editor

kk7ds:
This was already working for the memory properties editor, but now
works for the expanded view as well.

Fixes #11904


kk7ds:
cc @paulober

PR #1301, opened 2025-03-25
kg816/818: Fix query string processing

mildebrandt:
Sometime between legacy and now, the strings in wouxun.py were changed to bytes. Most radios in this file define a tuple of bytes to be sent to the radio....if one doesn't work, the next one is tried. For the 816/818, only one `bytes` object is defined. The code that processes this checked if the input was a `str`. It failed on a `bytes` object and started processing the input as an iterable.

Related to #11208

This is why the person in issue 11208 got an error reading their radio, even though chirp doesn't directly support it. Several users in several forums suggest using the 816 driver to program the 805 at their own 
risk.

### DIFF
--- a/chirp/drivers/wouxun.py
+++ b/chirp/drivers/wouxun.py
@@ -60,7 +60,7 @@ class KGUVD1PRadio(chirp_common.CloneModeRadio,
     MODEL = "KG-UVD1P"
     _model = b"KG669V"
 
-    _querymodel = (b"HiWOUXUN\x02", b"PROGUV6X\x02")
+    _querymodels = (b"HiWOUXUN\x02", b"PROGUV6X\x02")
 
     CHARSET = list("0123456789") + \
         [chr(x + ord("A")) for x in range(0, 26)] + list("?+-")
@@ -227,22 +227,11 @@ class KGUVD1PRadio(chirp_common.CloneModeRadio,
             "6. Click OK to upload image to device.\n")
         return rp
 
-    @classmethod
-    def _get_querymodel(cls):
-        if isinstance(cls._querymodel, str):
-            while True:
-                yield cls._querymodel
-        else:
-            i = 0
-            while True:
-                yield cls._querymodel[i % len(cls._querymodel)]
-                i += 1
-
     def _identify(self):
         """Do the original Wouxun identification dance"""
-        query = self._get_querymodel()
         for _i in range(0, 10):
-            self.pipe.write(next(query))
+            model = self._querymodels[_i % len(self._querymodels)]
+            self.pipe.write(model)
             resp = self.pipe.read(9)
             if len(resp) != 9:
                 LOG.debug("Got:\n%s" % util.hexprint(resp))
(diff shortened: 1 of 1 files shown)


chirp/drivers/wouxun.py
```diff
@@ -229,7 +229,7 @@ def get_prompts(cls):
 
     @classmethod
     def _get_querymodel(cls):
-        if isinstance(cls._querymodel, str):
+        if isinstance(cls._querymodel, (str, bytes)):
```
kk7ds, line 232: I don't think this is the right solution, as `str` is never valid for this anymore. I think the best thing to do is just get rid of this whole thing and make `_querymodel` just always iterable (and also probably remove this `_get_querymodel()` thing entirely).

Would you mind making that change and testing with whatever you have?

kk7ds: ...alternately I can make the change, if you're willing to test. I have something that uses this model somewhere, but  it would take me a non-trivial amount of time to find it (and the cable for it).

mildebrandt: I kinda thought you'd suggest that as I was looking at this. :) 

Sure, I can make that change...but can only test on the kg-808g physically.

kk7ds: Cool, thanks. I'll take responsibility for the other models :)

mildebrandt: Done and tested with my 805 radio.

kk7ds: Thanks. Should we add a model alias mapping for the 805 to the 816? I want to squash these commits anyway, so I can do that when I squash if you think it's appropriate. Also, sounds like this should `Fixes #11208` not just relate to it right?

mildebrandt: So, I'm not sure the 805 matches the 816...I haven't written to the radio yet using chirp. I'd like to create a driver that supports all the settings...which will take a bit. 

I could copy the memory layout, change it a bit, and maybe stuff that in this PR without the settings as a first pass. The number of channels is 128 for the 805 and 199 for the 816....that might be the only change, but I've only JUST started looking at it. Thoughts?

If you think this fixes #11208, that's fine. However, that person is trying to read/write an 805, and as it stands...this code will corrupt their radio if they write past channel 128 using the 816 driver.

mildebrandt: Personally, I'd rather keep the fix to 816/818 drivers separate from the creation of a new driver for the 805.

kk7ds: Okay, I just didn't know the details. Let's leave this as you have it, just fixing the `bytes` thing and cleaning up, related to that bug. Then keep actual support for the other radio as a separate thing. It's fine, btw, to subclass and disable features for the 805, we can merge that so people can at least use it for memories (if that's solid) and you can iterate on settings after. I really prefer smaller incremental patches anyway.

But, I want to squash these down. I'll do that now so we can merge. Thanks!

PR #1302, opened 2025-03-26
kt-5000: Initial support of QYT KT-5000.

iwane-pl:
Memories can be programmed, settings map doesn't match. Commit reuses the memory map of QYT KT-WP12, which isn't 100% accurate.

Currently settings page is disabled.

Related to #10530


chirp/drivers/btech.py
```diff
@@ -1156,7 +1160,7 @@ def get_settings(self):
             save = RadioSetting("settings.save", "Battery Save", rs)
             basic.append(save)
 
-        model_list = ["KT-8R", "KT-WP12", "WP-9900"]
+        model_list = ["KT-8R", "KT-WP12", "WP-9900", "KT-5000"]
```
kk7ds, line 1163: I assume these are part of your work on settings, but are not actually necessary for this patch to work since settings are disabled, right?

If so, let's remove (the changes to) these model checking lines from this patch and just keep the main hunk below. Otherwise another developer may be really confused why this model is in the list but has no settings in the UI until it's finished. Is that reasonable?

iwane-pl: 👍 To be honest, I left those conditions there and have forgotten about them. Cleaned them up - they'll be needed when settings make it to the driver.

kk7ds:
Yep, understood they'll be needed later, just no need for them now.

Just rebased and will merge once the tests pass.

PR #1304, opened 2025-03-27
Wouxun KG-805g: Add new radio driver

mildebrandt:
Add driver for Wouxun KG-805g. Currently supports only channel memory. Settings will come in a much later commit. I subclassed KGUVD1PRadio instead of KG816Radio since KG816 is more likely to change in a way that affects the KG805 code.

Fixes #11214
Fixes #11208
Fixes #7781


chirp/drivers/wouxun.py
```diff
@@ -1555,3 +1555,58 @@ class KG818Radio(KG816Radio):
...
+    """
+
+    def get_features(self):
+        rf = KGUVD1PRadio.get_features(self)
+        rf.has_settings = False
+        return rf
+
+    def process_mmap(self):
+        self._memobj = bitwise.parse(self._MEM_FORMAT, self._mmap)
+        # This sets our frequency ranges, so run it once after we process the
```
kk7ds, line 1603: Long line or errant cr in here. I can fix and rebase.

chirp/drivers/wouxun.py
```diff
@@ -1555,3 +1555,58 @@ class KG818Radio(KG816Radio):
...
+        # mmap to make sure they're set for later
+        self.get_settings()
+
+    def get_settings(self):
+        pass
+
+    @classmethod
+    def match_model(cls, filedata, filename):
+        return chirp_common.CloneModeRadio.match_model(filedata, filename)
```
kk7ds, line 1612: This is not necessary anymore for new models that are covered by the embedded metadata. Just `return False` here.

chirp/drivers/wouxun.py
kk7ds: Hmm, maybe not github just seems to be wrapping it too aggressively, nm.

mildebrandt: I'm not sure I understand, the line ends at 77 characters. It was copied from above, so not my comment. I had to remove the lion share of this function from the parent class.

PR #1306, opened 2025-03-28
KG805G fix: Duplex field calculated wrong

mildebrandt:
The duplex field was set to 'split' based on a bit in channel memory in the parent class, however that's incorrect for the 805G. I had to override the get_memory and set_memory methods to display the correct values in the UI.

### DIFF
--- a/chirp/drivers/wouxun.py
+++ b/chirp/drivers/wouxun.py
@@ -736,17 +736,7 @@ class KGUVD1PRadio(chirp_common.CloneModeRadio,
     def _is_txinh(self, _mem):
         return _mem.tx_freq.get_raw() == b"\xFF\xFF\xFF\xFF"
 
-    def get_memory(self, number):
-        _mem = self._memobj.memory[number - 1]
-        _nam = self._memobj.names[number - 1]
-
-        mem = chirp_common.Memory()
-        mem.number = number
-
-        if _mem.get_raw() == (b"\xff" * 16):
-            mem.empty = True
-            return mem
-
+    def _set_duplex_offset_freq(self, _mem, mem):
         mem.freq = int(_mem.rx_freq) * 10
         if _mem.splitdup:
             mem.duplex = "split"
@@ -763,6 +753,19 @@ class KGUVD1PRadio(chirp_common.CloneModeRadio,
             mem.offset = int(_mem.tx_freq) * 10
         else:
             mem.offset = abs(int(_mem.tx_freq) - int(_mem.rx_freq)) * 10
+
+    def get_memory(self, number):
+        _mem = self._memobj.memory[number - 1]
+        _nam = self._memobj.names[number - 1]
+
+        mem = chirp_common.Memory()
+        mem.number = number
+
+        if _mem.get_raw() == (b"\xff" * 16):
+            mem.empty = True
+            return mem
+
+        self._set_duplex_offset_freq(_mem, mem)
 
         if not _mem.skip:
             mem.skip = "S"
@@ -831,6 +834,9 @@ class KGUVD1PRadio(chirp_common.CloneModeRadio,
         LOG.debug("Set TX %s (%i) RX %s (%i)" %
                   (tx_mode, _mem.tx_tone, rx_mode, _mem.rx_tone))
 
+    def _set_split_duplex(self, _mem, mem):
+        _mem.splitdup = mem.duplex == "split"
+
     def set_memory(self, mem):
         _mem = self._memobj.memory[mem.number - 1]
         _nam = self._memobj.names[mem.number - 1]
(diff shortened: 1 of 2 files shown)


chirp/drivers/wouxun.py
```diff
@@ -1607,6 +1607,89 @@ def process_mmap(self):
     def get_settings(self):
         pass
 
+    def get_memory(self, number):
```
kk7ds, line 1610: So, it might not make sense, but could we just override the split/duplex part? Something like:
```
def get_memory(self, number):
    mem = super().get_memory(number)
    _mem = self._memobj.memory[number - 1]
    if abs(int(_mem.rx_freq) * 10 - int(_mem.tx_freq) * 10) > 70000000:
        ...
    return mem
```
This retains the behavior of the parent class (like if any other things get changed there you'll still benefit, while letting you properly override the behavior you need. Alternately, could split the get-and-set-the-duplex bit into a helper and just override that here.

Not a huge deal, but this copies a lot of code from the parent just to override one bit of the logic, which is less than ideal.

chirp/drivers/wouxun.py
```diff
@@ -1607,6 +1607,89 @@ def process_mmap(self):
...
+        mem.freq = int(_mem.rx_freq) * 10
+
+        if self._is_txinh(_mem):
+            # TX freq not set
+            mem.duplex = "off"
+            mem.offset = 0
+        elif int(_mem.rx_freq) == int(_mem.tx_freq):
+            mem.duplex = ""
+            mem.offset = 0
+        elif abs(int(_mem.rx_freq) * 10 - int(_mem.tx_freq) * 10) > 70000000:
```
kk7ds, line 1630: Can you use this helper instead?

https://github.com/kk7ds/chirp/blob/master/chirp/chirp_common.py#L1913

This helps to make the logic that fakes duplex for radios with separate tx/rx frequencies the same, especially if we ever change our heuristic for what we consider "too much to be offset" logic.

chirp/drivers/wouxun.py
mildebrandt: I looked into this, and the problem is that I changed the memory fields (because they don't match the KG-UVD1P) meaning I get a `No attribute splitdup in struct memory` error when the original function tries to access it. 

I could create a helper like you suggest, perhaps keying off the existence of the `splitdup` attribute to decide which logic to perform. Would that be preferable?

Also, I changed the `skip` attribute to `scan_add` (matching kg935g) because when set to `0` it skips, and when set to `1` it's included in the scan. The value matches closer to the name that way. I can revert that change too if using as much of the existing code is more important. I have no strong opinion either way. What would you prefer?

kk7ds: Ah I see, you've not got that thing.

No need to key off anything just run the helper in the base class, override the helper in the subclass. I meant something like this:
```
class Parent:
    def _set_duplex(self, mem, _mem):
        ... do the parent thing

    def get_memory(...):
        ...
        self._set_duplex(mem, _mem)
...

class Child(Parent):
    def _set_duplex(self, mem, _mem):
        ...do the child thing
```
> Also, I changed the skip attribute to scan_add (matching kg935g) because when set to 0 it skips, and when set to 1 it's included in the scan. The value matches closer to the name that way. I can revert that change too if using as much of the existing code is more important. I have no strong opinion either way. What would you prefer?

Well, this is why I think re-using as much as possible is best, because someone coming along and reading this whole driver is probably not going to notice that difference, and so if they end up trying to make some other change they expect to affect all of these in the same way, they're not going to notice the subtle bit flip in the child class.

Anyway, I'm not trying to nit out on everything, I'm just the guy that has to maintain these things when CHIRP enters its third decade and so I like to see differences as explicitly as possible. We have waaay too much copying code for no reason in the tree, and the scars from doing the py2 -> py3 conversion run pretty deep.

chirp/drivers/wouxun.py
mildebrandt: In `split_to_offset`:

```python
def split_to_offset(mem, rxfreq, txfreq):                                                                                                             
    """Set the freq, offset, and duplex fields of a memory based on                                                                                   
    a separate rx/tx frequency.                                                                                                                       
    """                                                                                                                                               
    if abs(txfreq - rxfreq) > to_MHz(70):                                                                                                             
        mem.freq = rxfreq
        mem.offset = txfreq
        mem.duplex = 'split'                                                                                                                          
    else:                                                                                                                                             
        offset = txfreq - rxfreq                                                                                                                      
        if offset < 0:
            mem.duplex = '-'
        elif offset > 0:                                                                                                                              
            mem.duplex = '+'
        mem.offset = abs(offset)
```

Is it purposely not setting `freq` in the `else` block?

kk7ds: Probably not *intentionally* just that most drivers do that pretty much first thing, and process the duplex/offset stuff later. Can't really explain why it's done in the split case and not below :)

Feel free to add it if you want to omit it from the calling code. I doubt it'll break anything but the tests should prove it.

mildebrandt:
I think I've addressed the feedback. Let me know if there's anything else. Thanks!

kk7ds:
Thanks, just need to squash the fixup commit into the parent (no need to keep history of the wrong/temporary thing forever). I'll do it when I rebase.

PR #1310, opened 2025-04-01
TH9000: Add programmable buttons to settings.

cmnybo:
I added support for configuring the programmable buttons on the THT TH-9000.


chirp/drivers/th9000.py
```diff
@@ -807,6 +867,96 @@ def _filter(name):
                               int(_freqrange.rxrangehi) / 10))
         freqrange.append(rs)
 
+        rs = RadioSetting(
+            "btn_p1", "P1 Short Press",
+            RadioSettingValueList(
```
kk7ds, line 872: There's a newer/better way to do this, which is `RadioSettingValueMap` which tolerates  invalid data better than the `List` variant. This is okay as it is, but you might want to consider converting these to use the map method for better safety in the future.

kk7ds:
I squashed the fix patch into the first one. In the future please just do that and update the PR. No need to keep the bad+good commits in the history forever :)

Thanks!

chirp/drivers/th9000.py
kk7ds: You did use `current_index` though, which is also good for safety. Just FYI is all.

PR #1311, opened 2025-04-16
New settings for Wouxun KG-805g

mildebrandt:
Add support for the following settings:
 - power save
 - squelch
 - roger beep
 - transmit time out
 - transmit time out alarm
 - vox
 - beep
 - auto lock

### DIFF
--- a/chirp/drivers/wouxun.py
+++ b/chirp/drivers/wouxun.py
@@ -330,6 +330,64 @@ class KGUVD1PRadio(chirp_common.CloneModeRadio,
         rf.can_odd_split = True
         return rf
 
+    @property
+    def transmit_time_out_setting(self):
+        options = ["%s" % x for x in range(15, 615, 15)]
+        rs = RadioSetting(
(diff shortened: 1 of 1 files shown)


chirp/drivers/wouxun.py
```diff
@@ -330,6 +330,64 @@ def get_features(self):
...
+                              self._memobj.settings.power_save))
+        return rs
+
+    @property
+    def vox_setting(self):
+        options = ["OFF"] + ["%s" % x for x in range(1, 11)]
+        rs = RadioSetting(
+            "vox",
+            "Vox",
+            RadioSettingValueList(
```
mildebrandt, line 379: Note: This was a `RadioSettingValueInteger` that went from 0 to 10. I changed it because the radio itself displays "OFF" at 0 and the `RadioSettingValueList` gives a better UX (drop-down choice rather than +/- to iterate through options).

Similar for `tx_time_out_alert`.

PR #1312, opened 2025-04-19
Update bg_BG.po

StoyanDimitrov:


kk7ds:
In the future, if you could make the commit message something like "Update Bulgarian Translation" that would be better. The users see the first line on the mailing list announcements, and most do not know what `bg_BG.po` is :)

Thanks!

PR #1314, opened 2025-04-19
th350: Fix FM preset failure in settings on US version

kk7ds:
The US version does not have FM presets so the base class needs to
check for and exclude that code if not.

Fixes #11948


kk7ds:
Yep, those are pre-existing, unrelated, and not critical. Thanks for the confirmation.

PR #1317, opened 2025-05-03
New Model Retevis RA25

ascendr:
My humble attempt at a new driver for Retevis RA25. This Radio is supposedly the same as the Anytone 799UV and Radioddity DB20 but I only have the Retevis so it's locked to that model for now. 

Related to Issue #9241


ascendr:
Fixed up failing tests

kk7ds:
Thanks, I won't get to review this until next week, just FYI.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+
+def send_cmd(serial, cmd):
+    try:
+        resp = b''
+        serial.write(cmd)
+        serial.flush()
+        timeout = time() + SERIAL_TIMEOUT
+        while time() < timeout:
+            if serial.inWaiting():
+                resp += serial.read()
```
kk7ds, line 231: Please don't do this. The serial layer already lets you read the amount of data you want with a timeout. Code that tries to read non-constant responses is almost always very fragile on different platforms. Please set a longish timeout (like 1s) and read exactly the number of bytes you expect for each response.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+        raise errors.RadioError(f"Error sending to serial {e}")
+    return resp
+
+
+def start_programming(serial):
+    try:
+        resp = send_cmd(serial, b'PROGRAM')
+        if resp != b'\x06':
+            log.error("Enter Programming failed")
+            raise errors.RadioError("ACK resp not received")
```
kk7ds, line 242: This will just get eaten a line later and throw away the actual exception (which looks more useful than the one you're raising on L244). Also, I very much recommend logging what you got as it makes debugging this sort of thing much easier. Sometimes it's helpful to see that someone is getting nothing, or getting something the radio would never send (indicating they chose the wrong serial port, etc).

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+def parse_response(r):
+    if sum(r[:-1]) % 256 != r[-1]:
+        raise errors.RadioError("Data check sum failed")
+    return r[5:-1]
+
+
+def get_ident(serial):
+    try:
+        bs = send_cmd(serial, b'\x02')
+        model = bs[1:8].decode()
```
kk7ds, line 266: Unless the radio is sending UTF-8, then this is wrong. Use struct.unpack() to get binary data.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+        else:
+            raise errors.RadioError(
+                f'ERROR: Model: {model} Ver: {ver} is not supported!')
+        return bs
+    except Exception as e:
+        raise errors.RadioError(e)
+
+
+def get_bands(mode):
+    idx = int(hex(mode)[2:]) - 1
```
kk7ds, line 282: This will fail if the value is ever between 9 and 16:
```
>>> int(hex(12)[2:])
  File "<python-input-2>", line 1, in <module>
    int(hex(12)[2:])
    ~~~^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'c'
```
Isn't this just a no-op for anything that doesn't crash?

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+        data = ident_bytes
+        status.max = sum((t[1] * t[2]) for t in MEMORY_REGIONS if t[3] != 'S')
+        for reg in MEMORY_REGIONS:
+            base, blocks, blk_sz, mod, struc, label = reg
+            if mod != 'S':
+                status.msg = f"Downloading: {label}..."
+                base_int = int(base, 16)
+                addrs = [hex(base_int+i)[2:] for i in range(0, blocks)]
+                for addr in addrs:
+                    cmd = bytes.fromhex('52' + addr.zfill(8))
```
kk7ds, line 311: Please use `struct.pack()` like the rest of the chirp codebase.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+
+    except Exception as e:
+        raise errors.RadioError(f"Exception During Download: {e}")
+    finally:
+        end_programming(serial)
+    return memmap.MemoryMapBytes(data)
+
+
+def write_data(serial, addr, data):
+    pfx = bytes.fromhex('57' + addr.zfill(8))
```
kk7ds, line 328: Same

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+    data = pfx + data
+    chk_int = sum(data) % 256
+    chk_byte = int.to_bytes(chk_int, 1, "little")
```
kk7ds, line 331: `struct.pack()` here.. You can and should use it to pack the whole header in cases like this. It's much easier to reason about.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+            if calc_blksz != blk_sz:
+                raise errors.RadioError(f"Expect:{blk_sz} but" /
+                                        " calc:{calc_blksz} block size")
+            for j in range(blocks):
+                si = j * blk_sz
+                ei = (j+1) * blk_sz
+                data = struc_data[si:ei]
+                # fix up 0xA padding to 0x0 in all DTMF char structs
+                if struc in dtmf_fixup:
+                    data = data.replace(b'\n', bytes(1))
```
kk7ds, line 371: Please don't do this sort of thing during upload - it makes debugging super difficult. You should try to keep the memory map correct all the time and just upload it when it's time to upload. Doing this sort of translation in upload makes it hard to debug why it doesn't end up in the radio like it's stored in CHIRP.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+                addr = addrs[j].zfill(8)  # pad out 0
+                # print(f"Write {struc} -> {addr}: len:{len(data)}\n {data}")
```
kk7ds, line 373: Please remove these leftover commented-out debug prints

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+    # DTMF Encoding E=>* & F=># possible BUG
+    VALID_DTMF = [str(i) for i in range(0, 10)] + ['A', 'B',
+                                                   'C', 'D', 'E', 'F']
+
+    @classmethod
+    def get_prompts(cls):
+        rp = chirp_common.RadioPrompts()
+        rp.experimental = \
+            ('This Retevis RA-25 driver is experimental'
+             'Please report bugs')
```
kk7ds, line 410: You're missing a space between "experimental" and "please". Probably also some punctuation.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+        rf.has_rx_dtcs = True
+        rf.has_cross = True
+        rf.has_tuning_step = True
+        rf.has_bank = False
+        rf.has_settings = True
+        rf.memory_bounds = (1, 200)
+        rf.can_odd_split = False
+        rf.has_offset = True
+        return rf
+    # Do a download of the radio from the serial port
```
kk7ds, line 451: You can remove these comments from the template driver too :)

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+} radio_info;
+
+struct{
+    u8      chans_present_1[32];
+    u8      chans_present_2[32];
+} chans_present;
+struct{
+    u8      chans_scan_1[32];
+    u8      chans_scan_2[32];
+} chans_scanned;
```
kk7ds, line 50: From the look of your uses of these, I think you want to use the `bit` type, which maps individual bits with an index so you don't have to do it yourself.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+            self.set_chan_status(_mem_field, mem.number -
+                                 1, 0)  # deactivate channel
+            _mem_field = _mem_chans_scanned.chans_scan_1 if mem.number - \
+                1 <= 256 else _mem_chans_scanned.chans_scan_2
+            self.set_chan_status(_mem_field, mem.number -
+                                 1, 0)  # skip scanning
+            _mem.fill_raw(b'\xFF')  # clear memory
+            return
+
+        _mem.freq = mem.freq / 10
```
kk7ds, line 556: Should be `// 10` to get an integer.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+        if value == 1:
+            _mem_field[byte_num] |= (1 << bit_pos)
+        else:
+            _mem_field[byte_num] &= ~(1 << bit_pos)
+
+    def get_offset(self, _mem, mem):
+        # gets offset from _mem for ui
+        _hex = hex(_mem.tx_offset)[2:].zfill(8)
+        str_flt = _hex[:3] + '.' + _hex[3:]
+        mem.offset = float(str_flt) * 1000000
```
kk7ds, line 599: I'm not sure what's going on here, but this sort of converting back and forth between strings and int/float is not ideal. This looks to me like `bbcd` or `lbcd` types would be what you're looking for...

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+
+        # Immutables Info
+        info_menu = RadioSettingGroup("immut_info", "Radio Info")
+        group.append(info_menu)
+
+        _warning = """
+            To Upload to the Radio this setting must match the Radio Mode.
+            The Radio Mode can be changed by holding down [V/M] during powerup.
+            Note: Changing the Radio Mode will set Radio memory to defaults."""
+
```
kk7ds, line 1002: This will end up with some spaces in places you don't expect when rendered. Can you make this a single string and let the UI format it? So:
```
_warning = ('To upload the ...'
            'The radio ... ',
   ... )
```

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1098 @@
...
+    @classmethod
+    def match_model(cls, filedata, filename):
+        return False
+
+
+def set_value2(self, value):
+    """RadioSettingValueString Override Fixes
+     - fix for applying specified pad char
+     - ignore pad char in charset checks
+     - UX related to length checks"""
```
kk7ds, line 1085: I'm not sure these are really best handled this way, but you might as well just subclass `RadioSettingValueString` in your file here and implement these in `set_value()` itself. The UI won't care that you subclassed and that's cleaner OO.

chirp/drivers/retevis_ra25.py
ascendr: I explored using bit type however the radio is using LE and bitwise.bitDataElement is BE.

chirp/drivers/retevis_ra25.py
ascendr: RadioSettingValueString set_value() is missing some logic to handle when self._mem_pad_char is set. The pad chars also cause UX issues if one isn't sure to clear the entire field or remove all padding chars before changing the value. This results in length errors. Finally stripping the new line char (passed in when field set and someone presses enter) and the pad char conflict with the charset (if they are not permitted). I sub classed as suggested as it is cleaner. If there is interest in fixing this in the base I can submit a PR.

chirp/drivers/retevis_ra25.py
ascendr: I made several attempts to get around doing this. The 0x00 byte needed by the radio is rejected by char/string data types. If I've missed something, I'm happy to revisit.

chirp/drivers/retevis_ra25.py
kk7ds: `lbit` is LE: https://github.com/kk7ds/chirp/blob/master/chirp/bitwise.py#L21

chirp/drivers/retevis_ra25.py
kk7ds: Right, but `set_value()` is not for applying to the memory object, it's for setting it for the UI. We don't want to pad in the UI, only in the memory object. So either use the pad character yourself when you're setting the value (in `set_settings()` or use `MemSetting` and `apply_to_memobj()`. That's intended to be the newer paradigm, where you just name all the Setting objects with the path into the memory object and they auto-apply. In that situation, the pad character and width are set properly.

chirp/drivers/retevis_ra25.py
ascendr: so I set the bit data type. and MSB was idx 0. i.e  looked like this 0b10000000 instead of needed 0b00000001

chirp/drivers/retevis_ra25.py
kk7ds: Okay, I'm not sure what you're saying. Are the bad values in the DTMF strings bad in existing memory on the radio (my original assumption), or are they wrong after chirp edits them and sets them in memory? If the latter, then you need to fix that in `set_settings()` for sure and not here. That would just mean you're not setting the values in memory like they need to be stored in the radio, and that's what needs to be fixed.

ascendr:
Implemented suggested changes as well as tested the odd DTMF encoding I observed. Confirmed . 0-9 ABCD represented as ascii hex, but * & # represented as ascii hex E and F respectively. Needed to do some translation in get/set settings to handle the differences in Memory vs UI.

chirp/drivers/retevis_ra25.py
kk7ds: Again, not `bit` but `lbit`. Are you asserting that `lbit` doesn't actually work opposite of `bit`? Because there's a unit test that shows it's working...

ascendr: facepalm. I missed lbit type entirely somehow. I'll fix this.

chirp/drivers/retevis_ra25.py
ascendr: The radio expects the DTMF strings need to be padded with 0x0. Using char data type and RadioSettingValueString for the UI doesn't allow for 0x0 as a pad char. I tried setting it with set_settings() however using the Memory Browser to view the field showed it padded with (space) 0x20.

chirp/drivers/retevis_ra25.py
ascendr: I didn't see MemSetting I'll look at it. Unless I did something wrong the UI field is padded and RSVS does not respect setting the mem_pad_char arg.

chirp/drivers/retevis_ra25.py
ascendr: changed to DTMF strings to int type so I can set null bytes via set_settings()

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1058 @@
...
+from chirp import chirp_common, bitwise, memmap, errors, directory
+from chirp.settings import RadioSetting, RadioSettingGroup, \
+    RadioSettingValueBoolean, RadioSettingValueList, \
+    RadioSettingValueInteger, RadioSettingValueString, \
+    RadioSettingValueMap, RadioSettings
+
+import struct
+import logging
+
+log = logging.getLogger(__name__)
```
kk7ds, line 25: This should be `LOG` to be consistent with the rest of the code please.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1058 @@
...
+            raise errors.RadioError(err)
+        return bs
+    except Exception as e:
+        raise errors.RadioError(e)
+
+
+def get_bands(mode):
+    # Radio skips 0xA-0XF, 0x10 == 10 :/
+    try:
+        idx = int(hex(mode)[2:]) - 1
```
kk7ds, line 281: This still makes no sense to me:
```
>>> int(hex(2)[2:])
2
```
Why are you converting to hex, snipping off the `0x` and re-parsing this as a base-ten number (which can break) and then indexing it as such in your own array? Just index your array as  the proper number? And if this ever happens, this will break:
```
>>> int(hex(12)[2:])
  File "<python-input-5>", line 1, in <module>
    int(hex(12)[2:])
    ~~~^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'c'
```

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1058 @@
...
+        status = chirp_common.Status()
+        status.msg = "Connecting to Radio..."
+        radio.status_fn(status)
+        r = start_programming(serial)
+        if not r:
+            raise errors.RadioError('Failed to enter programming mode')
+        ident_bytes = get_ident(serial)
+        data = ident_bytes
+        status.max = sum((t[1] * t[2]) for t in MEMORY_REGIONS if t[3] != 'S')
+        for reg in MEMORY_REGIONS:
```
kk7ds, line 302: I meant to ask this before, but.. you can never ever change this later. Is this what the OEM software does? It reads only portions of the memory? Or does it just only *write* these regions.

In general, it's much better to read the whole thing and only write back the relevant/mutable pieces. If you ever decide you need to read another region in the middle here you'll be stuck and have to consider how not to break everyone's existing image. After more than 15 years of doing this, please take my advice: unless the radio refuses to let you read these regions, you should read them and keep the image file linear. It's fine to only write back certain regions (lots of driver do that).

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1058 @@
...
+    chk_int = sum(data) % 256
+    chk_byte = struct.pack('>B', chk_int)
+    cmd = data + chk_byte
+    r = send_cmd(serial, cmd, 1)
+    if r != b'\x06':
+        err = (f'Received bad ACK from Radio '
+               f'expect 0x06 got {hex(r[0])}')
+        log.error(err)
+        raise errors.RadioError(err)
+    return
```
kk7ds, line 337: This is a nit, but it's conventional in python not to put a naked return like this when you don't need it :)

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1058 @@
...
+
+    def set_chan_status(self, _mem_field, chan_n, value):
+        if chan_n <= 256:
+            _mem_field[0]['bitfield'][chan_n].set_value(value)
+        else:
+            _mem_field[1]['bitfield'][chan_n].set_value(value)
+
+    def get_offset(self, _mem, mem):
+        # gets offset from _mem for ui
+        n = bitwise.bcd_to_int(_mem.tx_offset)
```
kk7ds, line 574: This is the same as `int(_mem.tx_offset)` right?

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1058 @@
...
+                            _i = 0  # offset _mem setting idx for pad chars
+                            for i in range(len(element.value)):
+                                if element.value[i] == '*':
+                                    _val[i - _i] = ord('E')
+                                elif element.value[i] == '#':
+                                    _val[i - _i] = ord('F')
+                                elif element.value[i] in self.VALID_DTMF:
+                                    _val[i - _i] = ord(element.value[i])
+                                else:
+                                    _i += 1
```
kk7ds, line 1048: Seems better to encapsulate this in a helper like you did for the get case above, but... this is what I was asking for - do the proper conversion during set and keep the memory consistent with what is uploaded. Thanks for changing it.

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1058 @@
...
+struct{
+    char    model[8];
+    u8      mode;
+    char    ver[4];
+    u8      ri_unk[2];
+    u8      ack;
+} radio_info;
+
+struct{
+    lbit      bitfield[256];
```
kk7ds, line 40: Much nicer, thanks

kk7ds (commented):
The critical thing for merge is the memory range issue, the rest could be a follow-up...

chirp/drivers/retevis_ra25.py
ascendr: need the BCD conversion here ex \x00\x24\x37\x50 = 243750

kk7ds: But that's what the `bcd` types do.. they `int()` straight to the integer value. The integer cast handler for a bcd array literally does the same thing as this helper. You're already doing this for `mem.freq` which is the same.

I just pulled this down locally and replaced it with `int()` and confirmed. No need for `get_offset()` and `set_offset()`.

chirp/drivers/retevis_ra25.py
ascendr: I hadn't considered issues with backwards compatibility. The radio supports 500 channels but I was only working with the first 200 and thought of adding support later for the additional, hence the skipping of a few regions. I ran several checks and I'm not able to grab contiguous memory from several regions. The radio rejects the request. e.g. I pull 0x22000, but a request to 0x20000 return 0 bytes. I am now grabbing all memory regions and only skipping upload of immutable.

kk7ds: It would be best to attempt to fetch, and if the radio rejects, just fill those sections with zeroes (as long as the radio doesn't abort the download. If it does, then fill with zero anyway). If you never upload them, then no problem. We've had situations before where later firmware revisions use those regions for something. If you skip them in your memory map, you're screwed later, so best to just lay it out contiguously.

Should make your upload and download routines simpler since you don't have to handle the mapping, everything is just linear and 1:1 and then you just need a list of "what to upload".

ascendr: This is a large effort. I would be guessing how large the regions are that I can't retrieve, so the null padded image still may not be the correct size? Is this a hard requirement?

kk7ds: How do you not know? You know the addresses of the all the blocks you are pulling currently, so you know how many bytes are between them and thus know how much to pad, right? You can fully populate the image up to the highest address you know of, and if there turns out to be more later past that address, it's easy to add more at the end. It's not easy to add more in the middle, of course.

ascendr: This radio seems to be different than others. Known regions are 0x0020000 - 0x8002001, so this would be a ~128MB image.

ascendr: Given that a full image is rather large, one strategy could be to read all known regions. Using the Retevis software I've changed every setting and tested all bounds of lists, etc. There are regions that I am not currently reading with the driver ( as I didn't plan to implement them e.g. 5-tone ) that should definitely be added. I could also switch the mmap parse strategy to use seek directives therefore skipping the unused/implemented regions. On upload I would skip the same. This strategy would allow for future driver feature implementation without breaking previous image compatibility.

kk7ds: I think maybe you're misinterpreting the address, as obviously it doesn't have a 128MiB image. Perhaps what you're interpreting as a large address is actually an address and a bank or something? Like maybe the top address is actually `0x80020` with a bank of `0x00` or `0x01` or similar? The protocol here is so very close to tons of other radios that I think it's unlikely that something else is going on and rather that there's something being misunderstood.

Not all OEM software reads the image sequentially, which could be confusing things a bit. How are you capturing the conversation with the OEM software and can you make a trace available somewhere?

ascendr: I'm using tshark. Raw is the tshark output. reassemble is assembled by each command and formatted removes the PC request and the cmd ack from the radio. I keep the radio echo of the address requested, then data, then the checksum byte.  
[captures.zip](https://github.com/user-attachments/files/20112826/captures.zip)

ascendr: A bunch more testing, and I'm not seeing a pattern that can be followed to simply read all memory. The LSB sometimes acts like a bank or address, other times it's ignored and returns the same data whether you request 00-FF. The next byte returns data at a specified length allowing you to request a specific byte length be returned. It appears length requests beyond a 32 byte length are just 00 padded (which makes sense). Bytes 3 & 4 act as bank/address locations. Whether a 3/4 byte bank/address combination has a Byte1 address range is not deterministic. I tried to brute force test this by reading the "base" but most of the unknowns are returning 00s and many reject my ability to write. Without knowing whether 04022000 is 32 bytes or 04022000-04022000FF (8KB) is troubling. I understand the reasoning and agree having a full memory image is superior, however I don't think it's feasible in this case without a lot of guesswork. Through my testing I thoroughly checked all areas affected by radio features, memory channels, and immutable info. I've added all of these areas to the image so that future enhancements won't require altering the image. I would like to update my PR with these changes, and get the new model out for everyone. Attaching some of my testing notes in case if interested. [testing-notes-cmd.txt](https://github.com/user-attachments/files/20127947/testing-notes-cmd.txt)

chirp/drivers/retevis_ra25.py
ascendr: agreed

ascendr:
I was able to get my hands on an Anytone 779UV and it is indeed the same as the Retevis RA25. This New Model now also addresses Issue #8953

ascendr:
Tested and added Anytone AT-779UV as subclass.

ascendr:
Anything needed to get this merged?

chirp/drivers/retevis_ra25.py
```diff
@@ -0,0 +1,1131 @@
...
+        LOG.error(err)
+    return r[5:-1]
+
+
+def get_ident(radio):
+    try:
+
+        bs = send_cmd(radio.pipe, b'\x02', 16)
+        model, mode, ver = struct.unpack('>x7sB4s3x', bs)
+        print(radio.ALLOWED_RADIO_TYPES)
```
kk7ds, line 313: Oops, looks like a leftover debug print.

ascendr:
I've addressed the outstanding issues including refactoring the memory upload/download routines ditching the struct labels in favor of the more standard memory offset based approach.  I've also added VFO support and the Radioddity DB20G model.

chirp/drivers/retevis_ra25.py
ascendr: Refactored memory upload/download and moved to the standard (and more supportable) memory image offset approach.

kk7ds:
Rebased and tagged the two bugs in the commit message

PR #1318, opened 2025-05-08
Add a link in the about box to open online documentation

dforsi:
This PR adds a clickable text linked to https://chirpmyradio.com/projects/chirp/wiki/Documentation and moves the exsting link to the full license text to be between the text box and the Close/Agree buttons:

Schermata_20250508_202847

The idea is that the documentation is closer to the top because might it be read more than once, while the license is probably read once (or never...) and the windows seems to be more balanced.
Putting both links above seems more confusing:

Schermata_20250508_202419


kk7ds:
PR check failure is just a false positive because of the use of the word "license".

PR #1323, opened 2025-05-16
Fix tox CHIRP_TESTIMG option

ascendr:
Fix ability to target a specific driver in tox testing via environment variable


kk7ds:
Okay, but this is old code anyway. You should do something like `tox -e driver -- -k driver_module` to run just the tests from that driver. That means it'll pick up subclasses and things to avoid looking like the base class worked but others didn't (within the same module at least).

PR #1331, opened 2025-05-31
New Model Radioddity GM-30

ascendr:
Feature complete* driver for the Radioddity GM-30. Reference Issue #9237.  
*The only feature from CPS that cannot be safely implemented is radio read/write password protections, this data is read only in the driver.


chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+    if resp[0] != 0x57:
+        raise errors.RadioError(f"Read CMD resp failed got {resp}")
+    if len(resp[echo_ack:]) != exlen:
+        raise errors.RadioError(f"Read CMD resp expect len={exlen} got {resp}")
+    do_ack_ack(serial)
+    return resp[echo_ack:]
+
+
+def enter_prog(serial):
+    cmd = bytes.fromhex('50534541524348')
```
kk7ds: Why not just use the actual byte string you want to send? This is so much harder for someone to debug later...

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+
+
+def do_read_tlmap(serial):
+    # build the data type / addr location map
+    tl_map = {m[0]: 0 for m in TYPE_MAP}
+    for i in range(1, 16):
+        addr = (i << 4 | 0x0f)
+        cmd = struct.pack('>4B', 0xff, addr, 0x00, 0x01)
+        resp = do_read_cmd(serial, cmd, 1)
+        r_int = int.from_bytes(resp, byteorder='big')
```
kk7ds: `struct.parse` ... please.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        if r_int in tl_map:
+            tl_map[r_int] = (i << 4)
```
kk7ds, line 294: Some comments in here and better variable names would really help readability. This is all very obscure. As the person who is still maintaining code other people wrote 15 years ago, please think of the next guy who has to fix this in the future.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+    return data
+
+
+def do_upload_block(serial, loc, offset, radio, status):
+    _mem = radio._memobj.get_raw()
+    for i in range(16):
+        for pre in range(0x0, 0xc1, 0x40):
+            x = offset + (i * 0x100) + pre
+            block = _mem[x:x + 0x40]
+            addr = struct.pack('>BBBBB', 0x57, pre, loc + i, 0x00, 0x40)
```
kk7ds, line 323: Also here.. What is `pre` here? Looks like you're striping across memory for some reason?

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+
+    def sync_in(self):
+        try:
+            data = do_download(self)
+        except errors.RadioError:
+            raise
+        except Exception as e:
+            err = f'Error during download {e}'
+            LOG.error(err)
+            raise errors.RadioError(err)
```
kk7ds, line 454: Don't you want a `finally: exit_prog()` here?

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        self._mmap = data
+        self.process_mmap()
+
+    def sync_out(self):
+        do_upload(self)
```
kk7ds: Don't you need the usual exception translation here?

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        if 0 < mem.number <= 250:
+            _mem = self._memobj.memory[number-1]
+            _name = self._memobj.memnames[number-1]["name"]
+            str_name = self.get_str_name(_name)
+            mem.name = str_name.rstrip()
+            mem.tuning_step = self.VALID_STEPS[0x0]  # step 2.5k only
+            mem.immutable = ["tuning_step"]
+        else:
+            _mem = self._memobj.vfomemory[number-251]
+            mem.name = "VFO-A" if number == 251 else "VFO-B"
```
kk7ds: This is not how you do this. You need to make this a "special" memory. The `mem.name` is the alpha tag for the user's use. If these memories don't have alpha tags, then you should make `mem.name` immutable. The special name is what you're looking for here.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+
+        if _mem.rx_freq.get_raw() == b'\xff\xff\xff\xff':
+            mem.empty = True
+            return mem
+
+        mem.freq = int(_mem.rx_freq) * 10
+
+        if _mem.tx_freq.get_raw() == b'\xff\xff\xff\xff':
+            mem.duplex = ''
+
```
kk7ds: Remove this blank line please - it makes the following `elif` look unrelated.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        elif int(_mem.tx_freq) - int(_mem.rx_freq) > 0:
+            # '+' duplex
+            mem.duplex = '+'
+            mem.offset = (int(_mem.tx_freq) - int(_mem.rx_freq)) * 10
+
```
kk7ds: here too

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        mem.power = self.POWER_LEVELS[_mem.power]
+        mem.skip = "" if _mem.scan else "S"
+
+        if 23 <= mem.number <= 54:
+            mem.offset = 5 * 1000000
+            mem.duplex = '+'
+        else:
+            mem.offset = 0
+            mem.duplex = ""
+        mem.immutable += ['offset', 'duplex']
```
kk7ds: `offset` and `duplex` are immutable on every memory in this radio?

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        if 0 < mem.number <= 54:
+            mem.immutable += ['empty']
```
kk7ds: All memories are non-deletable?

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        if mem.number > 250:
+            mem.immutable += ['empty', 'name']
+        if 0 < mem.number <= 30:
+            mem.immutable = mem.immutable + ['freq']
+        if 8 <= mem.number <= 14:
+            mem.immutable = mem.immutable + ['mode', 'power']
```
kk7ds: This is a ton of immutable stuff. What's the point of actually being able to program this radio with chirp?

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+        return mem
+
+    def _bch_to_int(self, _array):
+        # [0x54,0xc7] -> 12754 [0x72, 0x80] -> 8072
+        value = []
+        for bch in _array:
+            hn, ln = f'{int(bch):x}'.rjust(2, '0')
+            value.append(int(ln, 16))
+            value.append(int(hn, 16))
+        return int("".join(str(i) for i in value[::-1]))
```
kk7ds: Please stop proposing things like this. Doing byte swaps, encodings, etc through strings is inefficient and unconventional. Bitwise should have the types you need for these things.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,976 @@
...
+            txtone = ("DTCS",  _memtxtone % 12000, "R")
+        else:
+            # CTCSS
+            txtone = ("Tone",  _memtxtone / 10, None)
+
+        chirp_common.split_tone_decode(mem, txtone, rxtone)
+
+    def validate_memory(self, mem):
+        if 31 <= mem.number <= 54:
+            # DIY RPTR
```
kk7ds: ?

chirp/drivers/radioddity_gm30.py
ascendr: The radio uses an odd "BCD" encoding where something like [0x54 ,0xc7] needs to translate to int(12754). The bcd encoding in bitwise doesn't allow for an a-f to nibble to convert to base 10. I'm open to other way of doing this if you can kindly point me in the right direction.

ascendr: Would this be preferable? 
def bch_to_int(ba):
    ln = ((ba[1] >> 4) * 1000) + ((ba[1] & 0x0f) * 100)
    hn = ((ba[0] >> 4) * 10) + (ba[0] & 0x0f)
    return (ln+hn)

chirp/drivers/radioddity_gm30.py
ascendr: agreed

chirp/drivers/radioddity_gm30.py
ascendr: agreed, added comments

chirp/drivers/radioddity_gm30.py
ascendr: explained in new code comments

chirp/drivers/radioddity_gm30.py
ascendr: finally: exit_prog() is handled within do_download

chirp/drivers/radioddity_gm30.py
ascendr: yes, added

chirp/drivers/radioddity_gm30.py
ascendr: Removed VFOs. Special memory implementation isn't straight forward due to VFO channels in separate memory region than channels.

chirp/drivers/radioddity_gm30.py
ascendr: yes, this is a FCC certified GMRS radio, so 54 channels are pre-set with respective offsets and duplex. The user can not  change these for compliance

chirp/drivers/radioddity_gm30.py
ascendr: Channels 1-54 are pre-configured and cannot be deleted. This behavior follows the CPS

chirp/drivers/radioddity_gm30.py
ascendr: The radio has 54 pre-configured channels. The frequencies are locked on < 30, duplex and offset are locked on dedicated DIY & RPTR channels. All channels accept Tones. Many accept power change. Channels 55-250 are programmable with VHF-UHF outside of GMRS but RX only. The radio also has DTMF and other settings that are useful.

chirp/drivers/radioddity_gm30.py
ascendr: I've switch to lbcd type and set_raw to handle radios choice of encoding DCS Inverted by adding 12K to the value. note: bitwise int_to_bcd doesn't handle a value this high e.g. 12754 (DCS 754I) needs to be encoded as  [0x54 ,0xc7]

chirp/drivers/radioddity_gm30.py
ascendr: validating that Chirp users only enter valid GMRS repeater frequencies on channels 31-54.

chirp/drivers/radioddity_gm30.py
kk7ds: Right, so this is really just BCD with a high-order bit flag, right? This is a common pattern in other radios, and they just mask off the bits `0xC000` or whatever.

> I've switch to lbcd type and set_raw to handle radios choice of encoding DCS Inverted by adding 12K to the value. note: bitwise int_to_bcd doesn't handle a value this high e.g. 12754 (DCS 754I) needs to be encoded as [0x54 ,0xc7]

I will think about how we can represent this in bitwise since it's a common thing, but yes, adding and subtracting the decimal equivalent is basically doing base ten masking, which is preferable I think.

chirp/drivers/radioddity_gm30.py
kk7ds: Can you put something like that in the comment or just remove it? Obviously from the code below it's clear what is being enforced, but the comment makes it less so, IMHO.

chirp/drivers/radioddity_gm30.py
kk7ds: Okay, so RX-only channels should be `duplex=off` as that is how CHIRP indicates that they're transmit disabled ("duplex" meaning "what happens when I transmit" and "" means "no change" while "off" means "don't"). So those channels should be forced that way to indicate as much to the user.

chirp/drivers/radioddity_gm30.py
kk7ds: Ack, sorry I didn't look at the range and assumed 54 was the top of the memory since it's the highest one mentioned here.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,967 @@
...
+
+    def get_memory(self, number):
+        mem = chirp_common.Memory()
+        mem.number = number
+        _mem = self._memobj.memory[number-1]
+        _name = self._memobj.memnames[number-1]["name"]
+        str_name = self.get_str_name(_name)
+        mem.name = str_name.rstrip()
+        mem.tuning_step = self.VALID_STEPS[0x0]  # step 2.5k only
+        mem.immutable = ["tuning_step"]
```
kk7ds: From the look of this, you should set `has_tuning_step=False` and then you don't need to make it mutable or hard-code it to anything. You should still set `valid_tuning_steps` though.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,967 @@
...
+        raise errors.RadioError(f"Error during upload {e}")
+    finally:
+        exit_prog(serial)
+
+
+@directory.register
+class RadiodditysetGM30(chirp_common.CloneModeRadio):
+    """Radioddity GM30"""
+    VENDOR = "Radioddity"
+    MODEL = "GM30"
```
kk7ds: Looks like this should be "GM-30" like all their other models and their marketing material: https://www.radioddity.com/products/radioddity-gm-30

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,949 @@
...
+            err = f'Error during upload {e}'
+            LOG.error(err)
+            raise errors.RadioError(err)
+
+    def val_or_def(self, memitem):
+        return memitem if memitem is not None else 0
+
+    def get_str_name(self, _name):
+        return "".join(chr(int(i)) if 31 < int(i) < 127
+                       else ' ' for i in _name)
```
kk7ds: Why are you not just using a `char` type for name? It makes debugging so much easier because the name is actually decoded in the developer dialogs.

You can do this filtering on a string with something like:
```
return ''.join(filter(lambda x: x.isprintable(), str(_name))
```
or perhaps more CHIRP-like:
```
return ''.join(filter(lambda x: x in chirp_common.CHARSET_ASCII, str(_name))
```

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,949 @@
...
+        mem.tuning_step = self.VALID_STEPS[_mem.step]
+
+        if _mem.rx_freq.get_raw() == b'\xff\xff\xff\xff':
+            mem.empty = True
+            return mem
+
+        mem.freq = int(_mem.rx_freq) * 10
+
+        if _mem.tx_freq.get_raw() == b'\xff\xff\xff\xff':
+            mem.duplex = ''
```
kk7ds: Doesn't all `0xFF` mean "transmit disabled" and not "simplex"? If so, this should be `mem.duplex='off'`

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,949 @@
...
+        elif int(_mem.tx_freq) - int(_mem.rx_freq) > 0:
+            # '+' duplex
+            mem.duplex = '+'
+            mem.offset = (int(_mem.tx_freq) - int(_mem.rx_freq)) * 10
+        elif int(_mem.tx_freq) - int(_mem.rx_freq) < 0:
+            # '-' duplex
+            mem.duplex = '-'
+            mem.offset = (int(_mem.rx_freq) - int(_mem.tx_freq)) * 10
```
kk7ds, line 506: People will complain if you do this because it makes it harder for them to use separate transmit/receive frequencies. CHIRP has a standard 70MHz split threshold so it would be better if you use `chirp_common.split_to_offset()` to be consistent.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,949 @@
...
+                    'Only GMRS repeater freq. permitted'
+                    ' on channels 31 - 54')]
+        return super().validate_memory(mem)
+
+    def set_memory(self, mem):
+        number = mem.number
+        _mem = self._memobj.memory[number-1]
+        _name = self._memobj.memnames[number-1]
+        newname = [ord(c) for c in mem.name]
+        newname = newname + [0] * (6 - len(newname))
```
kk7ds: `str.ljust()[:6]`

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,949 @@
...
+
+        _mem.rx_freq = mem.freq // 10
+
+        if mem.duplex == "+":
+            _mem.tx_freq = (mem.freq + mem.offset) // 10
+
+        elif mem.duplex == "-":
+            _mem.tx_freq = (mem.freq - mem.offset) // 10
+
+        elif mem.duplex == "" or mem.duplex == "off":
```
kk7ds: These are not the same. Your `get_memory()` makes it look like this radio (like almost all others) wants a `0xFF` fill for "tx disabled" and the rx frequency for "simplex". Even if the radio is enforcing those things, I think it'd be better for consistency if you make this the same, unless there's some reason you're doing this. If there is, please explain in a comment.

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,949 @@
...
+            _mem.tx_freq = _mem.rx_freq
+
+        if mem.mode is not None:
```
kk7ds: This should not happen, so a raise would be appropriate here (and `IndexError` is fine)

chirp/drivers/radioddity_gm30.py
```diff
@@ -0,0 +1,949 @@
...
+        ["A", "B", "C", "D", "*", "#"]
+    ASCII_NUM = [str(i) for i in range(10)] + [' ']
+
+    VALID_STEPS = [2.5, 5.0, 6.25, 10, 12.5, 20, 25, 50]
+
+    def get_features(self):
+        rf = chirp_common.RadioFeatures()
+        rf.has_settings = True
+        rf.has_bank = False
+        rf.has_tuning_step = True
```
kk7ds: You didn't answer my previous question, but do you actually want/need to expose tuning step or not? If not, just say `False` here.

chirp/drivers/radioddity_gm30.py
ascendr: Yes, I had to revisit this as the Radio menu shows a list of tuning steps but I've confirmed selecting anything but 2.5K reverts to 2.5K. The CPS also has this field disabled for all channels. I've set valid tuning steps to only contain 2.5K and set `has_valid_tuning=False`. I'm handling ensuring this is set in memory properly during new memory init.

chirp/drivers/radioddity_gm30.py
ascendr: The CPS uses null byte padding so this was my attempt to keep that state, but since space is allowed char no issues. Tested to confirm.

chirp/drivers/radioddity_gm30.py
ascendr: yes, updated

chirp/drivers/radioddity_gm30.py
ascendr: This radio doesn't handle split as it's an FCC GMRS compliant radio. The only permitted offset is +5MHz

chirp/drivers/radioddity_gm30.py
ascendr: yes, was null byte padding name field but that's unnecessary.

chirp/drivers/radioddity_gm30.py
ascendr: You are correct. The radio enforces but no need to rely on that mechanism. Updated and tested to confirm.

chirp/drivers/radioddity_gm30.py
ascendr: agreed, added a helper for this.

chirp/drivers/radioddity_gm30.py
ascendr: Step logic addressed in another comment. Radio has a list of steps, but choosing one is reverted and CPS has this field disabled. Moved to `has_tuning_step=False`

chirp/drivers/radioddity_gm30.py
kk7ds: Yeah, so if the radio doesn't care, I'd say just set the flag=false and not expose it to the users. TBH, I don't know why some radios expose this for memories, except perhaps so when you do a Mem->VFO transfer it's set for dialing.

chirp/drivers/radioddity_gm30.py
kk7ds: You can null pad with `.ljust(6, '\x00')`

chirp/drivers/radioddity_gm30.py
kk7ds: Oh right, I was thinking for the high channels, but you said no TX there so fair enough.

chirp/drivers/radioddity_gm30.py
ascendr: I was assuming what was in the memory Browser tab is written to the radio. i.e. set() null pads but get() filters it out and space pads. I verified in raw data that the null padding of set() is maintained. Thanks for clarifying.

kk7ds:
Again, please put the proper bug reference syntax in the commit message (not the PR description):
```
Fixes: #9237
```
I just fixed it for you, but it'd be better if you can start doing that yourself.

PR #1332, opened 2025-06-10
Add ignore_bits to bcdDataType

kk7ds:
This makes it easier when radios do things like using the top few
bits of a BCD field as flags (this is common for DCS)

### DIFF
--- a/chirp/drivers/h777.py
+++ b/chirp/drivers/h777.py
@@ -98,6 +98,8 @@ TIMEOUTTIMER_LIST = ["Off", "30 seconds", "60 seconds", "90 seconds",
                      "120 seconds", "150 seconds", "180 seconds",
                      "210 seconds", "240 seconds", "270 seconds",
                      "300 seconds"]
+DTCS_FLAG = 0x80
+DTCS_REV_FLAG = 0x40
 
 
 def _h777_enter_programming_mode(radio):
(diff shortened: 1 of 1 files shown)


kk7ds:
@ascendr How about this? I was going to make this more complicated but I think this is simple and effective enough. Set the `ignore_bits` on one byte of the BCD and then you can `int()` it like normal, but still `get/set/clr_bits()` to access the special ones. I can follow up and convert at least one of the drivers that already does similar stuff to use this.

ascendr:
Assuming our current tx_tone bcdarray is `b'\x87\x54'`
the radio interprets this a "N" polarity i.e. `D754N` (DCS value +8000)
User changes polarity to Inverted (R in Chirp), the radio will want to see` b'\xc7\x54'`
These would be the operations to set in memory
`_mem.tx_tone[0].set_raw(b'\xc0')`
`_mem.tx_tone[0].ignore_bits(0xf0)`
`_mem.tx_tone.set_value(754)`
Then we need to clear the mask so that `get_memory` knows what polarity to display
`_mem.tx_tone[0].ignore_bits(0x00)`
`txint = int(_mem.tx_tone)`

kk7ds:
> Assuming our current tx_tone bcdarray is `b'\x87\x54'` the radio interprets this a "N" polarity i.e. `D754N` (DCS value +8000) User changes polarity to Inverted (R in Chirp), the radio will want to see` b'\xc7\x54'` These would be the operations to set in memory `_mem.tx_tone[0].set_raw(b'\xc0')` `_mem.tx_tone[0].ignore_bits(0xf0)` `_mem.tx_tone.set_value(754)` 

No, don't use `set_raw()`, you do:
```
_mem.tx_tone = 754
if polarity == 'N':
  _mem.tx_tone[0].clr_bits(0x40)
else:
  _mem.tx_tone[0].set_bits(0x40
```

>Then we need to clear the mask so that `get_memory` knows what polarity to display `_mem.tx_tone[0].ignore_bits(0x00)` `txint = int(_mem.tx_tone)`

No, you don't clear the mask, you leave it in place all the time, `int()` it always to get the numerical value (minus the ignored bits) and then test/set the flag bits themselves.

ascendr:
More testing ... We will need to set ignore_bits mask on both set and get tone/memory functions. New memory needs the mask set and get from _mem needs it as well. (+4 func total)

BCD can handle int up to 9000 + DCS, so in this radio's case Tone and DCS N are handled without masking. However by using the  masking we need to add set|clr_bits to all set cases (+4 additional func than without). No net + set|clr code needed on the get side however. 

Finally we need to: 
`_mem.tx_tone[1].clr_bits(0xf0)`
else set_bits is or'd with existing memory. 

Here is an example set implementation with lbcd mem:

```    def set_tones__mem(self, _mem, mem):
        # sets tones in _mem from ui edit
        _mem.tx_tone[1].ignore_bits(0xc0)
        _mem.rx_tone[1].ignore_bits(0xc0)
        ((txmode, txval, txpol),
         (rxmode, rxval, rxpol)) = chirp_common.split_tone_encode(mem)
        if txmode == "":
            _mem.tx_tone.set_raw(b'\xff\xff')
        if rxmode == "":
            _mem.rx_tone.set_raw(b'\xff\xff')
        if txmode == "Tone":
            _mem.tx_tone[1].clr_bits(0xf0)
            _mem.tx_tone[1].set_bits(0x00)
            _mem.tx_tone = int(txval * 10)
        if rxmode == "Tone":
            _mem.rx_tone[1].clr_bits(0xf0)
            _mem.rx_tone[1].set_bits(0x00)
            _mem.rx_tone = int(rxval * 10)
        if txmode == "DTCS" and txpol == "N":
            _mem.tx_tone[1].clr_bits(0xf0)
            _mem.tx_tone[1].set_bits(0x80)
            _mem.tx_tone = int(txval)
        if rxmode == "DTCS" and rxpol == "N":
            _mem.rx_tone[1].clr_bits(0xf0)
            _mem.rx_tone[1].set_bits(0x80)
            _mem.rx_tone = int(rxval)
        if txmode == "DTCS" and txpol == "R":
            _mem.tx_tone[1].clr_bits(0xf0)
            _mem.tx_tone[1].set_bits(0xc0)
            _mem.tx_tone.set_value(txval)
        if rxmode == "DTCS" and rxpol == "R":
            _mem.rx_tone[1].clr_bits(0xf0)
            _mem.rx_tone[1].set_bits(0xc0)
            _mem.rx_tone.set_value(rxval)```

kk7ds:
You're making it too complicated. I just pushed up an example which should be exactly the same as what you need to do. In fact, we could probably re-use the `encode/decode_tone()` helpers directly for this and many other drivers.

ascendr:
Indeed I was over complicating it, thanks for the h777 reference, it was very helpful. I have this implemented and it's passing tox for the GM-30. Once commited to master I'll update my PR with this and the additional changes.

PR #1340, opened 2025-07-25
ft4: Format nonmatching ID messages for better readability

StoyanDimitrov:
Before
Before

After
After

### DIFF
--- a/chirp/drivers/ft4.py
+++ b/chirp/drivers/ft4.py
@@ -345,13 +345,15 @@ def startcomms(radio, direction):
     if id_response != radio.id_str:
         substr0 = radio.id_str[:radio.id_str.find(b'\x00')]
         if id_response[:id_response.find(b'\x00')] != substr0:
-            msg = "ID mismatch. Expected" + util.hexprint(radio.id_str)
-            msg += ", Received:" + util.hexprint(id_response_mod)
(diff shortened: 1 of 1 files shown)


StoyanDimitrov:
It is different problem that both IDs in the message looks the same but in practice they are not.

chirp/drivers/ft4.py
```diff
@@ -345,13 +345,15 @@ def startcomms(radio, direction):
     if id_response != radio.id_str:
         substr0 = radio.id_str[:radio.id_str.find(b'\x00')]
         if id_response[:id_response.find(b'\x00')] != substr0:
-            msg = "ID mismatch. Expected" + util.hexprint(radio.id_str)
-            msg += ", Received:" + util.hexprint(id_response_mod)
+            msg = "ID mismatch.\nExpected:\n{}Received:\n{}".format(
+                util.hexprint(radio.id_str), util.hexprint(id_response_mod)
+                )
```
kk7ds, line 350: I really don't like `str.format` and most of the code in the rest of the tree uses `%s` or `%(name)s`. In the future, I'd prefer you choose one of the latter. Also, we really shouldn't be logging multi-line things here, especially with a platform-specific EOL character. But, we're already doing it and this is definitely an improvement so I'll approve.

kk7ds (commented):
Thanks!

PR #1342, opened 2025-07-29
Add Support for Retevis H777 V4.0 - related to #11787

KC9HI:
for review by Dan

# CHIRP PR Guidelines


KC9HI:
I think I was able to recover from my mistake without losing anything.

Attached is the image that has the first 4 channels cloned over-the-air from another radio. The frequencies don't correspond to any conventional STEP and the tones are not found in the factory software's supported CTCSS/DPL list.
[Retevis_H777-V4_bad_tones_on_chnls1-4.zip](https://github.com/user-attachments/files/21495238/Retevis_H777-V4_bad_tones_on_chnls1-4.zip)

kk7ds:
@KC9HI Let me know if this works for you (i.e. detects and uses the new variant properly) and what you think about the changes.

I looked in detail and determined the "sameness" check is finding all the similar/identical boilerplate between this and the other drivers, with very little code overlap. So, fine to merge with that complaining.

KC9HI:
@kk7ds This does still read saved images. It does not work for downloading or uploading. Screen capture of error messages and debug log files are attached.

Downloading says something about "no attribute 'VENDOR'"
Uploading says something about "object has no attribute '_magic'"

2025-08-01 14_15_54-Error communicating with radio(down)
[retevis_h777v4(down).log](https://github.com/user-attachments/files/21552995/retevis_h777v4.down.log)
2025-08-01 14_20_32-Error communicating with radio(up)
[retevis_h777v4(up).log](https://github.com/user-attachments/files/21552999/retevis_h777v4.up.log)

kk7ds:
Not sure what happened there, but the log wasn't attached. Either way, I got from the errors a couple of typos.

So, my old BF-888 will start programming mode even with the new `C777HAM` program string. Can you confirm if your new one will as well? If so, that will make the detection code a lot simpler and more reliable. Just try downloading from it using the BF-888 and see if at least returns its ident string.

KC9HI:
@kk7ds I don't think I have a classic H777. Retevis tried to find one to send me but didn't have one available. I've only got the models that use the same base name (H, S and Plus).

I did try my original BF-888S and was detected and read into CHIRP just fine.

I don't know why you can't access the log file. I use the paperclip to add them. I just ZIPped them and attached them. Hopefully that will work.

[retevis_h777v4_log.zip](https://github.com/user-attachments/files/21555965/retevis_h777v4_log.zip)

kk7ds:
Yep, the log worked that time, but I had what i needed before.

Sorry I wasn't clear - I meant could you try downloading from the new H777v4 with the old Retevis H777 driver (i.e. without this change applied) to see if it identifies... it won't download but I want to see if it gets past the ident part with the old exchange.

KC9HI:
@kk7ds The h777.py driver will fail with it as-is, but if I temporarily modify the PROGRAM_CMD and IDENT variables to the strings required by the V4 (as shown below), _h777_enter_programming_mode() completes successfully and the cloning process fails when trying to read the memory of the radio (where we would expect it to fail).

@directory.register
class H777Radio(chirp_common.CloneModeRadio):
    """HST H-777"""
    # VENDOR = "Heng Shun Tong (æ’é¡ºé€š)"
    # MODEL = "H-777"
    VENDOR = "Baofeng"
    MODEL = "BF-888"
    # PROGRAM_CMD = b'PROGRAM'
    PROGRAM_CMD = b'C777HAM'
    # IDENT = [b"P3107", ]
    IDENT = [b"\x00\x00\x00\x00\x00\x00\x00", ]
    BAUD_RATE = 9600

KC9HI:
@kk7ds When I watch a serial port capture when trying to download from the H777-V4, it sends the magic to needed to detect/clone an H777 and fails (as expected) and then I assume CHIRP is supposed to then send the magic for the H777-V4 to detect/clone it. However, the serial port capture shows the magic for the H777 ('PROGRAM') instead of the magic for the H777-V4 ('C777HAM').

kk7ds:
> The h777.py driver will fail with it as-is, but if I temporarily modify the PROGRAM_CMD and IDENT variables to the strings required by the V4 (as shown below), _h777_enter_programming_mode() completes successfully and the cloning process fails when trying to read the memory of the radio (where we would expect it to fail).

Okay, my H777 clone literally doesn't care about `PROGRAM_CMD` so I was hoping none of them did and that would make the detection easier. Sounds like your new one does, so no cheating.

Just pushed up another version that takes a slightly different tact. See if this works for your new one. If not, could you just upload a debug log to the referenced bug using CHIRP itself? It's much easier for me to look at it that way versus a zip attachment here.

Speaking of.. your initial commit says "related" but not "fixes" to the bug. Can we change that to fixes when we merge this?

KC9HI:
@kk7ds Downloading and uploading seemed to work OK. 

> Speaking of.. your initial commit says "related" but not "fixes" to the bug. Can we change that to fixes when we merge this?

Yeah... That was my (dumb) way of trying to say don't merge this yet. My plan was once we sorted everything out that "related to" would be changed to "fixes".

Also, can it be changed to just list "H777" in CHIRP? There is nothing on the box, on the radio or in the literature to indicate "V4". So the user will pick H777 and the image will be Retevis_H777-V4_yyyymmdd.img, right? I believe we have other Retevis models that do this.

kk7ds:
Okay cool, thanks for the confirmation.

The old and new driver need to be differently identified in some way if they're going to be different classes. If "v4" won't mean anything useful to people then I'm all for choosing something else. Perhaps just "v2" since this is the second iteration we know about? I wish these manufacturers would stop naming different radios the same thing just for the google hits, but alas...

KC9HI:
@kk7ds I am not talking about the driver. I'm talking about the choice in the CHIRP UI under Retevis. Right no it shows something like "H777 (+V4)". Can't we lose the "(+V4)" and have only "H777"?

kk7ds:
No, the UI adds that automatically from the list of other drivers that are all detected by the parent class. I initially had it not show the variant or model difference, but of course, people were upset because they were looking for things like "TD-H3 (HAM)".

Point being, the class itself _has_ to have a different model/variant to distinguish it for opening images (and other reasons) anyway. And, of course, on upload there's no detection so it shows the full model/variant there.

The GA-510 had no OEM-specified differentiator, they just called it "uh, the new one". The person that first worked on that chose "GA-510v2" so that's kinda the prior art here. Unless you have something better, I'd say we just go with that.

KC9HI:
OK. I struggled with what to call this one. Then when this ticket was created, the originator mentioned V3. I looked at the Amazon listing and it indeed said H777 version 3 and the seller was Retevis so that is what I chose. I told Retevis that was what I was going to do and said "no". They said it was actually version 4 and requested that I use V4 and said they they would update the Amazon listing to match (which they did). 

So what do I need to do now. Update the commit message and force what I have back to github or can you take care of it?

kk7ds:
Okay sounds like v4 is it then.

No worries, I'll update so you don't have to pull and push. Note that I'm going to change it to the same `MODEL=H777` and then `VARIANT=v4` to match the GA510.

Thanks for your help on this!

KC9HI:
@kk7ds Sounds good. Thanks for you help with this. This turned out way better than I ever expected it would.

PR #1345, opened 2025-07-31
Fixes #12088 Add fingerprint for RT-630 V0.09

TrimbleSoftware:
Fixes #12088 Add fingerprint for RT-630 V0.09

Adds additional fingerprint for Radtel RT-630 firmware upgrade V0.09 20250703 to allow existing driver to support new firmware.


TrimbleSoftware:
Just realized I referenced the wrong issue/bug number.  The correct number should be #12088. Edited commit and PR to fix.

PR #1355, opened 2025-08-16
Added Retevis HA1G and HA1UV models 

Tommy83033:
# CHIRP PR Guidelines


tests/py3_driver_testers.txt
```diff
@@ -329,6 +329,8 @@ Retevis_RT9000D_136-174,@KC9HI,8-Dec-2022
 Retevis_RT9000D_220-260,@KC9HI,8-Dec-2022
 Retevis_RT9000D_400-490,@KC9HI,8-Dec-2022
 Retevis_RT9000D_66-88,@KC9HI,8-Dec-2022
+Retevis_HA1G,@KC9HI,8-May-2025
+Retevis_HA1UV,@KC9HI,8-May-2025
```
kk7ds: You're not KC9HI so please don't add this. Also, this file is not generally updated anymore, so please just drop this.

tests/model_stats.tsv
```diff
@@ -513,3 +513,5 @@ TID_TD-H8	194
 TID_TD-H8_GMRS	7
 TID_TD-H8_HAM	2
 Baofeng_BF-888_Custom	6
+Retevis_HA1G    55400
+Retevis_HA1UV   55400
```
kk7ds: This file is generated, please do not add these lines, which skew the stats incorrectly.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1922 @@
...
+struct  {
+  u8 alias[14];  
+  u8 chmode:4,
+     chpro:4;
+  u8 fixedpower:1,
+     fixedbandwidth:1,
+     TailElimination:2,
+     power:2,
+     bandwidth:2;
+  u8 rxfreq[4];
```
kk7ds: From the code below, it looks like this should be `ul32 rxfreq`. You should never need `int.frombytes()`. It looks like there are many other cases here that should be proper data types and not just `u8` arrays.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1922 @@
...
+
+#seekto 0x0D42;
+struct  {
+    u16 chnum;
+    u16 chindex[1027];
+} channeldata;
+
+#seekto 0x154a;
+struct  {
+  u8 alias[14];
```
kk7ds: Looks like this should be `char alias[14]`

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1922 @@
...
+ALARM_LIST=[{"name":"OFF","id":255}]
+DTMFSYSTEM_LIST=[{"name":"OFF","id":15}]
+
+
+class HA1GBank(chirp_common.NamedBank):
+    """A HA1G bank"""
+    def get_name(self):
+        _zone_data=get_zone_index_list(self._model._radio)
+        _zone_list=self._model._radio._memobj.zones
+        if(self.index in _zone_data):
```
kk7ds: You need spaces around most of your operators and these C-style `if` statements don't match the rest of the code (in python it is `if self.index...`

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1922 @@
...
+    crcBytes = current_Page_Byte[-3:-1]
+    dataBytes = current_Page_Byte[:-3]
+    crcStr = calculate_crc16(dataBytes, 2)
+    return crcBytes == crcStr
+
+def Get_HandshakeBytes(currentModel):
+    model_str = currentModel.decode('ascii') if isinstance(currentModel, bytes) else str(currentModel)
+    data_part = b"\x00\x00\x00\x00\x01\x00" +f"{model_str}".encode('ascii')+ b"\x00" * (41-len(model_str))
+
+    # 2. 构建握手数据（不含CRC和结束符）
```
kk7ds: Comments only in english please.

kk7ds (changes requested):
Hi, thanks for working on this. I have only skimmed briefly, as it looks like there is a lot of work left to be done, but left a few comments for you to address as you do. The data types need to be fixed for sure. Also, the code style is very internally inconsistent and inconsistent with the rest of CHIRP. Please clean all of that up to be as close to the other code as possible (please use lowercase/underscore for function/method/variable names and CamelCase for classes).

Thanks!

chirp/drivers/retevis_ha1g.py
Tommy83033: Yes, this is the type of ul32, but this field is little endian, and ul32 receives big endian. I need to use the int.frombytes function to convert the u8 array into a little endian value.

Tommy83033: Of course, I am new to Python. If I use u32 rxfre to receive and convert to little endian value, I currently only know three ways: struct.unpack, int.frombytes and manual shifting.

chirp/drivers/retevis_ha1g.py
Tommy83033: Yes, I have changed it to char type. Thank you for your correction.

tests/py3_driver_testers.txt
Tommy83033: Sorry, I don't know what this file is for, I have deleted these two lines of code

tests/model_stats.tsv
Tommy83033: Sorry, I don't know what this file is for, I have deleted these two lines of code

chirp/drivers/retevis_ha1g.py
Tommy83033: Sorry, I am new to Python and I tend to write code in the style of C#. I will check and modify it.

chirp/drivers/retevis_ha1g.py
Tommy83033: OK, I'll check and fix it.

kk7ds:
You still need to rebase and squash your intermediate commits in order to pass the PR checks. However, you can ignore that for now until after the code has been reviewed.

chirp/drivers/retevis_ha1g.py
kk7ds: `u32` is big-endian, `ul32` is little endian. You must use these and not `int.frombytes()`

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    0x8641,
+    0x8201,
+    0x42C0,
+    0x4380,
+    0x8341,
+    0x4100,
+    0x81C1,
+    0x8081,
+    0x4040,
+]
```
kk7ds: We already have a CRC16 implementation in the tree. Can you re-use that?

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/uvk5.py#L367

If so, we can move that to a central location.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+class HA1GBank(chirp_common.NamedBank):
+    """A HA1G bank"""
+
+    def get_name(self):
+        _zone_data = get_zone_index_list(self._model._radio)
+        _zone_list = self._model._radio._memobj.zones
+        if self.index in _zone_data:
+            zone_name = "".join(
+                filter(_zone_list[self.index].name, NAMECHATSET, 12)
+                )
```
kk7ds: Please do not waste vertical space with close parens on single lines. This should be on the previous line.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+def do_download(self):
+    serial = self.pipe
+    all_bytes = bytearray(self._memsize)
+    handshake_result = handshake(self, serial)
+    if handshake_result == HandShakeStuts.Normal:
+        all_bytes = read_items(self, serial)
+    elif handshake_result == HandShakeStuts.RadioWrong:
+        raise errors.RadioError("Radio not match")
+    elif handshake_result == HandShakeStuts.PwdWrong:
+        raise errors.RadioError("download redio password Wrong")
```
kk7ds: Looks like this should be "Incorrect password for radio" or something. That said, how would the user specify a password? Perhaps this should just be "Radio is password protected"?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+
+
+def do_download(self):
+    serial = self.pipe
+    all_bytes = bytearray(self._memsize)
+    handshake_result = handshake(self, serial)
+    if handshake_result == HandShakeStuts.Normal:
+        all_bytes = read_items(self, serial)
+    elif handshake_result == HandShakeStuts.RadioWrong:
+        raise errors.RadioError("Radio not match")
```
kk7ds: This is not proper english. I assume this should be "Radio model mismatch" or similar?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+        banks = []
+        for bank in self.get_mappings():
+            if memory.number in self._channel_numbers_in_bank(bank):
+                banks.append(bank)
+        return banks
+
+
+def do_download(self):
+    serial = self.pipe
+    all_bytes = bytearray(self._memsize)
```
kk7ds: This is pointless as you overwrite it below on L753

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    handshake_result = handshake(self, serial)
+    if handshake_result == HandShakeStuts.Normal:
+        all_bytes = read_items(self, serial)
+    elif handshake_result == HandShakeStuts.RadioWrong:
+        raise errors.RadioError("Radio not match")
+    elif handshake_result == HandShakeStuts.PwdWrong:
+        raise errors.RadioError("download redio password Wrong")
+    else:
+        raise errors.RadioError("connect Wrong")
```
kk7ds: Presumably this should be something like "Unknown error communicating with radio"?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    exit_programming_mode(self)
```
kk7ds: You should make this a part of a `finally:` block to make sure it runs even if there is an error. Something like:
```
try:
    all_bytes = read_items(...)
except ...:
    ....
finally:
    exit_programming_mode(self)
```

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    handShake_Result = handshake(self, serial)
+    if handShake_Result == HandShakeStuts.Normal:
+        write_items(self, serial)
+    elif handShake_Result == HandShakeStuts.RadioWrong:
+        raise errors.RadioError("Radio not match")
+    elif handShake_Result == HandShakeStuts.PwdWrong:
+        raise errors.RadioError("download redio password Wrong")
+    else:
+        raise errors.RadioError("connect Wrong")
+    exit_programming_mode(self)
```
kk7ds: Same here.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+
+def handshake(self, serial):
+    databytes = b""
+    num = 0
+    while num <= 5:
+        serial.write(get_handshake_bytes(self.MODEL + " "))
+        databytes = serial.read(self.page_len)
+        flag, databytes = handle_connect_ver(
+            databytes, serial
+            )
```
kk7ds: All three of these lines can be combined into one.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+        raise errors.RadioError("download redio password Wrong")
+    else:
+        raise errors.RadioError("connect Wrong")
+    exit_programming_mode(self)
+
+
+def handshake(self, serial):
+    databytes = b""
+    num = 0
+    while num <= 5:
```
kk7ds: This should be a `for` loop if you're iterating. Something like:
```
for attempt in range(5):
```

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+     u8 rev_1:6,
+        DecodingEnable:1,
+        EncodingEnable:1;
+     u8 dtmfstatus;
+     u8 rev_2[12];
+} dtmfs[4];
+"""
+
+
+class HandShakeStuts(Enum):
```
kk7ds: I assume this should be something like "HandshakeStatuses"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    Normal = 0
+    Wrong = 1
+    PwdWrong = 3
+    RadioWrong = 4
+
+
+class Clone_TypeEnum(Enum):
```
kk7ds: "CloneTypes"? Although these appear to be memory region definitions or something so maybe "MemoryRegions" would be better.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    zoneData = 7
+    channelData = 8
+    scanData = 11
+    vfoScanData = 12
+    alarmData = 13
+    dTMFData = 15
+    outFactoryData = 17
+
+
+class File_AddrEnum(Enum):
```
kk7ds: I dunno what this is supposed to be, but "AddrEnum" is probably not necessary.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    status.max = self._memsize
+    data_bytes = self.get_mmap()
+    for item in Clone_TypeEnum:
+        if (
+            item == Clone_TypeEnum.radioHead
+            or item == Clone_TypeEnum.radioInfo
+            or item == Clone_TypeEnum.radioVer
+            or item == Clone_TypeEnum.zoneData
+            or item == Clone_TypeEnum.outFactoryData
+        ):
```
kk7ds: Don't do this, use `in`. Something like:
```
if item in [radioHead, radioInfo, ...]
```

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+        write_item_current_page_bytes(
+            self, serial, item_bytes, item.value, status
+            )
+
+
+def handshake_handle_connect_event(self, dataByte: bytes):
+    if dataByte[14:16] == b"\x00\x01":
+        if dataByte[20] != 1:
+            model_str = (
+                self.MODEL.decode("ascii")
```
kk7ds: `MODEL` must be a `str` not a `bytes` so it makes no sense to `decode()` here nor have the `if` condition below.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+                if isinstance(self.MODEL, bytes)
+                else str(self.MODEL)
+            )
+            radioType = radioType = dataByte[20: 20 + len(model_str)]
```
kk7ds: You're double-assigning here for no reason.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    if not new_bytes:
+        return False, new_bytes
+    for _ in range(max_retries):
+        if rdtp_page_data_crc16_ver(new_bytes):
+            return True, new_bytes
+        chunk = serial_conn.read(chunk_size)
+        if not chunk:
+            return False, new_bytes
+        new_bytes += chunk
+    return False, new_bytes
```
kk7ds, line 503: What this this function doing? It looks like maybe it's reading data until the last byte ends in a valid crc for the early bytes? Please don't do stuff like this. Read the amount of data you're expecting, then calculate the checksum with the expected amount. This sort of algorithm lends itself to non-deterministic behavior, especially across the multiple platforms that CHIRP supports.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+    # 2. sAssembling handshake data
+    handshake_data = (
+        b"RDTP\x01\x00\x00\x00\x00\x00\x00\x00"
+        + len(data_part).to_bytes(2, "little")
+        + data_part
+    )
+
+    # 3. CRC16
+    crc = calculate_crc16(handshake_data, 2)
+    return handshake_data + crc + b"\xff"
```
kk7ds: All of this should use `struct.pack()` for creating these binary blobs.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+        num = 0
+        success = False
+        while num <= 5:
+            serial.write(Get_ReadItemBytes(item, page_index))
+            data_bytes = serial.read(self.page_len)
+            flag, new_data_bytes = handle_connect_ver(data_bytes, serial)
+            if flag:
+                if new_data_bytes[14] != item:
+                    num += 1
+                    time.sleep(0.05)
```
kk7ds: This too is very fragile. Please explain why you need this.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+
+
+def get_read_current_page_bytes(self, item: int, serial, status):
+    page_count = 1
+    page_index = 0
+    item_bytes = b""
+    while page_index < page_count:
+        num = 0
+        success = False
+        while num <= 5:
```
kk7ds: Why are you double-nesting these while loops? Is the second one doing a retry loop or something?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+                num = 0
+                page_index += 1
+                item_bytes += new_data_bytes[20:-3]
+                success = True
+                break
+            success = False
+            num += 1
+            time.sleep(0.05)
+        if not success:
+            raise errors.RadioError("down Wrong")
```
kk7ds: This is not a proper error message. I assume this is just some sort of "radio reported failure" sort of error condition?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,3105 @@
...
+            File_AddrEnum.outFactoryData.value:
+                File_AddrEnum.outFactoryData.value
+            + 2688
+        ]
+    elif item == Clone_TypeEnum.vfoScanData:
+        item_bytes = all_bytes[
+            File_AddrEnum.vfoScanData.value:
+                File_AddrEnum.vfoScanData.value + 68
+        ]
+    return item_bytes
```
kk7ds: What are all these magic numbers added to the enum values? What this this function actually doing? This at least needs a lot of comments.

This is very obscure and hard to understand. Please figure out a better way to do this

kk7ds (changes requested):
Thanks for working on this, but I'm afraid it still needs a lot of work. It would probably be a good idea to do some more reading of other drivers in CHIRP to get a better idea of what the conventions are.

tests/py3_driver_testers.txt
Tommy83033: The codes for HA1G and HA1UV have been verified, but I cannot find the code file retevis_ha1g.py for my model in the chirp/drivers directory.

chirp/drivers/retevis_ha1g.py
Tommy83033: > Why are you double-nesting these while loops? Is the second one doing a retry loop or something?

Yes, will retry if interrupted

chirp/drivers/retevis_ha1g.py
Tommy83033: > This too is very fragile. Please explain why you need this.

flag, new_data_bytes = handle_connect_ver(data_bytes, serial) is used to verify and obtain a complete data packet that conforms to the rules.

if new_data_bytes[14] != item is for further verifying the identifier of the current packet.

chirp/drivers/retevis_ha1g.py
Tommy83033: > We already have a CRC16 implementation in the tree. Can you re-use that?
> 
> https://github.com/kk7ds/chirp/blob/master/chirp/drivers/uvk5.py#L367
> 
> If so, we can move that to a central location.

I checked out the CRC16 implementation for UVK5, but the CRC algorithm involved is different.

The CRC algorithm in retevis_ha1g.py uses the polynomial 0xA001, while uvk5.py uses 0x1021. Additionally, our bit ordering is different - UKK5 follows the XMODEM standard.

chirp/drivers/retevis_ha1g.py
Tommy83033: > What this this function doing? It looks like maybe it's reading data until the last byte ends in a valid crc for the early bytes? Please don't do stuff like this. Read the amount of data you're expecting, then calculate the checksum with the expected amount. This sort of algorithm lends itself to non-deterministic behavior, especially across the multiple platforms that CHIRP supports.

I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > Please do not waste vertical space with close parens on single lines. This should be on the previous line.
I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > Looks like this should be "Incorrect password for radio" or something. That said, how would the user specify a password? Perhaps this should just be "Radio is password protected"?

I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > This is not proper english. I assume this should be "Radio model mismatch" or similar?

I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > This is pointless as you overwrite it below on L753

I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > Presumably this should be something like "Unknown error communicating with radio"?

I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > You should make this a part of a `finally:` block to make sure it runs even if there is an error. Something like:
> 
> ```
> try:
>     all_bytes = read_items(...)
> except ...:
>     ....
> finally:
>     exit_programming_mode(self)
> ```

Yes, that's how it should be implemented. I have modified it.

chirp/drivers/retevis_ha1g.py
Tommy83033: ok, I have modified it.

chirp/drivers/retevis_ha1g.py
Tommy83033: > All three of these lines can be combined into one.

I checked the code and similar code has been modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > This should be a `for` loop if you're iterating. Something like:
> 
> ```
> for attempt in range(5):
> ```

Yes, that's how it should be implemented. I have modified it.

chirp/drivers/retevis_ha1g.py
Tommy83033: > I assume this should be something like "HandshakeStatuses"

yes

chirp/drivers/retevis_ha1g.py
Tommy83033: > MemoryRegions

yes, I have modified it.

chirp/drivers/retevis_ha1g.py
Tommy83033: > I dunno what this is supposed to be, but "AddrEnum" is probably not necessary.

I referred to other models and modified the definition of this block

chirp/drivers/retevis_ha1g.py
Tommy83033: > Don't do this, use `in`. Something like:
> 
> ```
> if item in [radioHead, radioInfo, ...]
> ```

yes, I have modified it.

chirp/drivers/retevis_ha1g.py
Tommy83033: > `MODEL` must be a `str` not a `bytes` so it makes no sense to `decode()` here nor have the `if` condition below.

yes, I have modified it.

chirp/drivers/retevis_ha1g.py
Tommy83033: > You're double-assigning here for no reason.

 yes,I have modified it.

chirp/drivers/retevis_ha1g.py
Tommy83033: > What are all these magic numbers added to the enum values? What this this function actually doing? This at least needs a lot of comments.
> 
> This is very obscure and hard to understand. Please figure out a better way to do this

I rewrote this code with reference to other models and added comments

chirp/drivers/retevis_ha1g.py
Tommy83033: > This is not a proper error message. I assume this is just some sort of "radio reported failure" sort of error condition?

I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: I have modified

chirp/drivers/retevis_ha1g.py
Tommy83033: > This too is very fragile. Please explain why you need this.

The bytes we're getting back here aren't in the expected format — might be some kind of interference or port issue.
I’ll try grabbing it 5 times with a 0.05s delay between each try, just to improve the chances of success.

chirp/drivers/retevis_ha1g.py
kk7ds: It looks to me like this (and the inner) function does this:

1. Sends some data
2. Reads `chunk_size` data
3. Guesses if the result should be CRC'd based on the length and whether or not the CRC matches
4. Returns the data it read and True/False based on what? Whether the CRC matched or not?

This is not a good pattern. You should send the thing you want, and read what you expect. If you expect the thing returned to be CRC'd then do it, don't make this function guess based on the length of the thing read.

Looks like `max_retries` is not used? Please remove.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+    new_bytes = serial_conn.read(chunk_size)
+    if rdtp_page_data_crc16_ver(new_bytes):
+        return True, new_bytes
+    return False, new_bytes
+
+
+def rdtp_page_data_crc16_ver(current_Page_Byte: bytes, chunk_size=1024):
+    byteLen = len(current_Page_Byte)
+    if byteLen <= 13:
+        return False
```
kk7ds, line 510: Please don't do this. If this function checks the CRC for a result, name it accordingly and make it do the thing. As best I can tell, this will return True if the CRC matches, or False if the CRC does not match *OR* False if the amount of data is too small to be CRC'd. That's a very bad behavior for a function to have..

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+    page_count = 255 if chunk_size > 255 else chunk_size
+    pageLen = current_Page_Byte[12] - 6 if current_Page_Byte[12] > 6 else 0
+    pageLen += current_Page_Byte[13] * page_count
+    if byteLen - 23 < pageLen:
```
kk7ds: Please, explain in comments what all these magic numbers are and why the logic is what it is. Don't just re-state the code in english (don't write _what_ .. write _why_)

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+
+def get_read_current_page_bytes(self, item: int, serial, status):
+    page_count = 1
+    page_index = 0
+    item_bytes = b""
+    max_retries = 5
+    retry_delay = 0.05
+    while page_index < page_count:
+        success = False
+        for attempt in range(max_retries):
```
kk7ds: Why are you retrying here? If we're already talking to the radio and it has responded so far, you should not need to retry in the middle of a communication unless you have some other issue with the protocol. The delay on L643 is not a good sign that you're implementing the protocol properly. Is this perhaps left-over from debugging and can now be removed?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+            flag, new_data_bytes = handle_connect_ver(
```
kk7ds: "handle connect ver" does not seem like a good name for this. I think this is "read a block from the radio" right?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+
+
+def Get_ReadItemBytes(data_type: int, page_index: int,
+                      data_code: int = 2):
+    data_part = (
+        b"RDTP\x01\x00\x00\x00\x00\x00\x00\x00\x08\x00"
+        + data_type.to_bytes(1, "little")
+        + data_code.to_bytes(1, "little")
+        + b"\x00\x00\x01\x00"
+        + page_index.to_bytes(2, "little"))
```
kk7ds: This function appears to format a message for the radio that says "read block of data" right?

Please use `struct.pack()` here and not all this bytestring concatenation, like other drivers do. What are all the magic numbers here? 1? 1? 2?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+                              data_buffer: bytes):
+    data_part = (
+        b"RDTP\x01\x00\x00\x00\x00\x00\x00\x00"
+        + (len(data_buffer) + 6).to_bytes(2, "little")
+        + data_type.to_bytes(1, "little")
+        + b"\x00"
+        + page_index.to_bytes(2, "little")
+        + page_count.to_bytes(2, "little")
+        + data_buffer)
+    return data_part + calculate_crc16(data_part, 2) + b"\xff"
```
kk7ds: This looks almost identical to the one above. I think it would be best to combine them into one and take a flag for "read or write"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+        for _ in range(8):
+            if cksum & 0x0001:
+                cksum = (cksum >> 1) ^ 0xA001
+            else:
+                cksum >>= 1
+    return struct.pack("<H", cksum)
+
+
+def calculate_crc16(byte_array, result_length) -> bytes:
+    return get_crc_bytes(byte_array, 0, len(byte_array))
```
kk7ds: This function does not need to exist because it only calls the other function and the other is only used here. Please combine.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+                                   item_bytes: bytes, item):
+    """
+    Put the given region bytes into the main memory array
+    - Uses MEMORY_REGIONS_RANGES to find start and length
+    """
+    if item not in MEMORY_REGIONS_RANGES:
+        LOG.debug(f"Unknown memory Range:{item}")
+    start, length = MEMORY_REGIONS_RANGES[item]
+    item_len = min(len(item_bytes), length)
+    all_bytes[start:start + item_len] = item_bytes[:item_len]
```
kk7ds, line 617: This is modifying the caller's copy of this array. Best to ether modify it (and not return it) or make a copy and return the modified copy.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2111 @@
...
+    return all_bytes
+
+
+def exit_programming_mode(self):
+    serial = self.pipe
+    try:
+        serial.write(Get_ReadItemBytes(111, 0, 0))
```
kk7ds: What is `111` here?

Tommy83033: > What is `111` here?

While the radio’s copying or uploading, it locks into read/write mode—you can’t operate it then.
'111'  is the  reboot  sign.
serial.write(Get_ReadItemBytes(111, 0, 0))，Radio is will  exiting read/write mode and  reboot ,you can operate it then.

kk7ds: Thanks, but my question is rhetorical. I'm telling you that it's impossible for a reader of the code to know these things. The magic numbers should probably be given constants at the top with meaningful names and comments.

chirp/drivers/retevis_ha1g.py
Tommy83033: > Please don't do this. If this function checks the CRC for a result, name it accordingly and make it do the thing. As best I can tell, this will return True if the CRC matches, or False if the CRC does not match _OR_ False if the amount of data is too small to be CRC'd. That's a very bad behavior for a function to have..

 Actually, the validate_packet function (yes, I renamed it) does both validations - it checks the packet length first, then does CRC. Both need to pass.

The length check comes first mainly for performance reasons. It's a quick way to filter out bad packets before we bother with the heavier CRC math. Saves us from doing expensive calculations on garbage data!

chirp/drivers/retevis_ha1g.py
Tommy83033: > Please, explain in comments what all these magic numbers are and why the logic is what it is. Don't just re-state the code in english (don't write _what_ .. write _why_)

I've documented this method with some comments - would appreciate if you could check if they meet the requirements!

def validate_packet(current_packet_Byte: bytes, chunk_size=1024):
    byteLen = len(current_packet_Byte)
    if byteLen <= 13:
        LOG.debug(f"Packet too short: {byteLen} bytes")
        return False  
    """
    I noticed Retevis has another model 
    that sends 250 bytes per packet, 
    and the total packet count calculation is different. 
    So I added this piece of code to handle 
    that case and make it reusable later.
    """
    packet_count = 256 if chunk_size > 255 else chunk_size+1
    """
     I noticed that bytes [12:13] always equal len(current_page_byte) - 17.
     By adding this length check, we can filter out garbage data 
     before it even gets to the CRC calculation - should give us a nice performance boost!
    """
    packet_len = current_packet_Byte[12] - 6 if current_packet_Byte[12] > 6 else 0
    packet_len += current_packet_Byte[13] * packet_count
    if byteLen - 17 < packet_len:
        LOG.debug(f"Packet too short: {byteLen} bytes")
        return False
    crcBytes = current_packet_Byte[-3:-1]
    crcStr = calculate_crc16(current_packet_Byte[:-3], 0, 2)
    return crcBytes == crcStr

chirp/drivers/retevis_ha1g.py
Tommy83033: > Why are you retrying here? If we're already talking to the radio and it has responded so far, you should not need to retry in the middle of a communication unless you have some other issue with the protocol. The delay on L643 is not a good sign that you're implementing the protocol properly. Is this perhaps left-over from debugging and can now be removed?

This implementation was necessary because when using their provided programming cable to copy radio data (especially on Windows 7 systems), I encountered persistent issues where the walkie-talkie would output corrupted data or sometimes provide no response at all. These problems consistently caused data copy failures. By adding a retry mechanism, I was able to significantly reduce the failure rate.

chirp/drivers/retevis_ha1g.py
Tommy83033: > This is modifying the caller's copy of this array. Best to ether modify it (and not return it) or make a copy and return the modified copy.

Yes,You are right, I modified it

chirp/drivers/retevis_ha1g.py
Tommy83033: > This function does not need to exist because it only calls the other function and the other is only used here. Please combine.

Yes,You are right, I modified it

chirp/drivers/retevis_ha1g.py
Tommy83033: > This looks almost identical to the one above. I think it would be best to combine them into one and take a flag for "read or write"

yes,I have combined the two methods

chirp/drivers/retevis_ha1g.py
Tommy83033: > This function appears to format a message for the radio that says "read block of data" right?
> 
> Please use `struct.pack()` here and not all this bytestring concatenation, like other drivers do. What are all the magic numbers here? 1? 1? 2?

I have rewritten the method and implemented it with struct.pack(). The previous method 1 1 2 only gets one or two little end bytes.

chirp/drivers/retevis_ha1g.py
Tommy83033: > "handle connect ver" does not seem like a good name for this. I think this is "read a block from the radio" right?

yes you are right, I renamed it to exchange_block_with_radio

chirp/drivers/retevis_ha1g.py
Tommy83033: > Thanks, but my question is rhetorical. I'm telling you that it's impossible for a reader of the code to know these things. The magic numbers should probably be given constants at the top with meaningful names and comments.

OK, I have modified it

chirp/drivers/retevis_ha1g.py
Tommy83033: > Why are you retrying here? If we're already talking to the radio and it has responded so far, you should not need to retry in the middle of a communication unless you have some other issue with the protocol. The delay on L643 is not a good sign that you're implementing the protocol properly. Is this perhaps left-over from debugging and can now be removed?

I added some comments to explain why I did this.
def get_read_current_packet_bytes(self, item: int, serial, status):
    packet_count = 1
    packet_index = 0
    item_bytes = b""
    # When receiving feedback from the radio,
    # packets might get lost, so we added a retry mechanism.
    max_retries = 5
    # If a packet gets lost, each retry waits 50 ms,
    # which helps lower the chance of losing it again.
    retry_delay = 0.05
    while packet_index < packet_count:

chirp/drivers/retevis_ha1g.py
Tommy83033: > It looks to me like this (and the inner) function does this:
> 
> 1. Sends some data
> 2. Reads `chunk_size` data
> 3. Guesses if the result should be CRC'd based on the length and whether or not the CRC matches
> 4. Returns the data it read and True/False based on what? Whether the CRC matched or not?
> 
> This is not a good pattern. You should send the thing you want, and read what you expect. If you expect the thing returned to be CRC'd then do it, don't make this function guess based on the length of the thing read.
> 
> Looks like `max_retries` is not used? Please remove.

Just like my answer to the following question：
The length check comes first mainly for performance reasons. It's a quick way to filter out bad packets before we bother with the heavier CRC math. Saves us from doing expensive calculations on garbage data!

chirp/drivers/retevis_ha1g.py
kk7ds: I understand why you're doing it, but the method should (a) be named something reasonable and (b) should have a consistent behavior.

chirp/drivers/retevis_ha1g.py
kk7ds: Packets don't get lost in the OS, or in the USB device. They may be sent too quickly for the radio to process, but that means you're not implementing the protocol correctly. Please do not do retries in reads like this.

chirp/drivers/retevis_ha1g.py
kk7ds: I understand why you check length before calculating CRC, my comment is actually about `validate_packet()` but also because of the behavior this function returns. Let me put it this way, I do not think you should be doing any retries in the read or write loops, and if the crc ever fails, you should abort (raise exception) and not just keep retrying. If you need all these retries to make it work, I suspect you are communicating with the radio too quickly.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2122 @@
...
+    retry_delay = 0.05
+    picket_len = self.write_page_len
+    packet_count = math.ceil(len(item_Bytes) / picket_len)
+    if packet_count == 0:
+        return True
+    for i in range(0, packet_count):
+        for attempt in range(max_retries):
+            current_packet_bytes = get_send_packet_bytes(
+                item, i, 0, packet_count,
+                item_Bytes[i * picket_len: (i + 1) * picket_len])
```
kk7ds: Should be `packet_len`?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2122 @@
...
+            flag, new_data_bytes = exchange_block_with_radio(
+                get_send_packet_bytes(item, 0, 2, 1,
+                                      packet_index.to_bytes(2, "little")),
+                serial, self.read_packet_len)
+            if (not flag
+               or (len(new_data_bytes) > 14 and new_data_bytes[14] != item)):
+                time.sleep(retry_delay)
+                continue
+            if packet_index == 0:
+                packet_count = new_data_bytes[18] | (new_data_bytes[19] << 8)
```
kk7ds: `struct.unpack()`

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2122 @@
...
+    # packets might get lost, so we added a retry mechanism.
+    max_retries = 5
+    # If a packet gets lost, each retry waits 50 ms,
+    # which helps lower the chance of losing it again.
+    retry_delay = 0.05
+    while packet_index < packet_count:
+        success = False
+        for attempt in range(max_retries):
+            flag, new_data_bytes = exchange_block_with_radio(
+                get_send_packet_bytes(item, 0, 2, 1,
```
kk7ds: What is 0, 2, 1?

Please do not answer here, add a comment or better a named constant.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,2122 @@
...
+                                      packet_index.to_bytes(2, "little")),
```
kk7ds: Please do not do this, I don't know how many times I have to ask? If this is always 2, then why not just encode it directly: `b'\x0200'` ?

chirp/drivers/retevis_ha1g.py
Tommy83033: > Should be `packet_len`?

yes

chirp/drivers/retevis_ha1g.py
Tommy83033: > Please do not answer here, add a comment or better a named constant.

ok，I defined the constant

chirp/drivers/retevis_ha1g.py
Tommy83033: > Please do not do this, I don't know how many times I have to ask? If this is always 2, then why not just encode it directly: `b'\x0200'` ?

Actually, this is just turning the number into 2 bytes (little-endian). It's basically doing what struct.pack('<H', packet_index) does.

kk7ds: I understand, please either use `struct.pack()` or encode the buffer directly.. I think the latter would be more compact and just as readable.

chirp/drivers/retevis_ha1g.py
Tommy83033: > `struct.unpack()`

ok, I modified it

chirp/drivers/retevis_ha1g.py
Tommy83033: > I understand, please either use `struct.pack()` or encode the buffer directly.. I think the latter would be more compact and just as readable.

ok

chirp/drivers/retevis_ha1g.py
Tommy83033: > Packets don't get lost in the OS, or in the USB device. They may be sent too quickly for the radio to process, but that means you're not implementing the protocol correctly. Please do not do retries in reads like this.

OK, I deleted it.

chirp/drivers/retevis_ha1g.py
Tommy83033: > I understand why you're doing it, but the method should (a) be named something reasonable and (b) should have a consistent behavior.

Yes, the name I used before was a bit inappropriate, so I changed it to validate_packet

chirp/drivers/retevis_ha1g.py
Tommy83033: > I understand why you check length before calculating CRC, my comment is actually about `validate_packet()` but also because of the behavior this function returns. Let me put it this way, I do not think you should be doing any retries in the read or write loops, and if the crc ever fails, you should abort (raise exception) and not just keep retrying. If you need all these retries to make it work, I suspect you are communicating with the radio too quickly.

OK, I deleted the resend mechanism.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+    mem.extra.append(
+        RadioSetting(
+            "autoscan",
+            "Auto Scan System",
+            RadioSettingValueList(AUTOSCAN_LIST,
+                                  current_index=_mem.autoscan)))
+    mem.extra.append(
+        RadioSetting(
+            "alarmlist",
+            "Alarm Sytem",
```
kk7ds: "System" is misspelled

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+                                             _mem.dtmfsignalinglist))))
+    mem.extra.append(
+        RadioSetting(
+            "offlineorreversal",
+            "Talkaround & Reversal",
+            RadioSettingValueList(OFFLINE_REVERSAL_LIST,
+                                  current_index=_mem.offlineorreversal)))
+    mem.extra.append(
+        RadioSetting(
+            "vox", "vox",
```
kk7ds: Please capitalize VOX for display

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            RadioSettingValueBoolean(_mem.vox)))
+    mem.extra.append(
+        RadioSetting(
+            "companding", "compander",
```
kk7ds: "Compander" should be capitalized.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            RadioSettingValueBoolean(_mem.companding)))
+    mem.extra.append(
+        RadioSetting("scramble", "scramble",
```
kk7ds: ...also "Scramble"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+                crc >>= 1
+    return struct.pack("<H", crc)
+
+
+def _get_memory(self, mem, _mem, ch_index):
+    mem.extra = RadioSettingGroup("Extra", "extra")
+    mem.extra.append(
+        RadioSetting(
+            "tottime",
+            "TOT[S]",
```
kk7ds: Please remove "[S]" from the display name. Your elements already include units of seconds in them.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+POWER_LEVELS = [
+    chirp_common.PowerLevel("Low", watts=0),
+    chirp_common.PowerLevel("High", watts=2)]
+TIMEOUTTIMER_LIST = [{"name": "%ss" % (x * 5), "id": x}
+                     for x in range(1, 64, 1)]
+TOTPERSMISSION_LIST = ["Always", "CTCSS/DCS Match",
+                       "Channel Free", "Receive Only"]
+SQUELCHLEVEL_LIST = ["AlwaysOpen", "1", "2", "3", "4", "5", "6", "7", "8", "9"]
+AUTOSCAN_LIST = ["OFF", "Auto Scan System"]
+OFFLINE_REVERSAL_LIST = ["OFF", "Freq Reversql", "Talk Around"]
```
kk7ds: "Reversql" doesn't mean anything to me. What does this mean?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+    {"name": "Prog PTT", "id": 31}]
+
+Fre_Step_List = [
+    {"name": "2.5KHZ", "id": 2500},
+    {"name": "5KHZ", "id": 5000},
+    {"name": "6.25KHZ", "id": 6250},
+    {"name": "8.33KHz", "id": 8330},
+    {"name": "10KHZ", "id": 10000},
+    {"name": "12.5KHZ", "id": 12500},
+    {"name": "25KHZ", "id": 25000}]
```
kk7ds: These should all be "kHz" consistently.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+    MemoryRegions.vfoScanData: (50148, 68),
+    MemoryRegions.alarmData: (50244, 258),
+    MemoryRegions.dTMFData: (51272, 842)}
+
+DTMFCHARSET = "0123456789ABCDabcd#*"
+NAMECHATSET = chirp_common.CHARSET_ALPHANUMERIC + "-/;,._!? *#@$%&+=/<>~(){}]'"
+SPECIAL_MEMORIES = {"VFOA": -2, "VFOB": -1}
+BANDWIDEH_LIST = ["NFM", "FM"]
+POWER_LEVELS = [
+    chirp_common.PowerLevel("Low", watts=0),
```
kk7ds: `watts=0` makes no sense ... what should this be?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            _settings.sk1_long, SIDE_KEY_LIST))
+    common.append(
+        get_radiosetting_by_key(
+            self, _settings, "sk2_short", "Short Press SK2",
+            _settings.sk2_short, short_dict))
+    common.append(
+        get_radiosetting_by_key(
+            self, _settings, "sk2_long", "Long Press SK2",
+            _settings.sk2_long, SIDE_KEY_LIST))
+    opts = ["55HZ", "120°", "180°", "240°"]
```
kk7ds: This should  be "Hz"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            "settings.tailsoundeliminationsfre", "CTC Tail Elimination",
+            RadioSettingValueList(
+                opts, current_index=_settings.tailsoundeliminationsfre)))
+    if _settings.salezone != 2:
+        opts = ["NOAA-%s" % x for x in range(1, 13, 1)]
+        common.append(
+            RadioSetting(
+                "settings.wxch", "NOAA Channel",
+                RadioSettingValueList(opts, current_index=_settings.wxch)))
+    opts = ["1000HZ", "1450HZ", "1750HZ", "2100HZ"]
```
kk7ds: These should be "Hz"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            "settings.singletone", "Single Tone",
+            RadioSettingValueList(opts, current_index=_settings.singletone)))
+    common.append(
+        RadioSetting(
+            "settings.tailsoundeliminationswitch",
+            "Tail Elimination Switch",
+            RadioSettingValueBoolean(_settings.tailsoundeliminationswitch)))
+    common.append(
+        RadioSetting(
+            "settings.rogerbeep", "RogerBeep",
```
kk7ds: Needs a space: "Roger Beep"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+                                  current_index=_settings.poweron_type_2)))
+    opts_dict = get_scan_item_list(self)
+    common.append(
+        get_radiosetting_by_key(
+            self, _settings, "scanlist", "Enable Scan List",
+            _settings.scanlist, opts_dict))
+    short_dict = SIDE_KEY_LIST[:12]
+    common.append(
+        get_radiosetting_by_key(
+            self, _settings, "tk1_short", "TK Short Press",
```
kk7ds: Does "TK" mean "Top Key"? If so, please spell it out.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            _settings.tk1_short, short_dict))
+    common.append(
+        get_radiosetting_by_key(
+            self, _settings, "tk1_long", "TK Long Press",
+            _settings.tk1_long, short_dict))
+    common.append(
+        get_radiosetting_by_key(
+            self, _settings, "sk1_short", "Short Press SK1",
```
kk7ds: SK means "Side Key"? You have it spelled out elsewhere, please do so here as well.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+
+
+def get_dtmf_list(self, dtmf_list):
+    _dtmf_data = self._memobj.dtmfdata
+    _dtmf_list = self._memobj.dtmfs
+    dtmf_count = _dtmf_data.dtmfnum
+    dtmf_count = 4 if dtmf_count > 4 else dtmf_count
+    if dtmf_count <= 0:
+        return
+    dtmf_list.set_shortname("dtmf list")
```
kk7ds: This should be capitalized properly: "DTMF List"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+        rs = RadioSetting(
+            "DecodingEnable_%s" % index, "IsDecodingEnable",
+            RadioSettingValueBoolean(dtmf_item.DecodingEnable))
+        rs.set_apply_callback(set_dtmf_list_callback,
+                              self, i, "DecodingEnable")
+        rsg.append(rs)
+        for j in range(1, 11):
+            rs = RadioSetting(
+                "dtmfsetting.fastcall_%d_%s" % (index, j),
+                "FastCall_%s" % j,
```
kk7ds: This is not a proper display name for a setting. Use `FastCall %s` please.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+    dtmf.append(
+        RadioSetting(
+            "dtmfsetting.revive", "Revive Id",
+            RadioSettingValueString(
+                0, 10,
+                "".join(filter(_dtmf_comm.revive, DTMFCHARSET, 10, True)),
+                False, DTMFCHARSET)))
+    dtmf.append(
+        RadioSetting(
+            "dtmfsetting.bot", "Bot",
```
kk7ds: This should be "BOT"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            RadioSettingValueString(
+                0, 16,
+                "".join(filter(_dtmf_comm.bot, DTMFCHARSET, 16, True)),
+                False, DTMFCHARSET)))
+    dtmf.append(
+        RadioSetting(
+            "dtmfsetting.eot", "Eot",
```
kk7ds: EOT

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+        rs.set_apply_callback(set_dtmf_list_callback, self, i, "codelen")
+        rsg.append(rs)
+        opts = ["None", "A", "B", "C", "D", "*", "#"]
+        rs = RadioSetting(
+            "groupcode_%s" % index, "Group Code",
+            RadioSettingValueList(opts, current_index=dtmf_item.groupcode))
+        rs.set_apply_callback(set_dtmf_list_callback, self, i, "groupcode")
+        rsg.append(rs)
+        rs = RadioSetting(
+            "EncodingEnable_%s" % index, "IsEncodingEnable",
```
kk7ds: This is not a proper display name. Perhaps "Encoding Enabled" ?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+            RadioSettingValueBoolean(dtmf_item.EncodingEnable))
+        rs.set_apply_callback(set_dtmf_list_callback,
+                              self, i, "EncodingEnable")
+        rsg.append(rs)
+        rs = RadioSetting(
+            "DecodingEnable_%s" % index, "IsDecodingEnable",
```
kk7ds: Here too.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+        RadioSetting(
+            "dtmfsetting.showani", "Show ANI",
+            RadioSettingValueBoolean(_dtmf_comm.showani)))
+    dtmf.append(
+        RadioSetting(
+            "dtmfsetting.sidetone", "Side Tone",
+            RadioSettingValueBoolean(_dtmf_comm.sidetone)))
+    dtmf.append(
+        RadioSetting(
+            "dtmfsetting.callid", "Call Id",
```
kk7ds: This and the below similar ones should be "Call ID"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+        "1min", "2min", "3min", "4min", "5min", "15min", "30min",
+        "45min", "1h"]
+    common.append(
+        RadioSetting("settings.backlighttime", "Backlight Time",
+                     RadioSettingValueList(
+                         opts, current_index=_settings.backlighttime)))
+    opts_dict = [{"name": "%s" % x, "id": x} for x in range(1, 16, 1)]
+    common.append(
+        get_radiosetting_by_key(
+            self, _settings, "voxthreshold", "Vox Level",
```
kk7ds: This and below should be capitalized as "VOX"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+
+def get_dtmf_list(self, dtmf_list):
+    _dtmf_data = self._memobj.dtmfdata
+    _dtmf_list = self._memobj.dtmfs
+    dtmf_count = _dtmf_data.dtmfnum
+    dtmf_count = 4 if dtmf_count > 4 else dtmf_count
+    if dtmf_count <= 0:
+        return
+    dtmf_list.set_shortname("dtmf list")
+    for i in range(0, dtmf_count):
```
kk7ds: This is only 1 in the sample image (same for scan list and alarm list). When will this ever be >1?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+        index = i + 1
+        dtmf_item = _dtmf_list[i]
+        rsg = RadioSettingSubGroup("dtmf_list-%i" % i,
+                                   "dtmf_list %i" % (i + 1))
```
kk7ds: This is not a proper name for this section. Should be `DTMF List %i`

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+        opts = ["%ss" % (x + 1) for x in range(0, 16, 1)]
+        rs = RadioSetting(
+            "scanlist.hangtime_%s" % index,
+            "Scan Hang Time",
+            RadioSettingValueList(opts, current_index=_scan_item.hangtime))
+        rs.set_apply_callback(set_scan_list_callback, self, i, "hangtime")
+        rsg.append(rs)
+        rs = RadioSetting(
+            "scanlist.talkback_%s" % index,
+            "TalkBackEnable",
```
kk7ds: Should be "Talk Back Enable"

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+    MemoryRegions.settingData: (92, 100),
+    MemoryRegions.zoneData: (192, 3202),
+    MemoryRegions.channelData: (3394, 43136),
+    MemoryRegions.scanData: (46530, 3618),
+    MemoryRegions.vfoScanData: (50148, 68),
+    MemoryRegions.alarmData: (50244, 258),
+    MemoryRegions.dTMFData: (51272, 842)}
+
+DTMFCHARSET = "0123456789ABCDabcd#*"
+NAMECHATSET = chirp_common.CHARSET_ALPHANUMERIC + "-/;,._!? *#@$%&+=/<>~(){}]'"
```
kk7ds: I think this is supposed to be `NAMECHARSET` ?

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+    if index < len(_alarm_list):
+        if name == "alarmstatus":
+            alarm_index_dict = get_alarm_index_list(self)
+            if value == 1 and index not in alarm_index_dict:
+                alarm_index_dict.append(index)
+            elif value == 0 and index in alarm_index_dict:
+                alarm_index_dict.remove(index)
+            set_alarm_index_list(self, alarm_index_dict)
+        elif name == "name":
+            value = filter(value, NAMECHATSET, 12, True)
```
kk7ds: This is forcing the stored value to be uppercase, but the test images have mixed-case values in them, and the setting itself allows the user to enter lowercase characters. This is not okay. The same happens for DTMF and Scan list names.

chirp/drivers/retevis_ha1g.py
```diff
@@ -0,0 +1,1867 @@
...
+        index = i + 1
+        dtmf_item = _dtmf_list[i]
+        rsg = RadioSettingSubGroup("dtmf_list-%i" % i,
+                                   "dtmf_list %i" % (i + 1))
+        rs = RadioSetting(
+            "dtmflist.name_%s" % index, "DTMF Name",
+            RadioSettingValueString(
+                0, 12,
+                "".join(filter(dtmf_item.name, NAMECHATSET, 12)),
+                False, NAMECHATSET))
```
kk7ds: This should be marked as volatile so that the UI knows it needs to reload some data when you change it: `rs.set_volatile(True)`. Same for scan and alarm list names.

kk7ds (changes requested):
Tommy, please *please* try to go through this with a fine-toothed comb and fix typos, inconsistencies, etc. I shouldn't have to make so many comments about small items like this after so much review, but there is still a lot here to clean up.

chirp/drivers/retevis_ha1g.py
Tommy83033: > This should be marked as volatile so that the UI knows it needs to reload some data when you change it: `rs.set_volatile(True)`. Same for scan and alarm list names.

Ok, I have a quick question. When I set rs.set_volatile (True) for the DTMF System's name field and assign values repeatedly, the UI says it needs to refresh. But the data dropdown for DTMF System under Memories just won't refresh along with it. Any idea how to get that to sync up?

kk7ds: Yeah, I need to do some plumbing in the UI to make it refresh across from the settings editor to the memory editor. So, mark this as volatile (since it is) and I will work on getting it fixed so that it makes the memory editor refresh as well.

chirp/drivers/retevis_ha1g.py
Tommy83033: > This is only 1 in the sample image (same for scan list and alarm list). When will this ever be >1?

The DTMF list can have a maximum of 4 valid entries, but the walkie-talkie doesn't display all of them by default. It depends on how many the customer has set up. If the customer sets 1, only 1 will be shown when downloaded; if they set 4, then all 4 will be displayed.  (same for scan list and alarm list)

Tommy83033: > > This is only 1 in the sample image (same for scan list and alarm list). When will this ever be >1?
> 
> The DTMF list can have a maximum of 4 valid entries, but the walkie-talkie doesn't display all of them by default. It depends on how many the customer has set up. If the customer sets 1, only 1 will be shown when downloaded; if they set 4, then all 4 will be displayed. (same for scan list and alarm list)

Since the channel only loads valid DTMF lists, if I let users check to enable or disable entries themselves, the related data in the channel list won't refresh properly. That's why I'm only loading the valid data here.

chirp/drivers/retevis_ha1g.py
Tommy83033: > Yeah, I need to do some plumbing in the UI to make it refresh across from the settings editor to the memory editor. So, mark this as volatile (since it is) and I will work on getting it fixed so that it makes the memory editor refresh as well.

ok

chirp/drivers/retevis_ha1g.py
Tommy83033: > This is forcing the stored value to be uppercase, but the test images have mixed-case values in them, and the setting itself allows the user to enter lowercase characters. This is not okay. The same happens for DTMF and Scan list names.

Yeah,my mistake earlier – this should indeed allow mixed case, and I've now made that change.

chirp/drivers/retevis_ha1g.py
Tommy83033: > `NAMECHARSET`

Yeah

chirp/drivers/retevis_ha1g.py
Tommy83033: > SK means "Side Key"? You have it spelled out elsewhere, please do so here as well.

OK

chirp/drivers/retevis_ha1g.py
Tommy83033: > Does "TK" mean "Top Key"? If so, please spell it out.

ok

chirp/drivers/retevis_ha1g.py
Tommy83033: > Needs a space: "Roger Beep"

ok

chirp/drivers/retevis_ha1g.py
kk7ds: I'm not sure what you mean.. the user has to enable the other DTMF entries in the radio before they can be edited in CHIRP? People will just file bugs to complain about that.

All you need to do is let them enable them and protect against the bad data with good default values for when someone enables these.

chirp/drivers/retevis_ha1g.py
kk7ds: Since you have marked these as resolved _and_ pushed code since, but not fixed this and all the other things, I assume you're not done yet. So please poke me when it's time to review again.

chirp/drivers/retevis_ha1g.py
Tommy83033: > I'm not sure what you mean.. the user has to enable the other DTMF entries in the radio before they can be edited in CHIRP? People will just file bugs to complain about that.
> 
> All you need to do is let them enable them and protect against the bad data with good default values for when someone enables these.

I was trying to add a default field that users can enable themselves. But here's the thing—once they enable it, the data doesn't refresh in the channel list. That's definitely a bug.

Tommy83033: > I'm not sure what you mean.. the user has to enable the other DTMF entries in the radio before they can be edited in CHIRP? People will just file bugs to complain about that.
> 
> All you need to do is let them enable them and protect against the bad data with good default values for when someone enables these.

I think this part can be saved for later optimization. It doesn’t interfere with the machine’s normal operation, and there’s no such thing as a perfect program right out of the gate—I’ll keep maintaining it.

kk7ds: I'm not asking for perfect, I'm asking for consistent. So just remove the DTMF (and alarm and scan) lists entirely and we can add them back when they work.

Tommy83033: > I'm not asking for perfect, I'm asking for consistent. So just remove the DTMF (and alarm and scan) lists entirely and we can add them back when they work.

Okay, I temporarily commented out the code for DTMF List, Scan List and Alarm List displays. We’ll enable it again after fixing the remaining bugs

kk7ds: Thanks, but please don't just comment it out, it's confusing to have code in the tree that looks like it is effective but isn't. Just put it into a new commit and submit that as another PR on top of this one if you want to go ahead and start it. Although I see you're using your master branch so that might be hard. `git-stash` maybe then...

Tommy83033: > Thanks, but please don't just comment it out, it's confusing to have code in the tree that looks like it is effective but isn't. Just put it into a new commit and submit that as another PR on top of this one if you want to go ahead and start it. Although I see you're using your master branch so that might be hard. `git-stash` maybe then...

Okay

PR #1361, opened 2025-08-22
Update Russian translation

OlesyaGerasimenko:
Updated the Russian translation.

### DIFF
--- a/chirp/locale/ru.po
+++ b/chirp/locale/ru.po
@@ -1,17 +1,17 @@
-# Olesya Gerasimenko <translation-team@basealt.ru>, 2023.
+# SPDX-FileCopyrightText: 2023, 2025 Olesya Gerasimenko <translation-team@basealt.ru>
 msgid ""
 msgstr ""
 "Project-Id-Version: CHIRP\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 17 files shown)


chirp/locale/ru.po
```diff
@@ -194,7 +200,8 @@ msgstr ""
 msgid ""
 "1. Turn radio off.\n"
 "2. Connect cable\n"
-"3. Press and hold in the MHz, Low, and D/MR keys on the radio while turning it on\n"
+"3. Press and hold in the MHz, Low, and D/MR keys on the radio while turning"
+" it on\n"
```
kk7ds: I believe this sort of thing is why the PR check is failing. Are you using a po-specific editor tool to make the changes? I think I can just reformat this with the tools we're using and keep your translation changes. Since I can't really tell, I can push that up on top of this if you want, but I'd like it if you could confirm that it keeps all your changes. Is that okay?

OlesyaGerasimenko: Yes, I'm using Lokalize for making changes. It would be great if you reformatted and kept the changes. Thanks in advance!

kk7ds:
Okay, this is my best attempt to de-conflict it. You will have to tell me if the changes were properly copied over.

OlesyaGerasimenko:
Looks good to me, all the changes are there.

PR #1362, opened 2025-08-25
Ipython support

machshev:
Adds support for using ipython interactive console instead of the default, if ipython is installed. This is really useful for things like tab completion, which makes debugging drivers easier.

### DIFF
--- a/chirp/wxui/main.py
+++ b/chirp/wxui/main.py
@@ -1809,14 +1809,27 @@ class ChirpMain(wx.Frame):
 
     def _menu_interact_driver(self, event):
         LOG.warning('Going to interact with radio at the console')
+
         radio = self.current_editorset.current_editor.radio
-        import code
         locals = {'main': self,
                   'radio': radio}
+
         if self.current_editorset.radio != radio:
             locals['parent_radio'] = self.current_editorset.radio
-        code.interact(banner='Locals are: %s' % (', '.join(locals.keys())),
-                      local=locals)
+
+        banner = 'Locals are: %s' % (', '.join(locals.keys()))
+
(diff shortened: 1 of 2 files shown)


machshev:
The `.nvim.lua` I'm using locally is neovim config to disable LSP auto formatting. All the other projects I'm working on use `ruff` with auto formatting, so every file I touch was getting auto formatted using the default `ruff` settings. Hence adding it to the `.gitignore`.

There seems to be provision for other IDE config, so hopefully this is okay. If not I can just maintain it locally.

chirp/wxui/main.py
```diff
@@ -1809,14 +1809,27 @@ def _menu_reload_driver(self, event, andfile=False):
 
     def _menu_interact_driver(self, event):
         LOG.warning('Going to interact with radio at the console')
+
+        main = self
```
kk7ds: Are you doing this because the `embed()` call automatically imports locals from the calling scope or something? Are you not able to pass a subset to make it equivalent? I'm specifically passing locals (instead of `locals()`) to keep this restricted to specific uses so it would be preferred to keep it equivalent.

machshev: Yes the ipython version just inherits the locals within the function that it is called from. There are kwargs that can be passed, but I couldn't obviously see how to set the locals explicitly.

I can investigate further, but suspect it's a case of having to go read the source code to find a lower level API. The docs didn't seem that helpful with lots of indirection and abstraction.

machshev: Had to read the source code to work it out, but yes I've updated with an explicit local namespace.

PR #1365, opened 2025-08-28
baofeng_uv17Pro.py: Add Baofeng UV-5R Mini - fixes #12091

KC9HI:
# CHIRP PR Guidelines


KC9HI:
This model is a good example of why care need to be taken when some says that CHIRP reads from and writes to their 'unsupported' radio model OK with a existing supported model. This models requires the UV-17 Pro GPS magic to initiate cloning and needs the UV-17 Pro to compatible with download/upload. Only doing that appears to read from and write to the radio just fine. However, the memory ranges are different, valid bands are different and there are a number of differences in the per-channel and global settings that should also be accounted for.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -2139,3 +2144,87 @@ def get_settings_pro_dtmf(self, dtmfe, _mem):
         rs = RadioSetting("downcode.code", "Down Code", val)
         rs.set_apply_callback(self.apply_code, self._memobj.downcode, 16)
         dtmfe.append(rs)
+
+
+@directory.register
+class UV5RMini(UV17Pro):
+    """Baofeng UV-5R Mini"""
+    VENDOR = "Baofeng"
+    MODEL = "UV-5R-MINI"
```
kk7ds: Would it be okay to say "UV-5R Mini" here instead of using the extra dash and capitalizing "MINI"? Just looks better to my eye, unless this is how it's formatted on the radio itself or marketing materials.

KC9HI: Yes. I would rather have the 'space' but I was thinking that there was a requirement the have a dash or an underscore in the model name and I thought the '-' looked better than the '_'. So please change it.

kk7ds: There might have been a requirement a long time ago because of some tooling stuff around matching models or something, but we've got models in the tree with spaces now without any trouble. And yeah, I don't like the look of the underscore either.

I'd call you an old timer because you've been around long enough to be remembering rules we used to have, but that would make me old too, so... :-D

PR #1366, opened 2025-08-28
retevis_rt21.py: Add Retevis RT86S - fixes #11833

KC9HI:
# CHIRP PR Guidelines


chirp/drivers/retevis_rt21.py
```diff
@@ -1220,15 +1292,15 @@ def set_memory(self, mem):
                               "RT76",
                               "RT86",
                               ]:
-                _mem.set_raw("\xFF" * 13 + _rsvd)
+                _mem.set_raw(b"\xFF" * 13 + _rsvd)
```
kk7ds, line 1295: I know you're just converting these from the previous (and thanks) but just FYI, we have `fill_raw()` now, which takes one byte value and fills the thing with that byte instead of needing the count.

KC9HI: Wouldn't that fill all 16 bytes? I need to make sure that the last 3 bytes are set to the value stored in _rsvd. Unfortunately the radio uses the last 3 bytes of each memory row to store the radio's global settings. I've been meaning to ask you if there is a better way to do this because this winds up causing negative seeks building the MEM_FORMAT. There are at least 13 radios that store their global settings this way. I would definitely like to learn a better way of doing this (if it is possible).

kk7ds: Oh yeah sorry I totally missed that this was set raw plus the `_rsvd` value.

Looking at this, I'm confused now. The `reserved` field seems to be a `u8 reserved[3]` (as you say, three bytes). Are you just trying to clear the whole memory structure _except_ the last three bytes?

KC9HI: I don't know if I can explain this right, but this is what I understand.
The first 13 bytes are the channel memory. So only that could be cleared. The last 3 bytes are global memory and the must not be changed in the memory editor. They are used for global settings so I copy them to _rsvd and then when I clear the memory I copy them back.

I probably should have run this by you when I was doing the first radio that worked this way. It is mostly el cheapo 16 channel radios but it is expanding into some of the 22 and 30 channel GMRS radios like this one.

kk7ds: Ugh, yeah that's a bizarre choice by the radio designers.

What if `fill_raw()` took a length? Like `fill_raw(b'\xFF', length=13`? That would only work if it's at the end though, so useful here but maybe not in the long run. Another way would be to just wrap it twice:
```
struct memory_stupid {
  struct memory {
    lbcd rxfreq[4];
    ....
  } mem_real;
  u8 reserved[3];
};
struct memory_stupid memories[256];
```
Then you'd just grab the inner thing instead and use or fill that:
```
_mem = self._memobj.memories[num].mem_real
```
and that would always exclude the last three bytes.

Make sense?

PR #1370, opened 2025-08-30
ft8800: fix memory issues

machshev:
Various fixes for the FT-8800:
- There are 512 channels per sub-device not 500
- For the 8800 the max step is 50, not 100
- Correctly set the mode AM, FM, NFM (actually two settings AM/FM and Narrow/Wide).
- Also fix up the flags array length 250 -> 256

All of this was tested on my FT8800 and I can now correctly program the radio memory channels without the strange behaviors I was seeing before.

### DIFF
--- a/chirp/drivers/ft7800.py
+++ b/chirp/drivers/ft7800.py
@@ -144,7 +144,8 @@ u8 checksum;
 
 MODES = ["FM", "AM", "NFM"]
 DUPLEX = ["", "", "-", "+", "split"]
-STEPS = [5.0, 10.0, 12.5, 15.0, 20.0, 25.0, 50.0, 100.0]
+STEPS = (5.0, 10.0, 12.5, 15.0, 20.0, 25.0, 50.0, 100.0)
+STEPS_8800 = (5.0, 10.0, 12.5, 15.0, 20.0, 25.0, 50.0)
 SKIPS = ["", "S", "P", ""]
 
 CHARSET = ["%i" % int(x) for x in range(0, 10)] + \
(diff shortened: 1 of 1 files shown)


chirp/drivers/ft7800.py
```diff
@@ -880,7 +895,7 @@ def get_features(self):
         rf = FTx800Radio.get_features(self)
         rf.has_sub_devices = self.VARIANT == ""
         rf.has_bank = True
-        rf.memory_bounds = (1, 500)
+        rf.memory_bounds = (1, 512)
```
kk7ds, line 898: The extra memories are not normal channels right? They're specials, call channels, etc? If so, they should be implemented as such and not be presented as "channel 507" which won't match what the radio does. Look for some drivers with `valid_special_chans=` set for an example.

machshev: Thanks for the pointer re special channels!

However as far as I can see these are normal "standard" channels, on my radio they appear as 1-512 and act as standard channels as per the manual. There are also an additional 10 special channels labeled as `L1/U1` -> `L10/U10` and a further 5 "home channels".

I'm still working out the special channel types, and might find time to add those later.

See [YAESU FT-8800-R Manual](https://www.qrzcq.com/pub/RADIO_MANUALS/YAESU/YAESU--FT-8800-R-User-Manual.pdf) page 34 in the PDF reader (32 as marked on the pages themselves).

> Independent “Regular” Memory Channels for the “Main” and “Sub” bands, each con-
sisting of:
 512 “Standard” memory channels, numbered “001” through “512.”
 Five Home channels, providing storage and quick recall of one prime frequency on
each operating band.
 Ten sets of band-edge memories also known as “Programmable Memory Scan” chan-
nels, labeled “L1/U1” through “L10/U10.”
 Ten Memory Banks, labeled “BANK1” through “BANK10.”

kk7ds: Okay fair enough.. not sure how that got missed for so long :)

machshev:
Turns out the `FT9800` driver was coupled to the `FT8800` driver. So changing the `FT8800` driver broke the downstream driver. As far as I can see there is no reason for the `FT9800` to subclass the `FT8800` and much of the inherited behavior is then explicitly reverted back to the common base class `FTx800`.

So I've decoupled the `FT9800` driver as best as I can, which should mean the two radios are now independent.

chirp/drivers/ft7800.py
machshev: Guess not everyone has every radio to test everything ;)

PR #1371, opened 2025-08-30
FT-8800: add settings support

machshev:
Add support for settings on the FT-8800.

### DIFF
--- a/chirp/drivers/ft7800.py
+++ b/chirp/drivers/ft7800.py
@@ -21,7 +21,7 @@ from chirp.drivers import yaesu_clone
 from chirp import chirp_common, memmap, directory, bitwise, errors
 from chirp.settings import RadioSetting, RadioSettingGroup, \
     RadioSettingValueList, RadioSettingValueBoolean, \
-    RadioSettingValueString, RadioSettings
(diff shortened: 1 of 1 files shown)


kk7ds:
Are you planning to finish this and mark it as not a draft?

machshev:
Sorry, I've been busy recently on other projects. Installed the radio in the car in the boot behind a panel, so it's not so easy to get to now. The head unit is accessible but not the programming side. Unfortunately the serial cable it too long to work. The plan is to make up a USB extension cable using a USB to ethernet adaptor to be able to program it without taking half the car apart each time.

If you would prefer you could merge as is, and I'll open a new PR for further improvement. The current state is better than the original, although still not fully functional.

What would you like to do?

kk7ds:
Okay I have a few complaints with what's here but nothing serious. I was waiting until you marked it as ready. If you want to do that, I can just push up tweaks for my OCD things and then merge if I'm happy with it.

machshev:
Thanks! Just opened up the laptop, rebased on the latest master and pushed.
Hopefully I'll get round to working out the other settings at some point. In the meantime this adds some extra support for settings that were not previously available.

PR #1373, opened 2025-09-01
bff8hppro: Fix high power level watts

kk7ds:
This radio is advertised as 10W on high, so update the driver to match.

Fixes: #12145

### DIFF
--- a/chirp/wxui/memedit.py
+++ b/chirp/wxui/memedit.py
@@ -1458,6 +1458,11 @@ class ChirpMemEdit(common.ChirpEditor, common.ChirpSyncEditor):
                 self._grid.GetRowLabelValue(row)))
             return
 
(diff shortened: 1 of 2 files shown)


kk7ds:
@kc9hi Any problem with this?

The product pages for this claim 10W, and someone complained about it being 8W in the driver. I assume this was just a mistake, but if there's some reason for it let me know!

KC9HI:
I don't remember. There is no mention of the TX power level in the user manual. I have nothing in the pre-production specifications. I may have measured it on the pre-production sample unit that I had, it may have been told to me or I may have just put something in to facilitate development of the driver. The product pages so appear to claim 10W so I have no problem with it being changed.

kk7ds:
Cool, I figured as much. Thanks!

PR #1374, opened 2025-09-02
fix: test runner shebang

machshev:
Fix the shebang for fast-driver script.

The only reason this currently works with a broken shebang is because it's called with `python tools/fast-driver.py` in `tox.ini`. Whereas if you just do `./fast-driver.py ...` it fails saying it can't find `Python3` (should be python3).

### DIFF
--- a/tools/fast-driver.py
+++ b/tools/fast-driver.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env Python3
+#!/usr/bin/env python3
 
 import glob
 import os


kk7ds:
> The only reason this currently works with a broken shebang is because it's called with `python tools/fast-driver.py` in `tox.ini`. Whereas if you just do `./fast-driver.py ...` it fails saying it can't find `Python3` (should be python3).

...sure but that's also the only place it needs to or should be run :)

machshev:
Personally I like to run tests and linting locally before pushing my changes. Which means fewer loops around the github CI. I'm still getting familiar with how everything is setup with this project.

Having the ability to run pytest on just the files I've changes is really nice. Personally I'd prefer it if the project could use an auto formatter, rather than just a lint checker which checks the formatting. Just means developers can setup their IDE to automatically format the code on save. I've had to disable that in neovim so I'm not accidentally reformatting whole files for a small change.

Problem is, that means I now have to fix up a list of small lints manually. It's not the end of the world, but it's so much nicer to be able to forget the formatting and focus on the actual functionality.

If you did decide to go down this route, then I'd suggest `black` or better yet `ruff` (Rust implementation that is really fast, and also provides an LSP). They both more or less create the same format, there are some config options if you really don't like the default. Anyway, if you haven't thought about it already, then perhaps consider it. It would certainly make it easier for contributes, in my view at least.

In my projects I use `ruff` which is super fast and can replace `flake8`, `black`, and a whole host of other linters in one tool. You specify how many lints you want to enable, the default is very minimal. The ruff lsp then provides linting and auto formatting in one.

If you want I can have a go at putting together a PR?

kk7ds:
> Personally I like to run tests and linting locally before pushing my changes. Which means fewer loops around the github CI. I'm still getting familiar with how everything is setup with this project.

You should, of course be running all the tests locally first, and you should do it using the prepared environments just like the github CI does. In this example:
```
tox -e fast-driver
```
If you want to run just one test or driver specifically, use:
```
tox -e driver -- -k test_foo_bar (or whatever pytest args you want)
```
And of course, for the lints:
```
tox -e style
```
Just running `tox` will hit them all.

> Having the ability to run pytest on just the files I've changes is really nice. Personally I'd prefer it if the project could use an auto formatter, rather than just a lint checker which checks the formatting. Just means developers can setup their IDE to automatically format the code on save. I've had to disable that in neovim so I'm not accidentally reformatting whole files for a small change.

I am forced against my will to use a formatter in my day job and I *refuse* to use one here where I get to choose. I hate them and the code (format) they spit out.

> If you want I can have a go at putting together a PR?

Please do not :)

PR #1376, opened 2025-09-05
Fixes #12108 and several undefined subroutines. Now handles all YSF digital "Rooms" in each "Category"

Darieb:
Yaesu FT-1D, FT2D, FT3D drivers: Settings for Digital -> WIres-X (YSF) will now show all selected YSF channels (up to 20; they're called "Rooms" in Yaesu manuals) in the Radio's five Digital "Categories". These are similar to memory Banks but for YSF digital channels.

At least one rarely-used settings group (Backtrack) called an erroneously-named function. That's been fixed.

No new image required.


Darieb:
Rats. I'll need some hand-holding, I fear. I don't quite understand what I did wrong to fail the PR test. I know I did a "git pull" just before trying the pull request, trying to ensure that all was the same... expected nothing to change the FT-1D at all.

Will backing out the "git pull" fix this? How do I do that?

PR #1380, opened 2025-09-05
Added two messages to the translations

giovanni69:
I just noticed that two strings are not included in the translation files.
The two strings are present in wxui/main.py:
Open last serial trace
No traces stored


kk7ds:
Ah, yes I just added these recently and must have forgotten to mark them, thanks. This needs the catalog rebuilt. I will do that here before I merge, so don't worry about the PR check fail.

giovanni69:
Thank you!

PR #1391, opened 2025-09-09
Fixes #12109 Related to #11708 Add support for Radtel RT-900, RT-900 BT, RT-910, RT-910 BT and RT-920

TrimbleSoftware:
Fixes #12109 Related to #11708 Add support for Radtel RT-900, RT-900 BT, RT-910, RT-910 BT and RT-920

**NOTE This PR does not provide _support_ for the KDH BT8000 / KSUN UV-110D referred to in issue #11708**

Adds CHIRP- next support for the following new Radtel models using the latest firmware:

**RT-900** 512 Channels
**RT-900 Bluetooth** 999 Channels
**RT-910** 512 Channels in 8 zones of 64 channels
**RT-910 Bluetooth** 960 Channels in 15 zones of 64 channels
**RT-920**  960 Channels in 15 zones of 64 channels.  **VHF and UHF functionality only**. **SW, HF and SSB support is not included.**

These drivers have been _decoupled_ from all the test driver super() class dependencies. It was becoming to difficult to code around the limitations of the parent classes that were contained in two different modules. This module doesn't subclass any other drivers, but it does import some methods for entering and exiting programming mode.

That included module passes all the local tox tests.


kk7ds:
> These drivers have been _decoupled_ from all the test driver super() class dependencies. It was becoming to difficult to code around the limitations of the parent classes that were contained in two different modules. This module doesn't subclass any other drivers, but it does import some methods for entering and exiting programming mode.

How much is copied from the original superclass into this/these to avoid the linkage? Unless it's just simple things like entering programming mode, I'll need some convincing.

TrimbleSoftware:
> > These drivers have been _decoupled_ from all the test driver super() class dependencies. It was becoming to difficult to code around the limitations of the parent classes that were contained in two different modules. This module doesn't subclass any other drivers, but it does import some methods for entering and exiting programming mode.
> 
> How much is copied from the original superclass into this/these to avoid the linkage? Unless it's just simple things like entering programming mode, I'll need some convincing.

The only **non**-heavy modified methods in this code are the enter programming mode and upload/download code. Everything else is a mash-up of the code from the super() class methods and the test driver code, but without all the code to work-around the problems and side effects the super() classes caused. 

The super() classes had a lot of settings that these radios don't use, the mem.extra names and settings are different and the _mem structures didn't have meaningful names.

Here are some of the reasons I made the choice to refactor this code:

1) The super().get_settings() would often fail with exceptions when the image memory was accessed and there were unknown/nonsensical values in there that the super() methods tried to used to decode those settings that are not used by these radios. I had to go as far as filling that section of the _mem structure with 0xff values (of course this region of the image was not written back to the radio) as a _hack_ to get around that error for testing...

2) The earlier "test drivers" had  to _remove_ certain settings groups that were part of the super() classes that did not apply to these radios like ANI Codes and ANI Names. Made for ugly, hard to maintain code.

3) The mem.extra items from the super() classes had to be removed and replaced by ones that these radios use with appropriate names and list options.

4) the _mem structure used in the super() classes has _**almost**_ the correct memory map, but there were considerable "unknown..." or "unused..." names that are in memory locations used by these radios for settings and memory channels. That makes for hard  to maintain code when the mem structures don't have meaningful memory map names. The test driver code was littered with these non-meaningful struct member names.

Would you like to see a _consolidate_ diff of the super() classes, test driver code and this resultant code?

kk7ds:
> Would you like to see a _consolidate_ diff of the super() classes, test driver code and this resultant code?

Do you have it handy or is it easy to generate? If so, I'd take a look. If not, I'll trust your thoughtful analysis. I just want to make sure this is substantially different enough to justify the other driver. Some people will copy a whole driver, make a bunch of relatively meaningless changes and submit it which only creates more maintenance headache. If you say there's a substantial feature delta then that's probably a good sign that a split is warranted.

TrimbleSoftware:
> Do you have it handy or is it easy to generate? If so, I'd take a look.

Here is the diff3 output. It's a bit _hard_ to read, but you'll see a lot of changes between the three files:

MYFILE == \radtel_rt900.py (code in this PR) 
OLDFILE ==  \mml_jc8810.py (base super() class)
YOURFILE == \kdh_uv110D-BT-RT900-RT920.py (last "test driver" code and first lever of super() class)

[diff3_output.txt](https://github.com/user-attachments/files/22245465/diff3_output.txt)

kk7ds:
Yeah, too much going on in the diff to really be useful.

I'll trust you, I just want to avoid copying drivers making small changes and calling it a new one. That generates a lot to maintain. Sounds like it's justified here.

TrimbleSoftware:
> Yeah, too much going on in the diff to really be useful.
> 
> I'll trust you, I just want to avoid copying drivers making small changes and calling it a new one. That generates a lot to maintain. Sounds like it's justified here.

one thing I forgot to mention also is this decoupling removes a few _negative seeks_ because the super().super() base class has a mem map that straddles some locations theses radios need for settings. I'm aware that just a few negative seeks wouldn't justify a whole new independent driver, but it does show one more reason I decided to fork the base classes.

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+    for addr in range(0, radio._memsize, radio.BLOCK_SIZE):
+        status.cur = addr + radio.BLOCK_SIZE
+        radio.status_fn(status)
+
+        block = _read_block(radio, addr, radio.BLOCK_SIZE)
+        data += block
+
+        radio.pipe.log('Sending request for %04x' % addr)
+
+    _exit_programming_mode(radio)
```
kk7ds: I'm trying to push people to use try..finally to make sure we exit the programming mode always. This helps avoid the case where we put the radio in programming mode, then do something and fail, and leave the radio stuck in that mode because we never sent the exit command. It's less important for radios that time out of that mode quickly. Is this one of those?

If not, I think the better pattern is to make your `sync_in()` look like this:
```
try:
    do_download(self)
finally:
    _exit_programming_mode(radio)
```

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+        for addr in range(start_addr, end_addr, radio.BLOCK_SIZE_UP):
+            status.cur = addr + radio.BLOCK_SIZE_UP
+            radio.status_fn(status)
+            radio.pipe.log('Sending address %04x' % addr)
+            _write_block(radio, addr, radio.BLOCK_SIZE_UP)
+
+    _exit_programming_mode(radio)
+
+
+def _split(rf, f1, f2):
```
kk7ds: This has been copied a lot between the drivers, can we unify this somewhere? We already have `chirp_common.in_range()` which is close, but not quite the same.

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+            data = do_download(self)
+        except errors.RadioError:
+            # Pass through any real errors we raise
+            raise
+        except Exception:
+            # If anything unexpected happens, make sure we raise
+            # a RadioError and log the problem
+            LOG.exception('Unexpected error during download')
+            raise errors.RadioError('Unexpected error communicating '
+                                    'with the radio')
```
kk7ds, line 511: ...finally exit here. Same for upload.

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+        # secret radio mode setting (no menu). PTT + 8 for Super Mode
+        rs = RadioSettingValueMap(self._RADIO_MODE_MAP, _radio_mode)
+        rset = RadioSetting(
+            "opmode.radio_mode", "Radio Operating Mode", rs
+        )
+        rset.set_warning(
+            'This should only be used to change the operating MODE of your '
+            'radio if you understand the legalities and implications of '
+            'doing so. The change may enable the radio to transmit on '
+            'frequencies it is not Type Accepted to do and my be in '
```
kk7ds: **may** be?

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+            [str(x) for x in self.POWER_LEVELS],
+            current_index=vfo.a.txpower
+        )
+        rset = RadioSetting("vfo.a.txpower", "TX Power", rs)
+        achannel.append(rset)
+
+        # Menu 05: Wide/Narrow Band
+        rs = RadioSettingValueList(
+            self._bandwidth_list, current_index=vfo.a.widenarr
+        )
```
kk7ds: I assume you're using one of the newer formatters that like to waste a lot of vertical space with things like parens on their own line. At least in the future, if you could avoid that I'd appreciate it. Some vertical space usage is helpful of course, but several lines with nothing but a close paren is just a style I really hate. </rant>

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+            _mem.txtone = int(mem.rtone * 10)
+            _mem.rxtone = 0
+        elif mem.tmode == "TSQL":
+            _mem.txtone = int(mem.ctone * 10)
+            _mem.rxtone = int(mem.ctone * 10)
+        elif mem.tmode == "DTCS":
+            rxmode = txmode = "DTCS"
+            _mem.txtone = DTCS.index(mem.dtcs) + 1
+            _mem.rxtone = DTCS.index(mem.dtcs) + 1
+        elif mem.tmode == "Cross":
```
kk7ds: Can you use `chirp_common.split_tone_encode()` please? That helps to make sure these all behave the same.

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+        msgs = []
+        in_range = chirp_common.in_range
+        AM_mode = 'AM' in mem.mode
+
+        if in_range(mem.freq, self._AIRBANDS) and not AM_mode:
+            msgs.append(chirp_common.ValidationMessage(
+                _('Frequency in this range requires AM mode')))
+        return msgs + super().validate_memory(mem)
+
+    def set_settings(self, settings):
```
kk7ds, line 1363: I'm also trying to push for using `MemSetting` as this massive and obscure `set_settings()` routine just keeps getting copied from driver to driver. Here's an example:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/tk280.py#L1626

Basically, name the settings like they are in the memory object and let `apply_to()` mostly just set everything for you, similar to what this routine is doing, but in a standard way. There will be some things that will be too obscure and bespoke to work that way, but apply callbacks can make those specific cases easier.

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1678 @@
...
+    @classmethod
+    def get_prompts(cls):
+        rp = chirp_common.RadioPrompts()
+        rp.experimental = \
+            ('This driver is a beta version ONLY for the RT-900 BT'
+             ' running Firmware V.1.20P.\n'
+             '\n'
+             'Please save an unedited copy of your first successful\n'
+             'download to a CHIRP Radio Images(*.img) file.\n\n'
+             'PROCEED AT YOUR OWN RISK!'
```
kk7ds, line 1051: We could probably avoid duplicating this warning in each of the subclasses.. Also note that the internal backup system does this for you always anyway. I'm not sure if I dislike this kind of warning per-radio or not... it's a good idea with any radio (which is why it's automatic now).

TrimbleSoftware: I only added it per radio to display the firmware version that the driver is know and tested to work on. As you know these "_cheap_ radio" manufactures are always changing the firmware in unpredictable ways. Some of which make a driver no longer work on that radio. Too bad the radio's firmware has no way to get the firmware version from the memory blob. On all of these the _fingerprint_ doesn't ever change between firmware versions, much less different models. About the only thing they change is the _magic_ to enter programming mode and only sometimes...

chirp/drivers/radtel_rt900.py
TrimbleSoftware: That's sound advise. I've always been a fan of try, catch, finally in c#. I'll add it.

chirp/drivers/radtel_rt900.py
TrimbleSoftware: Yes!

chirp/drivers/radtel_rt900.py
TrimbleSoftware: The hanging parens are to get around the tox style over/under indented errors that are kind of a problem with the 80 char line length limit with all the line folding that has to be done. I use VS Code for editing and it's default indenting is mostly not liked by tox style check...

OF all the style check, I'd like to see that one relaxed someday. _There's not too many 80 col printers left around,_ LOL!

chirp/drivers/radtel_rt900.py
TrimbleSoftware: Thanks, I'll look at that as an example.

chirp/drivers/radtel_rt900.py
TrimbleSoftware: OK

chirp/drivers/radtel_rt900.py
TrimbleSoftware: OK, , I'll look at it.

TrimbleSoftware: You mean add a new method chirp_common.split() to place this functionality into a reusable container? I'd assume you'd want a separate PR for that change?

kk7ds: Doesn't have to be a separate PR, but definitely a separate commit. This is already copied elsewhere, so I won't hold up this PR for the refactor, but I'd sure appreciate it if you could circle back and remove all the duplicates (this one included) in favor of a common implementation.

TrimbleSoftware: Created new PR #1397 to add is_split() method to chirp_common for uses with this new driver

TrimbleSoftware:
Latest commits won't pass PR Driver tests until changes to add is_split() to chirp_common.py in PR #1397 are migrated.

`>               if chirp_common.is_split(self.get_features().valid_bands,
                                         mem.freq, int(_mem.txfreq) * 10):
E                                        AttributeError: module 'chirp.chirp_common' has no attribute 'is_split'`

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1600 @@
...
+
+    _has_bt_denoise = True
+    _has_am_per_channel = True
+    _has_am_switch = not _has_am_per_channel
+    _has_single_mode = True
+    _valid_chars = chirp_common.CHARSET_ALPHANUMERIC + \
+        "`~!@#$%^&*()-=_+[]\\{}|;':\",./<>?"
+
+    _steps = [2.5, 5.0, 6.25, 10.0, 12.5, 20.0, 25.0, 50.0, 8.33]
+    # _step_list = ['%#.3g K' % x for x in _steps]
```
kk7ds: This should be removed.

chirp/drivers/radtel_rt900.py
kk7ds: Still needs fixing?

chirp/drivers/radtel_rt900.py
kk7ds: Thanks for refactoring to use that. So much better than copying the set-settings-of-doom method into every driver, IMHO.

chirp/drivers/radtel_rt900.py
```diff
@@ -0,0 +1,1600 @@
...
+    except Exception:
+        raise errors.RadioError("Error communicating with radio")
+    if ident not in radio._fingerprint2:
+        LOG.debug(util.hexprint(ident))
+        raise errors.RadioError("Radio returned unknown secondary"
+                                " identification string")
+
+
+def _exit_programming_mode(radio):
+    mml_jc8810._exit_programming_mode(radio)
```
kk7ds, line 317: ...probably don't need this and can just call straight to the other implementation right?

kk7ds (commented):
I'd also like to squash the fixup commits for the final merge.. I'll do that and fix the typo so we can get this done. Thanks!

PR #1392, opened 2025-09-09
Fixes Bug #12108 and several undefined subroutines.

Darieb:
Fixes Bug #12108. CHIRP now handles WiresX more correctly: all 20 "rooms" for all five "categories" and minimizes displaying too many settings at once by using hierarchical settings menus. 
Also removes reference to an obsolete routine name which was undefined and used in Backtrack settings.

Passes all CHIRP tests for style, Yaesu radios FT-1D_R, FT2D_R, FT3D_R, FTM-3200D_R, and FTM-7250D.

### DIFF
--- a/chirp/drivers/ft1d.py
+++ b/chirp/drivers/ft1d.py
@@ -1,6 +1,6 @@
 # Copyright 2010 Dan Smith <dsmith@danplanet.com>
 # Copyright 2014 Angus Ainslie <angus@akkea.ca>
-# Copyright 2023 Declan Rieb <WD5EQY@arrl.net>
+# Copyright 2023,2025 Declan Rieb <WD5EQY@arrl.net>
 # Sections of digital settings applied from ft70.py, thus
 # Copyright 2017 Nicolas Pike <nick@zbm2.com>
 #
@@ -709,7 +709,8 @@ class FT1Radio(yaesu_clone.YaesuCloneModeRadio):
     VENDOR = "Yaesu"
     MODEL = "FT-1D"
     VARIANT = "R"
-    FORMATS = [directory.register_format('FT1D ADMS-6', '*.ft1d')]
+    FORMATS = [directory.register_format('FT1D ADMS-6', '*.ft1d'),
+               directory.register_format('Yaesu SDcard', '*.dat')]
     class_specials = SPECIALS
     _model = b"AH44M"
     _memsize = 130507
@@ -1045,10 +1046,7 @@ class FT1Radio(yaesu_clone.YaesuCloneModeRadio):
             mem.offset = int(_mem.offset) * 1000
             mem.rtone = mem.ctone = chirp_common.TONES[_mem.tone]
             self._get_tmode(mem, _mem)
-            if mem.duplex is None:
-                mem.duplex = DUPLEX[""]
-            else:
-                mem.duplex = DUPLEX[_mem.duplex]
+            mem.duplex = DUPLEX[_mem.duplex] if _mem.duplex else ""
             if mem.duplex == "split":
                 mem.offset = chirp_common.fix_rounded_step(mem.offset)
             mem.mode = self._decode_mode(_mem)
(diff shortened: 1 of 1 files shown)


Darieb:
Is there a way to get your automatic tests to include testing of dependent drivers? I'd been guilty of only doing FT-1D_R tests, and just recently remembered to test FT2D, FT3D, FTM-3200D and FTM-7250D as well; they all depend on FT-1D_R.

chirp/drivers/ft1d.py
```diff
@@ -709,7 +709,8 @@ class FT1Radio(yaesu_clone.YaesuCloneModeRadio):
     VENDOR = "Yaesu"
     MODEL = "FT-1D"
     VARIANT = "R"
-    FORMATS = [directory.register_format('FT1D ADMS-6', '*.ft1d')]
+    FORMATS = [directory.register_format('FT1D ADMS-6', '*.ft1d'),
+               directory.register_format('Yaesu SDcard', '*.dat')]
```
kk7ds, line 713: Nothing else required to open these, they're exactly like a `.ft1d` right?

chirp/drivers/ft1d.py
```diff
@@ -1045,10 +1046,7 @@ def get_memory(self, number):
             mem.offset = int(_mem.offset) * 1000
             mem.rtone = mem.ctone = chirp_common.TONES[_mem.tone]
             self._get_tmode(mem, _mem)
-            if mem.duplex is None:
-                mem.duplex = DUPLEX[""]
-            else:
-                mem.duplex = DUPLEX[_mem.duplex]
+            mem.duplex = DUPLEX[_mem.duplex] if _mem.duplex else ""
```
kk7ds, line 1049: The old code looks definitely wrong, but ... I'm not sure `_mem.duplex` would ever equate to `False` since it's a bitwise object. Does this actually fix a problem or are you just "gardening" ?

chirp/drivers/ft1d.py
```diff
@@ -2518,37 +2487,37 @@ def apply_WEhemi(cls, setting, obj):
 
     def apply_bt_lat(cls, setting, obj):
         val = setting.value.get_value()
-        val = cls.backtrack_zero_pad(val, 3)
+        val = cls.zero_pad(val, 3)
```
kk7ds, line 2490: Are these related to the stated bug or again just gardening? Seems like `zero_pad()` shouldn't be a method at all, but just a helper function at the top of this file, right? Do any classes actually override this?

chirp/drivers/ft1d.py
```diff
@@ -1907,66 +1905,37 @@ def _get_digital_settings(self):
...
+                cname = f"WiresX_settings.RoomsPerCategory[{i}]."\
+                    f"Rooms[{j}].name"
+                dname = f"Cat.[{i + 1}] ({c.strip()}) Room{j + 1: 02d}"
+                rs = RadioSetting(cname, dname, val)
+                WXCmenu.append(rs)
+                idn = str(r.Rooms[j].ID).strip('\xff').ljust(5)
+                val = RadioSettingValueString(0, 5, idn)
+                vname = f"WiresX_settings.RoomsPerCategory[{i}]."\
+                    f"Rooms[{j}].ID"
+                rs = RadioSetting(vname, f"   YSF ID{j + 1} (5 digits)",
```
kk7ds, line 1933: Please don't do this sort of hacking of the GUI from the driver. They might look right on your screen, but they likely don't on everyone's and could be entirely not appropriate with a different UI framework. Are you trying to set them off visually? If so, please use `RadioSettingSubGroup()` or something.

kk7ds (changes requested):
I think if you look at the check output you'll see that it _is_ running the dependent radios as well. But, I hope you're running the full set locally before you submit anyway.

Please separate the gardening fixes from the actual functional changes/fixes into separate commits. Also, please give those commits useful subject lines per the PR guidelines. Right now, the email notification to the list would just say the build "Fixes bug XXX" and unless a user knows what that bug is, would not see that they might be interested in this. Please follow the usual convention of something like:
```
ft1d: Improve WiresX room settings layout

Fixes: #12108
```

chirp/drivers/ft1d.py
```diff
@@ -1907,66 +1905,37 @@ def _get_digital_settings(self):
...
-                                  (j+1), val)
-                rs.set_apply_callback(self.apply_WiresX_roomname,
-                                      r.Rooms[j])
-                WXmenu.append(rs)
+            for j in range(20):
+                cn = str(r.Rooms[j].name).strip('\xff').ljust(16)
+                val = RadioSettingValueString(0, 16, cn)
+                cname = f"WiresX_settings.RoomsPerCategory[{i}]."\
+                    f"Rooms[{j}].name"
+                dname = f"Cat.[{i + 1}] ({c.strip()}) Room{j + 1: 02d}"
```
kk7ds, line 1926: This is a highly unusual name for a setting:

`Cat.[1] () Room 2`

Can you make these more human friendly, like `Category 1 Rooms` and `Category 1 Room 2` please?

Also, please don't make the name dependent on other settings. It just means that the display isn't consistent across refreshes/reloads. I think it's fine to just mention them by category and room number.

PR #1393, opened 2025-09-10
Add Retevis RA86 mobile radio - fixes #10674

KC9HI:
# CHIRP PR Guidelines


chirp/drivers/retevis_ra86.py
```diff
@@ -0,0 +1,609 @@
...
+struct memory chnl18;
+struct memory rptr18;
+struct memory chnl19;
+struct memory rptr19;
+struct memory chnl20;
+struct memory rptr20;
+struct memory chnl21;
+struct memory rptr21;
+struct memory chnl22;
+struct memory rptr22;
```
kk7ds, line 58: Just so you know, you can put these in a struct and array to avoid having to make them all by hand:
```
struct {
  struct memory channel;
  struct memory repeater;
} midchannels[7];
```
and access them like `mem.midchannels[i].channel`

KC9HI: I didn't like that so I tried to get Chat GPT and others to show me how to condense that. Nothing that they showed me was that simple so I finally gave up. I am in no hurry with this, so I could update and resubmit tomorrow if you want.

kk7ds: Might as well just merge it as-is to get it out for tomorrow and you can circle back with a revision. Not critical to fix but probably worth it in the long run to be cleaner.

PR #1394, opened 2025-09-10
FT1D:improve WiresX settings./nFixes: #12108

Darieb:
Yaesu FT-1D, FT2D, FT3D fixes WiresX settings to handle all rooms in category.

Fixes #12108.

### DIFF
--- a/chirp/drivers/ft1d.py
+++ b/chirp/drivers/ft1d.py
@@ -1,6 +1,6 @@
 # Copyright 2010 Dan Smith <dsmith@danplanet.com>
 # Copyright 2014 Angus Ainslie <angus@akkea.ca>
-# Copyright 2023 Declan Rieb <WD5EQY@arrl.net>
+# Copyright 2023,2025 Declan Rieb <WD5EQY@arrl.net>
 # Sections of digital settings applied from ft70.py, thus
 # Copyright 2017 Nicolas Pike <nick@zbm2.com>
 #
@@ -27,7 +27,7 @@ from chirp import memmap
 from chirp.settings import RadioSettingGroup, RadioSetting, RadioSettings, \
             RadioSettingValueInteger, RadioSettingValueString, \
             RadioSettingValueList, RadioSettingValueBoolean, \
-            InvalidValueError
+            InvalidValueError, RadioSettingSubGroup
 from chirp import util
 
 LOG = logging.getLogger(__name__)
@@ -1045,10 +1045,7 @@ class FT1Radio(yaesu_clone.YaesuCloneModeRadio):
             mem.offset = int(_mem.offset) * 1000
             mem.rtone = mem.ctone = chirp_common.TONES[_mem.tone]
             self._get_tmode(mem, _mem)
-            if mem.duplex is None:
-                mem.duplex = DUPLEX[""]
-            else:
-                mem.duplex = DUPLEX[_mem.duplex]
+            mem.duplex = DUPLEX[_mem.duplex] if _mem.duplex else ""
             if mem.duplex == "split":
                 mem.offset = chirp_common.fix_rounded_step(mem.offset)
             mem.mode = self._decode_mode(_mem)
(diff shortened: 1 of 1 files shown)


Darieb:
Resubmitted, removing most gardening. Thanks for the thoughtful comments.
- The two DUPLEX logic errors should have affected WiresX activity, IMO. I left it in.
- The SDcard format is definitely NOT the same as .ft1d, but should not have been included here. That requires another fork, and would help avoid problems with USB connections, since Yaesu counts on the SDcard for memory backups.
- The helper routine you noticed is defined above its usage, but not with the name called in Backtrack. That's definitely gardening, so expect another fix. Of course, nobody has complained about it, so it's not high priority. 
- I don't quite understand the specifics of "hacking of the GUI", but implemented RadioSettingSubGroup anyway. I was trying to make settings pages that wouldn't be excessively long or empty space (5 categories x 20 rooms/category means at least 100 full lines of output.) It works just OK now.

kk7ds:
> Resubmitted, removing most gardening. Thanks for the thoughtful comments.

The gardening is okay, it should just be in separate commits from the functional stuff. Also, please re-use the PR for a given topic until it's actually merged, don't close and open a new one. That way the comments and conversation all stay together.

> * The two DUPLEX logic errors should have affected WiresX activity, IMO. I left it in.

Okay, I was missing how that could happen...

> * The SDcard format is definitely NOT the same as .ft1d, but should not have been included here. That requires another fork, and would help avoid problems with USB connections, since Yaesu counts on the SDcard for memory backups.

Okay glad I asked :)

> * The helper routine you noticed is defined above its usage, but not with the name called in Backtrack. That's definitely gardening, so expect another fix. Of course, nobody has complained about it, so it's not high priority.

Sounds good.

> * I don't quite understand the specifics of "hacking of the GUI", but implemented RadioSettingSubGroup anyway. I was trying to make settings pages that wouldn't be excessively long or empty space (5 categories x 20 rooms/category means at least 100 full lines of output.) It works just OK now.

The space-prefixing of names so they "appear" nested underneath the previous item is what I meant by "hacking the UI". In the UI, things are aligned by policy so they work on multiple devices and screens properly. Things like adding space to the front of the names just isn't a reliable and repeatable way to do that and thus shouldn't.

It's late here so I'll circle back tomorrow to have another look, thanks!

Darieb:
Thanks for the replies. I understand better what you meant by hacking the UI. I've fixed that now. And I rather like that the display highlights the different categories' boundaries. So another draft will be pushed RSN.

But when displaying the settings -> Digital -> Wires-X window, mousing over the settings window produces multiples of one error, which doesn't yet affect anything as far as i can tell. But there may need to be a WXUI fix to come, at least to check if "setting" is an object:

<computer>Traceback (most recent call last):
  File "/Users/declan/chirp/chirp/wxui/common.py", line 407, in _mouseover
    if tip and tip == setting.get_name():
                      ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_name' 
</computer>

**From wxui/common.py, line 369-413**
class ChirpSettingGrid(wx.Panel):
    def __init__(self, settinggroup, *a, **k):
<snip>
    def _mouseover(self, event):
        prop = self.pg.HitTest(event.GetPosition()).GetProperty()
        tip = None
        if prop:
            setting = self.get_setting_by_name(prop.GetName())
            tip = setting.__doc__ or None
            if tip and tip == setting.get_name():
                # This is the default, which makes no sense, but it's been
                # that way for ages, so just avoid exposing it here if it's
                # set to the default.
                tip = None

        event.GetEventObject().SetToolTip(tip)

kk7ds:
Probably means you're adding None to a SettingGroup somewhere, but as I always say, without seeing the code I'm just guessing. Push it up here if you can't figure it out, else I can't really help.

Darieb:
Mousing over the separator line that's associated with the RadioSettingsSubGroup gives the error. That's line 1914 of the new push, whose commit message I bollixed yet again.
*sigh*

chirp/drivers/ft1d.py
```diff
@@ -1907,66 +1904,37 @@ def _get_digital_settings(self):
...
             val = RadioSettingValueString(0, 16, c)
-            rs = RadioSetting(cname, "Category %d" % (i+1), val)
-            rs.set_apply_callback(self.apply_WiresX_category,
-                                  wxc.Category[i].name)
+            rs = RadioSetting(cname, f"Category {i+1: 2d} Name", val)
             WXmenu.append(rs)
 
+            WXCmenu = RadioSettingSubGroup(
+                        f"WiresX_settings.Category[{i}].RoomsPerCategory",
+                        f"*** Category{i + 1} Rooms ***")
```
kk7ds, line 1916: Please remove the asterisks. This, again, is trying to influence the display (emphasis in this case) which is not something the driver should be doing. The subgroup header is already shaded to set it off from the others.

chirp/drivers/ft1d.py
```diff
@@ -1045,10 +1045,7 @@ def get_memory(self, number):
             mem.offset = int(_mem.offset) * 1000
             mem.rtone = mem.ctone = chirp_common.TONES[_mem.tone]
             self._get_tmode(mem, _mem)
-            if mem.duplex is None:
-                mem.duplex = DUPLEX[""]
-            else:
-                mem.duplex = DUPLEX[_mem.duplex]
+            mem.duplex = DUPLEX[_mem.duplex] if _mem.duplex else ""
```
kk7ds, line 1048: I still don't understand how this is related to the WiresX stuff. If it is, it's totally fine, I just want to understand.

Darieb:
Got rid of the asterisks. 
The duplex logic was wrong, but it affects my Wirex-X access because my hotspot is simplex only. WIthout the correct duplex, I cannot test WIresX online. 
UI still complains when I mouseover the SubGroup; I could find no "None"s that I've sent.

Is there a way to run the first "PR" test locally? I clearly don't know what I'm messing up with the commit message(s). At least PR is pass-fail. Presently, I run:
tox -e style
tox -e driver -- -k Yaesu_FT-1D_R
tox -e driver -- -k Yaesu_FT2D_R
tox -e driver -- -k Yaesu_FT3D_R
tox -e driver -- -k Yaesu_FTM-3200D_R
tox -e driver -- -k Yaesu_FTM-7250

Yeah, git has way too many options for me to play with in bits and pieces, and this is my only application thereof. I built the ft1d.py git diff, and git apply-ed it to a refreshed copy of your ft1d.py. To do that, I started a new branch "fixdigi". Then I removed the asterisks. I've got commit-ed the changes to my fork of your chirp, but not pushed anything (yet.)

kk7ds:
> The duplex logic was wrong, but it affects my Wirex-X access because my hotspot is simplex only. WIthout the correct duplex, I cannot test WIresX online.

I'm asking you to explain what the problem was because I don't see how mem.duplex could ever be None, so while your code looks right, I don't understand why it makes any difference. Could you _please_ just explain?

> UI still complains when I mouseover the SubGroup; I could find no "None"s that I've sent.

I didn't see this behavior so I thought you had fixed it. If you're still seeing it, perhaps it's a windows thing and I can try there to see.

> Is there a way to run the first "PR" test locally? I clearly don't know what I'm messing up with the commit message(s).

The check shows you exactly what the problem is: https://github.com/kk7ds/chirp/pull/1394/checks .. it's just running `tools/check_commit.sh` but the invocation depends on your git setup.

> Yeah, git has way too many options for me to play with in bits and pieces, and this is my only application thereof. I built the ft1d.py git diff, and git apply-ed it to a refreshed copy of your ft1d.py. To do that, I started a new branch "fixdigi". Then I removed the asterisks. I've got commit-ed the changes to my fork of your chirp, but not pushed anything (yet.)

Yeah, please stop doing that because you've now deleted the branch and closed this PR which means this conversation will be spread over (at least) three PRs, which makes it very difficult. Look up `git commit --amend`, `git rebase -i`, and `git push -f`.

PR #1397, opened 2025-09-17
Add helper method is_split() to chirp_common

TrimbleSoftware:
Add helper method is_split() to chirp_common

A local _split() method is defined in at least 12 driver modules and is **not** easily _reusable_ in other driver modules.

This change will make way for an easily _reusage_ equivalent of all those local _split() method implementations.

This change does not address any outstanding issue, but came about during a PR code review of a new driver (PR #1391) that uses a local _split() method yet again. This is an opportunity to finally make it reusable.

The future intent is the refactor the existing driver modules that have local defs of their own _split() method and replace those with a call to the one contained in this change.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -2137,3 +2137,19 @@ def in_range(freq, ranges):
         if lo <= freq <= hi:
             return True
     return False
+
+
+def is_split(rf, freq1, freq2):
+    """Check if two freqs are in the same radio band
+    Returns False if the two freqs are in the same band (not split)
+    or True if they are in separate bands (split)"""
+
+    # determine if the two freqs are in the same band
+    for low, high in rf.valid_bands:
+        if freq1 >= low and freq1 <= high and \
+                freq2 >= low and freq2 <= high:
+            # if the two freqs are on the same Band this is not a split
+            return False
+
+    # if you get here is because the freq pairs are split
+    return True
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -2139,13 +2139,13 @@ def in_range(freq, ranges):
     return False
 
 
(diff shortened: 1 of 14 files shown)


chirp/chirp_common.py
```diff
@@ -2137,3 +2137,19 @@ def in_range(freq, ranges):
         if lo <= freq <= hi:
             return True
     return False
+
+
+def is_split(rf, freq1, freq2):
```
kk7ds: I think either:
- We should make this take a list of bands instead of a `RadioFeatures` so it can be used elsewhere (like in the network sources). Existing uses will just pass `rf.valid_bands` here instead of `rf`
- We should just make this a method on `RadioFeatures` itself and call it `same_bands(self, freq1, freq2)`
What do you think? I'm tempted to go with the former.

Let's also refactor the other uses of this while adding it (I can do that, once we decide on the above).

Thanks!

TrimbleSoftware: > I think either:
> * We should make this take a list of bands instead of a `RadioFeatures` so it can be used elsewhere (like in the network sources). Existing uses will just pass `rf.valid_bands` here instead of `rf`

I like this approach so I committed an update that implements that. This is how it's called in _working_ code:

`if chirp_common.is_split(self.get_features().valid_bands, mem.freq, int(
                          _mem.txfreq) * 10):`

TrimbleSoftware:
Committed changes to 13 driver source files to use the new chirp_common.is_split() method to replace all local _split() method defs. I didn't manually test the results for all of the supported radios as I don't have all of these radios. I just have the Baofeng UV5R-Mini and it works fine. All the tox test pass too. Shouldn't need to update all the related test .img files.

kk7ds:
I was going to do the cleanup of the other drivers, but super awesome, thanks very much!

PR #1494, opened 2026-02-27
kgq10h: Add Wouxun KG-Q10H driver (Phase 1)

csbetz:
## Summary

- New clone-mode driver for the Wouxun KG-Q10H quad-band HT (136-174 / 222-225 / 350-390 / 400-480 MHz)
- Clone protocol at 115200 baud with XOR-chained encryption and adjusted checksum
- Memory rearrangement: 256-byte blocks reversed within 1024-byte regions, expressed as a simple formula (replaces the 128-entry lookup table from PR #796)
- 999 channels, 12-character names, 4 power levels (L/M/H/U), FM/NFM/AM modes
- Full CTCSS/DCS tone support with cross-mode
- Airband validation (108-136 MHz requires AM)
- Test image from a real KG-Q10H with 68 programmed channels

This is Phase 1 covering clone protocol and basic memory. Follow-up PRs will add extra channel properties (scrambler, compander, etc.) and radio settings.

Related to #796

## Test plan

- [x] `tox -e style` passes
- [x] `tox -e unit` passes (412 passed)
- [x] `tox -e driver -k KGQ10H` passes (32 passed, 9 skipped)
- [x] Round-trip download tested with physical KG-Q10H radio
- [x] Round-trip upload tested with physical KG-Q10H radio — channels verified intact


csbetz:
## Shared code with kg935g.py

Seeing your comments on other PRs, I wanted to flag that this is distinct from the prior Wouxun family, but bears some similarity to the standalong KG-935G.  The KG-Q10H and KG-935G share the same Wouxun encrypted serial protocol family. There are **9 functions/blocks totaling ~142 lines** that look functionally identical or very similar with some small changes:

| Function | Lines | Difference |
|---|---|---|
| `strxor()` | 2 | Identical |
| `encrypt()` | 5 | `_cryptbyte` (0x54) vs hard-coded 0x57 |
| `decrypt()` | 6 | Same as above |
| `_checksum()` | 6 | Identical logic (sum mod 256) |
| `_get_tone()` | 55 | Same algorithm|
| `_set_tone()` | 33 | Same algorithm|
| duplex/offset decode | 12 | Identical logic |
| name decode loop | 4 | Identical |
| `_do_download()` | 19 | Identical loop; Q10H adds `time.sleep(0.005)` for some stability |

I'm happy to extract these into a shared base class (e.g. `WouxunNewBase` in a `wouxun_kg_common.py` module) as a follow-up PR if that's preferred. The existing `wouxun_common.py` covers the older unencrypted protocol (KG-UV6 etc.) so this would be a separate shared module for the newer encrypted protocol family. Alternately, I could extract just the crypto and tone helpers and leave the rest. Or I can leave it alone.

Overal - not a ton of code, but let me know what you'd like me to do.

kk7ds:
Thanks for that analysis, and yes, please do consolidate :) I'm a bit under the weather at the moment so it will be a few days before I can even look at this. We should also get @mrtjr1 to have a look since he looks after a bunch of our wouxun drivers.

mrtjr1:
There is likely some consolidation that could be done if done strategically.   Wouxun is notorious for making minor changes to simple things like cryptbyte, name decoding ( see the 950 vs 980p test driver) and even their checksums ( Q10h series added what appears to be an offset based on a byte value in the packet) see the test driver.

mrtjr1:
Not to mention the Q10H series unique data storage structure that requires special address handling to convert the radio data into contiguous addresses to allow Chirp to access it properly.

mrtjr1 (commented):
I question the need for the big upload table map.  In the test driver, I was eliminating those big tables and just using a chirp address to radio address conversion function to simplify the driver.

I don't think I posted the latest test driver I am working on  with the convert function added to the Q10H yet.  I makes things much easier.
See the example  for the newer S65 which uses the same memory layout as a reference.

And I usually limit uploads to only mapped areas of memory to prevent the possibility of corrupting values that are unknown. 

https://chirpmyradio.com/attachments/12745

S65G_upload_map = (
    (0x400, 32, 480),
    )

def convert_address(addr_in):

    quotient,remainder = divmod(addr_in,0x0400)
    index = (remainder // 0x0100)
    addr_out = ((quotient+1)*0x0400) - ((index+1)*0x0100) + (remainder % 0x0100)
    return addr_out

    def _do_upload(self):
        # self._identify()
        # if self._RADIO_ID == self._model:
        LOG.debug("Starting Upload")
        cfgmap = S65G_upload_map

        for start, blocksize, count in cfgmap:
            end = start + (blocksize * count)
            LOG.debug("start = " + str(start))
            LOG.debug("end = " + str(end))
            LOG.debug("blksize = " + str(blocksize))

            for i in range(start, end, blocksize):
                # time.sleep(0.005)
                addr_out = convert_address(i)
                ptr = i
                ptr2 = addr_out

kk7ds:
> I question the need for the big upload table map. In the test driver, I was eliminating those big tables and just using a chirp address to radio address conversion function to simplify the driver.

Yeah, you know my feeling on that - I think we had this conversation before. The upload map approach is definitely less desirable than a function that can translate the addresses.

> I don't think I posted the latest test driver I am working on with the convert function added to the Q10H yet. I makes things much easier.

I'm a bit confused, it looks like this PR is just taking your previous/unfinished driver and iterating on it a bit? If you're still working on yours (as I thought was the case) then it would be unfortunate to regress that work. Maybe you can update your PR and the two of you can collaborate on a single implementation?

csbetz:
> > I question the need for the big upload table map. In the test driver, I was eliminating those big tables and just using a chirp address to radio address conversion function to simplify the driver.
> 
> Yeah, you know my feeling on that - I think we had this conversation before. The upload map approach is definitely less desirable than a function that can translate the addresses.
I agree with both of you bad construct... updating.  I used a formula on download and should have on upload.
> 
> > I don't think I posted the latest test driver I am working on with the convert function added to the Q10H yet. I makes things much easier.
> 
> I'm a bit confused, it looks like this PR is just taking your previous/unfinished driver and iterating on it a bit? If you're still working on yours (as I thought was the case) then it would be unfortunate to regress that work. Maybe you can update your PR and the two of you can collaborate on a single implementation?

I just wanted to be able to program my radio; I don't feel strongly about which driver.  I think I've made some more progress and will update later this weekend.  I've been sick too so slowly coming back.  

@kk7ds this will make the driver on the longer side rather than incremental as you've asked before, but more complete.  

@mrtjr1 I'm happy to work with you on a combined driver or stay out of the way, as my immediate goal of being able to program my radio is met.  I'd love to help others with the same, but would celebrate any solution.  I hadn't realized you were still actively developing, I should have played closer attention.

kk7ds:
I don't think the function based approach makes it longer or less iterative. When I say that, I usually mean "just implement memories first, then settings". The function approach is better IMHO because it makes it easier to understand what's going on than trying to look at a large table and reason about the pattern. I understand why the table might seem (or even be) simpler, but in this case I'd prefer the other.

And @csbetz I surely appreciate you trying to pick up something that seemed stalled/abandoned and push it through, no need to feel bad about that. If you had asked first we might have been able to save you some work, but definitely appreciate your help.

I'm not sure what parts @mrtjr1 has been stewing on over this time, but in the interest of making some progress would it help if we split the driver into minimal form (just clone and base memories) to get that merged and then follow up with the settings, memory extras (as appropriate) etc?

mrtjr1:
Let me take a look at putting a simple channel only driver together with the address function for q10h/G and then we can go from there.   I just have been real busy with elderly family care and work.

csbetz:
Got it -- and understand the family and work complexity.  Standing by to help where it is useful.

mrtjr1:
@csbetz you can program your radio complete with all settings by using the test driver located here https://chirpmyradio.com/issues/10692


I haven't updated the original PR because there were so many issues I had to resolve I just left it with Dan's comments so I knew what needed fixing.    
Recent discussions have reminded me that it is best to get the radio supported with cloning and channels first, then subsequently add the settings.   

I will try and pare down the driver to just do that to get it supported officially without the test driver. And then work to add more features.

mrtjr1:
I am pretty close to having a channel only driver ready for the Q10H and Q10G.   
Doing some final radio testing and then the pep and driver tests.

No more big lookup tables. 😀

csbetz:
Exciting.  Great to hear.

PR #1524, opened 2026-03-29
ar8200: Add new driver for AOR AR8200

nothingface:
This is a live mode driver as the communication protocol does not support cloning.  Tested on AOR AR8200 Mark II, international version.  Support is nearly complete with respect to functionality provided over the serial port.  Limitations due to the protocol, radio, and/or CHIRP are described in comments in the driver.

Fixes #12433

### DIFF
--- a/chirp/drivers/ar8200.py
+++ b/chirp/drivers/ar8200.py
@@ -1,0 +1,1692 @@
+# Copyright 2026 Darius Rad <alpha@area49.net>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+import logging
+import re
+import threading
+import time
+
+from chirp import chirp_common, errors, directory
+from chirp.settings import RadioSetting, RadioSettingGroup, \
+    RadioSettingSubGroup, RadioSettingValueBoolean, RadioSettingValueFloat, \
+    RadioSettingValueInteger, RadioSettingValueList, RadioSettingValueMap, \
+    RadioSettingValueString, RadioSettings
+
+LOG = logging.getLogger(__name__)
+
+# CHANNEL NUMBERING
+#
+# CHIRP identifies memories with sequentially numbered indices.  The
+# memory locations in the AR8200 are mapped to CHIRP memories as
+# follows.
+#
+# Memories 0 through 999
+#
+#   Scan memories by bank: A, a, B, b, ... J, j.  Each pair of banks
+#   (A and a, B and b, etc.) occupies 100 channels.  The split between
+#   the banks in the pair is determined by the bank size configured
+#   under settings.
+#
+# Memories 1000 through 1999
+#
+#   Scan pass frequencies by bank: A, a, B, b, ... J, j.  There are 50
+#   pass frequencies per bank.
+#
+# Memories 2000 through 2099
+#
+#   VFO pass frequencies.
+#
+# Memories 2100 through 2139
+#
+#   One channel per search bank: A through T, then a through t.
+#
+# Special channels
+#
+#   VFO: Single VFO mode channel configuration
+#   VFO-A and VFO-B: Twin VFO mode channel configuration
+#   QM0 through QM9: VFO quick memories
+
+# ISSUES AND LIMITATIONS
+#
+# After adjusting any of the following settings, it is necessary to
+# refresh certain other memory locations:
+#
+#  - Bank split size
+#  - Add or remove search pass frequencies
+#  - Change offset frequencies
+#  - Removing channels from select scan list
+#
+# The radio supports three AM modes, but CHIRP only knows about two.
+# Mode is set to AM for all, and an extra field specifies which AM
+# mode is used.
+#
+# Search pass frequencies cannot be modified, they can only be added
+# or removed.  When a pass frequency is removed, all others in the
+# same search bank are moved up.
+#
+# This driver does not provide a bank model because (1) CHIRP does not
+# support banks for live radios, and (2) the banks on the radio don't
+# easily map to the bank model provided by CHIRP.
+#
+# Quick memory channel details beyond the frequency are not available,
+# as reading this from the serial port is not supported.
+#
+# Adding to the select scan list is not supported over the serial
+# port.
+#
+# Auto mode also lets the radio control duplex and offset, tuning
+# step, and step adjust.
+#
+# Radio must be power cycled after using CHIRP.  The serial protocol
+# supports a command to terminate the session and release the keypad,
+# but there is no suitable place in CHIRP to make this call.  By the
+# time the __del__ method is called, the serial port is already
+# closed.
+#
+# Radio supports additional characters not presently reflected in the
+# valid character set, including arrows, Japanese characters (yen,
+# katakana), Greek letters, other European letters and accents, and
+# math symbols.
+
+
+@directory.register
+class AR8200Radio(chirp_common.LiveRadio):
+    """AOR AR8200"""
+    BAUD_RATE = 19200
+    VENDOR = "AOR"
+    MODEL = "AR8200"
+
+    # These need to match the indices of the MD command.
+    _MODES = [('WFM', 0), ('FM', 1), ('AM', 2), ('USB', 3), ('LSB', 4),
+              ('CW', 5), ('NFM', 6), ('WAM', 7), ('NAM', 8), ('Auto', 0xf)]
+
+    _AM_MODES = ['AM', 'WAM', 'NAM']
+
+    _SCAN_MEMORIES = 1000
+    _SEARCH_PASS_MEMORIES = 1100
+    _SCAN_BANKS = 'AaBbCcDdEeFfGgHhIiJj'
+    _SEARCH_BANKS = 'ABCDEFGHIJKLMNOPQRSTabcdefghijklmnopqrst'
+
+    _SPECIAL_CHANS = {
+        'VFO': -13,
+        'VFO-A': -12,
+        'VFO-B': -11,
+        'QM0': -10,
+        'QM1': -9,
+        'QM2': -8,
+        'QM3': -7,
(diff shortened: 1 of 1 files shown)


kk7ds:
Thanks for this, I'll try to review in the coming days. Obviously fine to ignore the PR check complaint about a missing image with a live driver :)

chirp/drivers/ar8200.py
```diff
@@ -0,0 +1,1640 @@
...
+#  - Change offset frequencies
+#  - Removing channels from select scan list
+#
+# Exactly one priority scan channel is supported.  It cannot be unset.
+# If multiple memory locations are marked for priority scan, the last
+# one to be written will be the priority channel.
+#
+# The radio supports three AM modes, but CHIRP only knows about two,
+# so WAM on the radio is mapped to AM on read, and cannot be set with
+# CHIRP.
```
kk7ds: So, typically the way we handle this is expose any of the three as `mode=AM` and then add a `mem.extra` setting that controls "which one". Would that work here?

chirp/drivers/ar8200.py
```diff
@@ -0,0 +1,1640 @@
...
+#
+#  - Priority scan setting
+#  - Bank split size
+#  - Add or remove search pass frequencies
+#  - Change offset frequencies
+#  - Removing channels from select scan list
+#
+# Exactly one priority scan channel is supported.  It cannot be unset.
+# If multiple memory locations are marked for priority scan, the last
+# one to be written will be the priority channel.
```
kk7ds: Maybe just don't expose this as a channel flag and make it a radio setting instead where the user just picks the channel they want?

chirp/drivers/ar8200.py
```diff
@@ -0,0 +1,1640 @@
...
+#
+# Quick memory channel details beyond the frequency are not available,
+# as reading this from the serial port is not supported.
+#
+# Adding to the select scan list is not supported over the serial
+# port.
+#
+# Unclear what concurrency guarantees are necessary in class methods.
+# Notably, if one tries to update settings while downloading memory,
+# there could be inconsistent information returned for search banks.
```
kk7ds: If there's interaction you need to handle it yourself. We have some drivers that do their own caching and/or locking for this sort of thing (which I see you doing below).

chirp/drivers/ar8200.py
```diff
@@ -0,0 +1,1640 @@
...
+#
+# The radio allows arbitrary tuning steps to be set, in the range .05
+# kHz to 999.95 kHz in 50 Hz increments.  Providing all valid steps in
+# valid_tuning_steps is unwieldy, so only a variety of common steps
+# are used.
```
kk7ds: Do you even need to expose `valid_tuning_steps` in that case?

chirp/drivers/ar8200.py
```diff
@@ -0,0 +1,1640 @@
...
+#
+# Auto mode also lets the radio control duplex and offset, tuning
+# step, and step adjust.
+#
+# Radio must be power cycled after using CHIRP.  The serial protocol
+# supports a command to terminate the session and release the keypad,
+# but there is no suitable place in CHIRP to make this call.  By the
+# time the __del__ method is called, the serial port is already
+# closed.
```
kk7ds, line 96: Fair point. I don't know of any other live radio with this requirement, but we could consider making a change here.

On that point, if the radio has to be put into a special mode to do this maybe it'd be better to consider it a clone mode radio? You'd need to pack data into a format you can save/load but it might be worth doing it that way.

chirp/drivers/ar8200.py
```diff
@@ -0,0 +1,1640 @@
...
+                mem.empty = True
+            else:
+                if chan != int(res[1]):
+                    raise errors.RadioError(
+                        "Unexpected result returned from radio")
+
+                mem.freq = int(res[2])
+                mem.name = 'QM{:1d}'.format(chan)
+
+            mem.comment = 'Quick Memo {}'.format(chan)
```
kk7ds: `mem.comment` is not for you to use for arbitrary display, it's for the user to make notes about the channel. Zero (that I know of) radio drivers support this, and very few storage formats do. So assuming this is just using this field for an additional long label, this  needs to change.

kk7ds (commented):
Since this is live and I can't play with it, could you paste some screenshots of the memory grid and settings tab?

nothingface:
> Since this is live and I can't play with it, could you paste some screenshots of the memory grid and settings tab?

Sure, I'll do that once I address the other issues you noted, as there likely will be some changes to the layout.

chirp/drivers/ar8200.py
nothingface: Yes, that would work, I can do that.

chirp/drivers/ar8200.py
nothingface: I considered doing that; I may have even started implementing it before I went with the current approach.  Anyway, sure, that would work fine.

chirp/drivers/ar8200.py
nothingface: Understood.  The current locking only ensures that a command will get its own response.  But to work properly, get/set_settings cannot be called at the same time as get/set_memory, so I'll update the locking to work at that level.

chirp/drivers/ar8200.py
nothingface: I suppose not.  I was under the impression that `valid_tuning_steps` was necessary.  I guess, based on your comment, that its absence means any tuning step is valid (and not none), so I'll make that change.

chirp/drivers/ar8200.py
nothingface: There isn't a special mode for the radio per se.  Serial commands can be sent at any time.  Once a serial command has been received by the radio, the radio continues to operate, but the controls on the unit
are locked out until the exit serial command is issued.

Exit could be sent at the end of get/set memory/settings.  Sending it after every get/set_memory would be terrible for performance, and there is no way (unless I'm missing something) to know when the end of a batch of memory updates is.

Unfortunately, pretending to be a clone radio would have awful performance.  It is already not very fast (10-15 minutes to download, if all channels are full), and having to update all memory locations to make even a small change would not be ideal.

chirp/drivers/ar8200.py
nothingface: Understood.  It is such a label, so I will change it.  I could make it an extra field, unless there is a more appropriate place for such information, or you feel it should just be omitted.

For some context, the memories on the radio aren't stored or referenced sequentially by an index.  There are twenty scan banks, which can each be adjusted in size to a certain extent, plus search banks, and a few other miscellaneous memory locations.  I considered treating everything as special channels, so they could be identified appropriately, but that seemed excessive, and may have issues with bank resizing, as channels appear and disappear.  The note (currently in the comment field) helps the user understand what each channel corresponds to.  Without that, they would need to do some math and reference a few settings to see where an index lands on the radio.

chirp/drivers/ar8200.py
kk7ds: I think with `nostep_tuning` (which you set below) you should be good. That basically makes you like a CSV file which pretty much can represent anything and you can do your own validation where you need to.

chirp/drivers/ar8200.py
kk7ds: Yeah, none of those options sound good. I'd just leave it as-is for the moment but maybe we can add a close hook to make it better.

chirp/drivers/ar8200.py
kk7ds: Would it make sense to make it a multi-sub-device radio? We have some radios that are arbitrarily slice-able and expose the slices as sub-devices. It brings complexity which isn't always worth it, but it is appropriate sometimes.

Either way, the point of the driver is to adapt the radio to CHIRP's common model as much as possible, and CHIRP is set up for (basically) sequential numbered memories. So even though that might be weird for some radios, it's kinda how it has to be. I guess an immutable extra would be an option, but it's definitely an abuse of a different kind...

chirp/drivers/ar8200.py
nothingface: Sounds good.

chirp/drivers/ar8200.py
nothingface: Will do.

chirp/drivers/ar8200.py
nothingface: > Would it make sense to make it a multi-sub-device radio?

I don't think so.  I did consider that and looked at other drivers for examples, but that didn't seem to match the behavior of the AR8200 as well.

I do get the point that the goal is to have the driver adapt whatever the radio does into a model that is consistent with CHIRP.  Notwithstanding other (ab)use of extra and immutable fields, I think the radio does more or less map to a single list of memories.  It's just that the radio wants to call them things like A15 and f32 and search bank R.  And a given index might be H60 in one situation and h30 in another, depending on settings.  I'll add some documentation to make this clearer.  It would be handy, though not essential, for the user to see directly in the channel list what the radio would call a given channel.

kk7ds: Roger that. I'll have to think about a way to do what you want without changing things too much. This is the first radio I know of to really require that sort of thing, so I wouldn't want to go to deep.

The Kenwood commercial radios are zone-based. Each zone we represent as a sub-device, but within there they're still assigned an index, so it still fits pretty well. Making everything a special would not be good because those are not very portable between radios and don't even get exported in things like CSV.

nothingface:
I have pushed an updated branch that addresses the feedback from this discussion, specifically:

  - Remove valid_tuning_steps, as the radio allows arbitrary steps within a wide range,
  - Handle AM modes as an extra setting, as the radio has more modes than Chirp understands,
  - Handle priority scan channel as a setting,
  - Move communication lock to class methods,
  - Added documentation on mapping of Chirp channel numbers to radio channels, and
  - Remove comment (can be easily added again if an appropriate place is found for this information), and
  - Included screenshots
00-scan-memories
01-scan-memories
02-search-memories
03-search-extra
04-vfo-qm-memories
05-settings-main
06-settings-vfo
07-settings-offset
08-settings-scan
09-settings-scan
10-settings-search
11-settings-search
12-settings-search-protect

chirp/drivers/ar8200.py
```diff
@@ -0,0 +1,1692 @@
...
+            mem.empty = True
+
+        extra_select.set_mutable(False)
+
+        mem.immutable += ['extra.selectindex', 'tmode', 'rtone']
+
+        self._memcache[mem.number] = mem
+        return mem
+
+    def validate_memory(self, memory):
```
kk7ds, line 894: This really is intended to be a fast operation, so making it require the lock is unfortunate. Seems like the offset list could be cached if we have already pulled it and if not, only fetch that under the lock? Is it too slow to just fetch that at init so we have it?

kk7ds:
The validate comment is not critical, perhaps just for follow up, so I'll merge this.

PR #1540, opened 2026-04-09
pyproject.toml: Replace setup.py with minimal packaging metadata

dbast:
Move the existing setuptools metadata into a minimal pyproject.toml and drop the legacy setup.py entry point. This keeps current packaging behavior compatible while unlocking uv workflows without adding any uv-specific repo configuration.

That enables fast and reliable one-liner setup and run commands such as `uv run --all-extras tox` and `uv run --all-extras chirp`. uv can bootstrap a suitable Python, install the project and extras, and run commands in the managed environment.

A dummy setup.py shim could be kept for legacy invocation, but there seems to be no obvious in-tree reason to do that. Folding more config files into pyproject.toml can happen later, but is not required for this change and the uv unlock.

Ubuntu 22.04 ships a recent enough pip/setuptools stack for pyproject.toml based builds, so this should remain compatible with current CI and developer environments.


PolyCatDev:
I'm personally down for this but we gotta wait and see what core contributors think.

dbast:
Just to confirm: All commands in the wiki still work https://chirpmyradio.com/projects/chirp/wiki/DevelopersPython3Environment#Linux

So `pip install .` or `pip install -r requirements.txt` remain working.

What changed, but not part of the wiki: `python setup.py sdist` is now `python -m build --sdist` (a dummy = almost empty setup.py can make the old command working while setuptools read all the information from pyproject.toml)

dbast:
closing as there seems to be no interest in this.

PR #1541, opened 2026-04-09
radtel_rt900: Fix setting labels, remove duplicates, fix exception handling

rhaist:
## Summary

- Correct `ALMODE_LIST` labels (`"Site/Tone/Code"` → `"On Site/Send Sound/Send Code"`) and `DTMFST_LIST` labels/order (`"Off/KeyBoard Side Tone/..."` → `"Off/DT-ST/ANI-ST/DT+ANI"`), verified against physical hardware and vendor firmware source
- Remove duplicate TDR setting from the basic group (already present in the spec group)
- Remove bogus "DTMF Code" setting from the VFO A/B block (was incorrectly pointing at `settings.pttid` with a commented-out reference to a non-existent `vfodtmf.code` field)
- Fix `IndexError` → `ValueError` in `_encode_tone` DTCS lookup (`tuple.index()` raises `ValueError`, not `IndexError`)
- Fix same exception mismatch in `apply_vfo_tone`
- Remove unreachable dead `case "Cross"` block in `_encode_tone` (`split_tone_encode()` always decomposes Cross mode before this code is reached)
- Document `unknown_9025` as consistently `0x01` across all RT-900 variants

## Test plan

- [x] All existing driver tests pass (`tox -e fast-driver`)
- [x] Verified ALMODE and DTMFST labels on physical RT-900 BT hardware
- [x] Verified DTMFST ordering against vendor firmware source

Related to #11708

### DIFF
--- a/chirp/drivers/radtel_rt900.py
+++ b/chirp/drivers/radtel_rt900.py
@@ -183,7 +183,7 @@ struct {
      tailcode:1;      //      Tail Code (RT-470L)
   u8 unknown_9023;    // 9023
   u8 single_mode ;    // 9024 Single Mode (RT-900 BT)
-  u8 unknown_9025;    // 9025
+  u8 unknown_9025;    // 9025 Always 0x01 on all RT-900 variants
   u8 unknown_9026;    // 9026
   u8 unknown_9027;    // 9027
   u8 unknown_9028;    // 9028
@@ -243,10 +243,10 @@ TXPOWER_LOW = 0x02
 
 ABR_LIST = ["On", "5 seconds", "10 seconds", "15 seconds", "20 seconds",
             "30 seconds", "1 minute", "2 minutes", "3 minutes"]
-ALMODE_LIST = ["Site", "Tone", "Code"]
+ALMODE_LIST = ["On Site", "Send Sound", "Send Code"]
 AUTOLK_LIST = ["Off"] + ABR_LIST[1:4]
 DTMFSPEED_LIST = ["50 ms", "100 ms", "200 ms", "300 ms", "500 ms"]
-DTMFST_LIST = ["Off", "KeyBoard Side Tone", "ANI Side Tone", "KB ST + ANI ST"]
+DTMFST_LIST = ["Off", "DT-ST", "ANI-ST", "DT+ANI"]
 DUALTX_LIST = ["Off", "A", "B"]
 LANGUAGE_LIST = ["English", "Chinese"]
 MDF_LIST = ["Name", "Frequency", "Channel"]
@@ -660,11 +660,6 @@ class RT900BT(chirp_common.CloneModeRadio):
         rset = MemSetting("settings.alarmsound", "Alarm Sound", rs)
         basic.append(rset)
 
-        # Menu 44: TDR
-        rs = RadioSettingValueBoolean(_settings.tdr)
-        rset = MemSetting("settings.tdr", "TDR", rs)
-        basic.append(rset)
-
         rs = RadioSettingValueInvertedBoolean(not _settings.fmradio)
         rset = MemSetting("settings.fmradio", "FM Radio", rs)
         basic.append(rset)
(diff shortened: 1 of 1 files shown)


TrimbleSoftware:
As the original author of the radtel_rt900 family of drivers, I'm thankful for the communities involvement with theses fixes.

I'm also working on a major enchantment of the RT-920 driver to add 3 new sub-devices for FM/AM/SSB channels as well as  the related settings. The change will also include Named Banks/Zones for the RT-920 and associated settings.

There will be a new variant driver for the Bajeton BJ7800 that will include the FM/AM/SSB devices and is based on the RT-920.

rhaist:
@TrimbleSoftware thank you - expecially for all the dilligent work on this. I think the community would benefit from some of the excellent improvement you made in your test driver from the mentioned issue. If you don't mind would like to cherry pick some for the stable driver here. Totally understand if you want to do that yourself.

TrimbleSoftware:
> @TrimbleSoftware thank you - expecially for all the dilligent work on this. I think the community would benefit from some of the excellent improvement you made in your test driver from the mentioned issue. If you don't mind would like to cherry pick some for the stable driver here. Totally understand if you want to do that yourself.

I have code ready for testing and have opened up a new Feature request issue t o track that mods. You can see and test the code here: 

https://chirpmyradio.com/issues/12445

If you're looking for something to do that would be very beneficial to this series of drives would be to figure out how to expose the radio firmware version to the CHIRP driver. With your understanding to the RT-900 BT firmware source you may be able the accomplish that. It's been on my holy-grail-wish-list for a long time...

rhaist:
@TrimbleSoftware — spent some time investigating the firmware-version wish. Short version: I don't think it's reachable with the current protocol, and wanted to share the findings before anyone burns more time on it.

**Source-side (non-BT open-source drop, 2025-03-05):** The version is a hardcoded C string literal — `sprintf(disBuf, "Firmware:%s", "V01  ")` in `Gui/DisplayMenu.c`. It lives in program flash (0x08000000+). The CPS `'R'` command in `CPS/ProgromFlash.c` only addresses SPI flash / RAM-backed structs in the 0x0000–0xFxxx space — there's no opcode that returns program-flash bytes, so even the exact offset wouldn't help.

**Empirical check on an RT-900 BT V1.20P dump:** Every region declared in `AddrMap.h` but not currently decoded by the driver — `MARCHINE_ID` (0xB000), `MARCHINE_NAME` (0xB0C0), `FACTORY_DATE` (0xC000) — is all 0xFF. A printable-ASCII scan across the entire non-channel area turned up only coincidental runs inside the squelch calibration table. No version-looking string anywhere.

One side finding worth knowing: the driver's `_fingerprint`/`_fingerprint2` bytes are exactly the `BAND_ADDR` (0xF230) read that `ProgromSendFreqRange` returns — the V1.20P dump has `01 01 36 01 80 01 04 00 05 20 …` at 0xF230, and applying the source's `i==5/10/15/16` skip pattern yields byte-for-byte `_fingerprint[0]`. So those bytes are the sub-band table, not a disguised version field we could reinterpret.

Net: nothing in reachable memory tracks the firmware version on this family, and more sample dumps won't change that. Leaving this note here mainly so the next person who gets the same idea doesn't start from zero.

TrimbleSoftware:
> @TrimbleSoftware — spent some time investigating the firmware-version wish. Short version: I don't think it's reachable with the current protocol, and wanted to share the findings before anyone burns more time on it.
> 
> **Source-side (non-BT open-source drop, 2025-03-05):** The version is a hardcoded C string literal — `sprintf(disBuf, "Firmware:%s", "V01 ")` in `Gui/DisplayMenu.c`. It lives in program flash (0x08000000+). The CPS `'R'` command in `CPS/ProgromFlash.c` only addresses SPI flash / RAM-backed structs in the 0x0000–0xFxxx space — there's no opcode that returns program-flash bytes, so even the exact offset wouldn't help.
> 
> **Empirical check on an RT-900 BT V1.20P dump:** Every region declared in `AddrMap.h` but not currently decoded by the driver — `MARCHINE_ID` (0xB000), `MARCHINE_NAME` (0xB0C0), `FACTORY_DATE` (0xC000) — is all 0xFF. A printable-ASCII scan across the entire non-channel area turned up only coincidental runs inside the squelch calibration table. No version-looking string anywhere.
> 
> One side finding worth knowing: the driver's `_fingerprint`/`_fingerprint2` bytes are exactly the `BAND_ADDR` (0xF230) read that `ProgromSendFreqRange` returns — the V1.20P dump has `01 01 36 01 80 01 04 00 05 20 …` at 0xF230, and applying the source's `i==5/10/15/16` skip pattern yields byte-for-byte `_fingerprint[0]`. So those bytes are the sub-band table, not a disguised version field we could reinterpret.
> 
> Net: nothing in reachable memory tracks the firmware version on this family, and more sample dumps won't change that. Leaving this note here mainly so the next person who gets the same idea doesn't start from zero.

Thanks for looking into this. It would have been so easy for the firmware authors to place that version string into readable memory. Too bad they didn't do that. It makes it vary difficult for CHIRP drivers to handle the firmware differences.

PR #1543, opened 2026-04-10
radtel_rt900: Expose LearnFHSS per-channel setting

rhaist:
## Summary

Expose the per-channel `LearnFHSS` flag as a `RadioSetting` in the RT-900 driver's `mem.extra` group. The bit was already modeled in `MEM_FORMAT` as `learning:1` (byte F bit 0) but was never wired to a UI setting, so users could not toggle it from CHIRP.

One-line change in `get_memory`; round-trip is handled by the existing `mem.extra` setattr loop in `set_memory`.

## Where this comes from

While investigating the RT-900 series driver against the vendor open-source firmware release [`Radtel_RT900+open+source+20250305+.zip`](https://www.radtels.com/pages/software-download) (labeled "Radtel RT-900（无BT）Open Source 250305" — **non-BT**, 2025-03-05 snapshot), I noticed:

1. `chirp/drivers/radtel_rt900.py` defines a `learning:1` bitfield at byte F bit 0 of the channel struct, but no code in the driver reads or writes `_mem.learning` anywhere outside the struct declaration.
2. The vendor firmware source reads the same bit as `chFlag3.Bit.b0` / `fhssFlag` in `work/source/Core/Radio.c`:
   ```c
   fhssFlag = channelInfo.chFlag3.Byte & 0x01;
   ```
3. The OEM Radtel CPS exposes this bit to users as a per-channel column labeled **`LearnFHSS`** — i.e. the OEM combines both the firmware name ("FHSS") and CHIRP's working name ("Learn"). This confirmed the feature is real and user-facing but simply missing from CHIRP's UI.

## Verification

Tested against a real physical **RT-900 BT** on macOS:

1. Downloaded a baseline image from the radio with CHIRP.
2. Configured a test channel (ch. 34) in the OEM Radtel CPS with only `LearnFHSS` toggled (all other fields unchanged from a previous known-reference write).
3. Wrote the config back to the radio via OEM CPS.
4. Re-downloaded with CHIRP.
5. Hex-diffed the two CHIRP images. The only relevant change on channel 34 was byte F: `0x68 → 0x69` — a clean single-bit flip at bit 0, exactly where CHIRP models `learning:1`. No other bytes in the channel struct moved.

This establishes that CHIRP's bit location is correct and that the OEM feature maps 1:1 to the field CHIRP already has.

The same verification round incidentally confirmed CHIRP's modeling of byte F bits `narrow`, `encrypt`, `bcl`, `scan`, and `am_modulation` against the OEM CPS output (they all matched byte-for-byte on the same test channel).

## Test plan

- [x] `tools/cpep8.py chirp/drivers/radtel_rt900.py` — clean
- [x] `mypy --config-file .mypy.ini chirp/drivers/radtel_rt900.py` — clean
- [x] `pytest tests/test_drivers.py -k "Radtel_RT-9 or Radioddity_GS-10B"` — **213 passed** across all six variants that inherit from `RT900BT` (RT-900, RT-900_BT, RT-910, RT-910_BT, RT-920, Radioddity GS-10B), including `test_memory_extra_flat`, `test_memory_extra_frozen`, and `test_copy` which exercise the new setting's round-trip.
- [x] Empirical round-trip of the new setting verified against a physical RT-900 BT (see "Verification" above).

## Scope note

This PR intentionally does **not** touch `MEM_FORMAT`, does not rename the `learning` field, and does not change any other setting, label, or default. Keeping it to a minimal, strictly-tested change so the diff is easy to review. A follow-up investigation into scramble (which the OEM CPS itself cannot currently persist on this radio) and into exposing other hidden struct fields will come as separate PRs.

### DIFF
--- a/chirp/drivers/radtel_rt900.py
+++ b/chirp/drivers/radtel_rt900.py
@@ -71,8 +71,8 @@ struct {
      bcl:1,           //     BCL
      scan:1,          //     Scan  0 = Skip, 1 = Scan
      am_modulation:1, //     Per chan AM modulation
-     learning:1;      //     Learning
-  lbcd code[3];       // 0-2 Code
+     learning:1;      //     FHSS Learning
+  ul24 code;          // 0-2 FHSS Code (little-endian, 0-0x7FFFFF)
   u8 unknown6;        // 3
   char name[12];      // 4-F 12-character Alpha Tag
 } memory[%d];
@@ -1247,6 +1247,25 @@ class RT900BT(chirp_common.CloneModeRadio):
         rset = RadioSetting("learning", "LearnFHSS", rs)
         mem.extra.append(rset)
 
+        # FHSS Code (24-bit little-endian, range 0x000000-0x7FFFFF).
+        # Displayed and accepted as a 6-digit uppercase hex string to
+        # match the OEM CPS convention.
+        def validate_fhss_code(value):
+            try:
+                v = int(str(value).strip(), 16)
+            except ValueError:
+                raise InvalidValueError(
+                    "FHSS Code must be a hex value (e.g. 1A2B3C)")
+            if not (0 <= v <= 0x7FFFFF):
+                raise InvalidValueError(
+                    "FHSS Code must be between 000000 and 7FFFFF")
+            return value
+
+        rs = RadioSettingValueString(0, 6, "%06X" % int(_mem.code))
+        rs.set_validate_callback(validate_fhss_code)
+        rset = RadioSetting("fhss_code", "FHSS Code (hex)", rs)
+        mem.extra.append(rset)
+
         # PTT-ID
         rs = RadioSettingValueList(PTTID_LIST, current_index=_mem.pttid)
         rset = RadioSetting("pttid", "PTT ID", rs)
@@ -1328,7 +1347,10 @@ class RT900BT(chirp_common.CloneModeRadio):
                 _mem.narrow = 0b1
 
         for setting in mem.extra:
-            setattr(_mem, setting.get_name(), setting.value)
+            if setting.get_name() == 'fhss_code':
+                _mem.code = int(str(setting.value).strip(), 16)
+            else:
+                setattr(_mem, setting.get_name(), setting.value)
 
     def validate_memory(self, mem):
         msgs = []
--- a/chirp/drivers/radtel_rt900.py
+++ b/chirp/drivers/radtel_rt900.py
@@ -1261,7 +1261,8 @@ class RT900BT(chirp_common.CloneModeRadio):
                     "FHSS Code must be between 000000 and 7FFFFF")
             return value
 
-        rs = RadioSettingValueString(0, 6, "%06X" % (int(_mem.code) & 0x7FFFFF))
+        rs = RadioSettingValueString(
+            0, 6, "%06X" % (int(_mem.code) & 0x7FFFFF))
         rs.set_validate_callback(validate_fhss_code)
         rset = RadioSetting("fhss_code", "FHSS Code (hex)", rs)
         mem.extra.append(rset)
--- a/chirp/drivers/radtel_rt900.py
+++ b/chirp/drivers/radtel_rt900.py
@@ -1261,7 +1261,7 @@ class RT900BT(chirp_common.CloneModeRadio):
                     "FHSS Code must be between 000000 and 7FFFFF")
             return value
 
-        rs = RadioSettingValueString(0, 6, "%06X" % int(_mem.code))
+        rs = RadioSettingValueString(0, 6, "%06X" % (int(_mem.code) & 0x7FFFFF))
         rs.set_validate_callback(validate_fhss_code)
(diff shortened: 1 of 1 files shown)


TrimbleSoftware:
While you doing the change you may as well add the "FHSS Code" field to mem.extra too.

It is a 3 byte (24-bit) little-endian value stored right after the learning field.

Here is a proposed code change clip that shows a change to MEM_FORMAT that should make the parsing/encoding eaiser:

```
MEM_FORMAT = """
struct {
  lbcd rxfreq[4];     // 0-3
  lbcd txfreq[4];     // 4-7
  ul16 rxtone;        // 8-9
  ul16 txtone;        // A-B
  u8 unknown1:4,      // C
     scode:4;         //     Signaling
  u8 unknown2:6,      // D
     pttid:2;         //     PTT-ID
  u8 scramble:6,      //     Scramble
     txpower:2;       //     Power Level  0 = H, 1 = L, 2 = M
  u8 unknown4:1,      // F
     narrow:1,        //     Bandwidth  0 = Wide, 1 = Narrow
     encrypt:2,       //     Encrypt
     bcl:1,           //     BCL
     scan:1,          //     Scan  0 = Skip, 1 = Scan
     am_modulation:1, //     Per chan AM modulation
     learning:1;      //     FHSS Learning
  ul24 code;          // 0-2 FHSS Code
  u8 unknown6;        // 3
  char name[12];      // 4-F 12-character Alpha Tag
} memory[%d];
```

The most interesting part will be accepting and validating the hex input in the mem.extra UI to mimic the way the factory CPS works. It only accepts a range between 0-7FFFFF for the value.

TrimbleSoftware:
@rhaist

This change still needs more work before the next CHIRP release or there will be a lot of unhappy RT-9xx users out there. I tested the code against my RT-900BT and RT-920 hardware and the factory CPS and found these problems with it:

0) You'll have to allow for a "blank" FHSS Code (a null value) which will equate to 0xFFFFFF being stored in raw memory.
The current code defaults all null (0xFFFFFF) values to show 7FFFFF in the UI grid, but leaves raw memory at 0xFFFFFF. The Validation callback will also have to allow for a blank (null) code to be entered in order to allow a Code to be removed.

1) The memory Copy function (Crtl+C) doesn't work now either. It gets a "can't pickle local object" error on the validation function. Try moving the validation function up to the module level (instead of having it nested in the get_memory() def):

image

2) The factory CPS sets the driver's **unknown6** field value to 0XA0 when the Code value is set to anything but 0xFFFFFF.
That field appears to be some sort of _flag_ and  the definition should be renamed accordingly. The driver should also set this field to 0xA0 when the is a Code set to a non-null value. When to Code value is 0xFFFFFF the Flag value should be re-set back to 0xFF.

`  u8 unknown6;        // 3`

3) Consider using an apply callback instead of the conditional in set_memory().
This will take some updated code like:

```
        for setting in mem.extra:
            if setting.has_apply_callback():
                # use callbacks on mem.extra
                # Radiosettings that need postprocessing
                setting.run_apply_callback()
            else:
```

If I can be of any additional assistance on this let me know.

kk7ds:
It would be better if the extra didn't have a validation function at all, but the UI should also probably catch this and convert it to a non-native import.

Let me know if we should revert this until these issues are resolved. Thursday night is the deadline for what goes into the build so I'd like to either have this reverted or fixed by then.

Thanks to you both!

TrimbleSoftware:
> Let me know if we should revert this until these issues are resolved. Thursday night is the deadline for what goes into the build so I'd like to either have this reverted or fixed by then.


It would probably be best to _revert_ the commits for now. Then we can fix and test the code properly and not be rushed doing it!

kk7ds:
Sure, well, I'm fixing the UI failing to copy if it can't pickle the object in #1550 anyway, which I think probably prevents the worst of the behaviors you highlight (to me at least). But we'll wait for @rhaist to comment until at least tomorrow afternoon. If nothing by then I'll revert.

rhaist:
I will look into this as soon as possible but I don't think I can manage this before the release window closes.

I am sorry this caused issues. I will be more careful with future updates.

rhaist:
Posted #1552 with fixes for the four issues @TrimbleSoftware raised:

* Pickle / Ctrl+C copy — `validate_fhss_code` hoisted to a module-level function so the `RadioSetting` is no longer an unpicklable closure.
* Blank FHSS Code — raw 0xFFFFFF now displays blank, validator accepts blank input, `set_memory` writes 0xFFFFFF back on clear.
* `unknown6` renamed to `code_flag` and driven from `set_memory` (0xA0 when Code set, 0xFF when blank), matching the OEM CPS.
* Kept the conditional in `set_memory` instead of switching to `set_apply_callback` — `set_apply_callback` wraps its callback in a `lambda` and re-introduces the exact pickle issue we're fixing.

Verified with `tox -e style`, `tox -e fast-driver -- -k "Radtel_RT-9 or Radioddity_GS-10B"` (213 passed), and a physical RT-900BT round-trip. Thanks again @TrimbleSoftware for the thorough review and @kk7ds for the patience.

PR #1401, opened 2025-09-25
Added flathub data folder

PolyCatDev:
I'm trying to get CHIRP onto flathub and the maintainers told me that these files need to be stored and managed upstream.

So I added a folder called `flathub` with all the files they require to have upstream.

PR #1544, opened 2026-04-10
Added flathub upstream data folder

PolyCatDev:
I've managed to make a flatpak that should meet flathub's requirements: https://github.com/PolyCatDev/flathub/tree/com.chirpmyradio.chirp

The files in this PR are what flathub mandates be hosted upstream.

This PR is similar to #1401 but this time there is an actual decent chance that chirp might get accepted into flathub.

A bit of discussion has also happened in a feature request: https://chirpmyradio.com/issues/11482

### Special thanks

- @KR4ERF for helping out with [their deployment](https://github.com/KR4ERF/radio.kr4erf.chirp.Chirp).


kk7ds:
This really should be just one commit since the xml is added (incorrectly) and then fixed. No need to persist that in the history forever :)

kk7ds:
Squashed and also fixed some trailing whitespace

PR #1545, opened 2026-04-11
TK7162: add initial support

machshev:
Add support for the TK 7162 which seems to be a 16 channel variant of the TK 7160 with cut down features.

Tested all the settings and memory map as best as I can with the radio itself. Some settings are yet to be implemented, however this allows general radio use, and key programming.

- [x] Memory get
- [x] Memory set
- [x] General settings
- [x] Key config


machshev:
This appeared to work, but unfortunately it looks like the memory map is not as simple as I thought.
The radio seems to have two alternative firmwares built in - "5-tone" and "FleetSyn/DTMF". When this is changed in KPG-99D(E) it clears the memory... implying a different memory map.

I'm still trying to understand this, so this is not ready to merge yet.

machshev:
With "FleetSyn/DTMF" the memory map is very similar to the existing radio driver, so reverted to a sub class. I've had to add some extra class attributes to allow extra configuration.

chirp/drivers/tk8160.py
```diff
@@ -127,14 +127,14 @@
     'Key Lock',
     'Talk Around',
     'Scan',
-    'Scan Add',
-    '?',
+    'Scan Del/Add',
+    'Emergancy',
```
kk7ds: "Emergency" (spelling)

chirp/drivers/tk8160.py
```diff
@@ -667,3 +706,284 @@ class TK8160RadioM(TKx160Radio):
...
+    'Loan Worker',
+    'Squelch Level',
+    'Channel Entry',
+    'Squelch Off',
+    'Monitor Momentary',
+    'Squelch Off Momentary',
+    'Volume Up',
+    'Volume Down',
+]
+VALID_KEY_SELECTION_7162_TRIANGLE = VALID_KEY_SELECTION_7162 + ['Emergancy']
```
kk7ds: Here too

chirp/drivers/tk8160.py
```diff
@@ -667,3 +706,284 @@ class TK8160RadioM(TKx160Radio):
...
+        _mem.unknown2_1 = 0x00
+        _mem.unknown2_2 = 0x00
+        _mem.unknown2_3 = 0x00
+        _mem.unknown3_0 = 0x00
+        _mem.unknown3_1 = 0x00
+        _mem.unknown4 = 0x1F
+        _mem.unknown5[0] = 0xFF
+        _mem.unknown5[1] = 0xFF
+
+
```
kk7ds: Too many blank lines here.

chirp/drivers/tk8160.py
```diff
@@ -667,3 +706,284 @@ class TK8160RadioM(TKx160Radio):
...
+            keys.append(
+                settings.RadioSetting(
+                    name,
+                    desc,
+                    settings.RadioSettingValueList(
+                        options=options,
+                        current_index=current_index,
+                    ),
+                )
+            )
```
kk7ds: Can you please not format these like this? This is a popular current fad to blow all these out like this for machine formatting, but it wastes a ton of vertical space and I try to keep this kind of style out of the chirp tree.

chirp/drivers/tk8160.py
```diff
@@ -164,21 +164,53 @@
...
-KEYS = {i: key for i, key in enumerate(_KEYS)
-        if not key.startswith('?')}
+KEYS = {i: key for i, key in enumerate(_KEYS) if key in VALID_KEY_SELECTION}
 
 KEY_NAMES = {
+    'Left Up': 'leftup',
+    'Left Down': 'leftdn',
+    'Right Up': 'rightup',
+    'Right Down': 'rightdn',
+    'S': 'skey',
```
kk7ds, line 209: All these refactors/fixes to the base class would be much better as a separate commit ahead of the one to add the 7162. If you have trouble splitting them, let me know and I can help.

chirp/drivers/tk8160.py
```diff
@@ -667,3 +706,284 @@ class TK8160RadioM(TKx160Radio):
...
+            raise errors.RadioError('Unsupported duplex mode %r' % mem.duplex)
+
+        txtone, rxtone = chirp_common.split_tone_encode(mem)
+        _mem.rxtone = tk8180.KenwoodTKx180Radio._encode_tone(*rxtone)
+        _mem.txtone = tk8180.KenwoodTKx180Radio._encode_tone(*txtone)
+
+        _mem.wide = mem.mode == 'FM'
+        _mem.highpower = str(mem.power) == 'High'
+        _mem.scanadd = mem.skip == ''
+        _mem.name = mem.name[:8].ljust(8)
```
kk7ds: Seems like a lot of this is probably duplicated from the base class. What are the actual differences you need for this model?

kk7ds (commented):
Thanks for working on this - I'm a huge kenwood commercial fan, so I'm glad to have more support in the tree.

Also, please remove `flake.lock` from the commits. I assume that was accidental.

chirp/drivers/tk8160.py
machshev: Turns out I didn't need this section

chirp/drivers/tk8160.py
machshev: Sure, but as a general comment it's really difficult to work in Chirp because of all the manual formatting. There is no consistency, some bits use single quotes others use double quotes. Different indents and preferences like this. I've had to spend a while working out how to turn off LSP auto formatting in my neovim setup otherwise it just goes and updates the whole file (this is the second time now).

Have you considered adding a `ruff format` config that enforces a style that you are happy with? It might mean having a little bit of a compromise, but there are lots of different settings and the biggest change to most files seems to be the quotes style choice.

This would really help contributing. It's a few years now since I've manually formatted Python code, and it's so nice not to have to think about and internally absorb all the formatting quirks for each repo I work on. Makes it easier to actually focus on adding functionality.

If you could maybe consider it, I'd be very grateful.

chirp/drivers/tk8160.py
machshev: I can revert this if you want. I updated it to match the official programming software while doing A/B comparisons working out the values for the new keys.

I've run out of time this evening, but can maybe get to it in a few days time.

chirp/drivers/tk8160.py
machshev: Yes I came to the same conclusion, I've reverted this whole method.

chirp/drivers/tk8160.py
kk7ds: Thanks, but I have to deal with machine-formatted code for my day job and I absolutely hate it.

chirp/drivers/tk8160.py
kk7ds: Just a separate commit is fine, and the fixes are appreciated, I just want to keep them bisectable from the new model. Again, I can do this if you want/need.

kk7ds:
I split this into the key fixes, the upload fix, and your actually-new code. Also fixed a couple other formatting nits, caveman-style.

I'd merge this now but will let you ack it on your tomorrow since I trod all over it. Hope it's okay for you.

kk7ds:
Oh also, can you include a test image for the 7162 since you've got some settings delta to make sure that at least gets run? I guess I could probably generate myself one with KPG-99, but I imagine you have one handy.

machshev:
> Oh also, can you include a test image for the 7162 since you've got some settings delta to make sure that at least gets run? I guess I could probably generate myself one with KPG-99, but I imagine you have one handy.

Sure, I've just uploaded the image I'm now using in my radio. Hopefully this is okay?
It's got some local repeaters and then the UK VHF simplex channels. There are still some settings missing in this PR, but maybe better to add them later in a smaller PR to save having to re-review the whole thing again.

Hopefully this is ready to merge now?

I've also got a TK-8302 on my desk... I'm hoping it's very similar and just needs the ident and bands updating (UHF version of the 7162?).

kk7ds:
Yep, images with real data in them are good, thanks. I wouldn't hold my breath on the 8302, that's quite a bit newer I think :) It might be more like the 8180, although sometimes the limited channel radios are quite a bit simpler. Might be closer to the 8102.

PR #1547, opened 2026-04-14
vx150: Add driver for Vertex Standard VX-150

srw777:
Supports all 199 memories, Home channel, L1-L5/U1-U5 scan limits, channel names, and the full settings menu.  Tested with a real radio.


chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+#seekto 0x0008;
+struct {
+    u8 flags;
+    u8 freq_flags;
+    u8 freq_khz;
+    u8 freq_mhz;
+    u8 unknown1[2];
+    u8 offset_low;
+    u8 offset_high;
+} vfo_active;
```
kk7ds: Looks like all these VFO structures are the same? If so, please define the structure and then re-use it:
```
struct vfo {
   u8 flags;
    ....
};

struct vfo vfo_active;
...
struct vfo vfo_other;
```

chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+
+#seekto 0x0013;
+struct {
+    u8 step;
+} settings;
```
kk7ds: Does this need to be in a struct?

chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+    u8 offset_low;
+    u8 offset_high;
+} vfo_other;
+
+#seekto 0x0026;
+struct {
+    u8 mem_only:1,
+       unknown1:5,
+       power:2;
+} settings3;
```
kk7ds, line 113: Same.

chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+       narrow:1,
+       clock_shift:1,
+       unknown3:4;
+    u8 unknown4;
+    u8 offset_low;
+    u8 offset_high;
+    u8 tone;
+    u8 dtcs;
+    u8 unknown5[6];
+} memory[199];
```
kk7ds: Looks like this is probably the same as `home` and also maybe actually the same as the VFOs and limit below? Please define and re-use whatever structs possibe.

chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+def _to_bcd(value):
+    """Encode an integer 0-99 as a single BCD byte."""
+    return ((value // 10) << 4) | (value % 10)
+
+
+def _decode_freq(mem):
+    """Return frequency in Hz from a memory struct entry."""
+    return (
+        (100 + _bcd(int(mem.freq_mhz))) * 1_000_000
+        + _bcd(int(mem.freq_khz)) * 10_000
```
kk7ds: Why do this instead of defining it as an actual BCD in the structure?

chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+        + (5000 if int(mem.freq_flags) & 0x80 else 0)
```
kk7ds: Please define the actual flag bit and don't do your own masking.

chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+        if raw16 == _EMPTY_CHANNEL:
+            mem.empty = True
+            return mem
+
+        mem.freq = _decode_freq(raw)
+        mem.skip = "S" if raw.skip else ""
+        mem.mode = "NFM" if raw.narrow else "FM"
+        mem.duplex = DUPLEX_MAP.get(int(raw.duplex), "")
+        mem.offset = (_bcd(int(raw.offset_high)) * 1_000_000 +
+                      _bcd(int(raw.offset_low)) * 10_000)
```
kk7ds: Same.

chirp/drivers/vx150.py
```diff
@@ -0,0 +1,869 @@
...
+        # Name
+        if raw_name is not None:
+            name = mem.name.rstrip()
+            encoded = _encode_str(name, CHARSET, 7)
+            for i in range(7):
+                raw_name.name[i] = encoded[i]
+            raw_name.show_name = 0x00 if name else 0xFF
+
+    @classmethod
+    def match_model(cls, filedata, filename):
```
kk7ds, line 776: Please see this: https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Image-detection

Just `return False` here since you're new-world-only.

kk7ds (commented):
I would say "thanks for working on this" but I'd be lying. Vertex radios have been nothing but a pain in the past :D (sarcasm aside I of course appreciate your efforts).

I recently removed a few that never got maintained and never really worked for most people. Have you tested this on multiple platforms? The timing stuff in the serial routines makes me very itchy and usually means it works in one or two places but not universally. I know Yaesu/Vertex radios in general (especially of this generation) are quite finicky, but we should try to make it as universal as possible.

chirp/drivers/vx150.py
srw777: Yeah.  I guess I didn't clean up a lot of cruft after my brute-force decoding the many memory dumps.  These aren't even needed.  I didn't implement any way to read or set these.  (I see the ADMS software allows setting the VFOs, but it seems kind of pointless to me.)  I'll remove them.

chirp/drivers/vx150.py
srw777: No. I was just following the pattern.  Changed.

chirp/drivers/vx150.py
srw777: How do I access the individual bits then?  In a struct I can access the bitfield easily.

chirp/drivers/vx150.py
srw777: Yes.  I have defined a single struct for channel and re-used it for memories, home, and limits.  Good call.

chirp/drivers/vx150.py
srw777: Thanks.  While typing my explanation, I realised I made a mistake here.  I have fixed it, used bitfield to peel out the 2.5khz part, and eliminated the bcd functions.

srw777:
Thanks for the review.  I think I have addressed all but one of your concerns, plus also de-structed ani and cwid.
I will go ahead and test this driver on different computers.  Honestly, even if you choose not to include it, it was mostly the proverbial itch to scratch.  It works on my VX-150.  In fact, the whole download/upload process feels more solid than with my FT-60.  I don't know how to describe it.  When and how long I press the PTT on the FT-60 seems to matter.  The VX-150 just does it's thing when I push PTT or MONI.

chirp/drivers/vx150.py
kk7ds: Totally pointless, IMHO. Lots of people implement that but I don't know why. Never have I ever wanted to set my VFO that way :)

chirp/drivers/vx150.py
kk7ds: Not a big deal, just makes the code to dereference it later longer.

chirp/drivers/vx150.py
kk7ds: Should still work on the bare memobj, but that's fine, scoping them to a settings struct makes reasonable sense for this at least.

kk7ds:
If you feel confident that it works, especially if it works on multiple platforms, we should put it in.

Once we're happy with it, it should be squashed down to a single "Add VX-150 support" commit and we should reference this issue so it gets tagged as complete: https://chirpmyradio.com/issues/893

Thanks!

srw777:
Well, I just spent the past 45 minutes trying unsuccessfully to get my counterfeit prolific Yaesu cable working on Windows so I'm not sure I'll be able to test there.  (Linux user)  I'll try to find a few more machines to test on, but they'll all be Linux, I guess.  I'll let you know.

kk7ds:
Likely need this procedure, which still works FWIW:

https://chirpmyradio.com/projects/chirp/wiki/ProlificDriverDeprecation

srw777:
Ugh. Clearly I don't know how to use git/github.  I think I got this right now.  Built a CH9340 programming cable and tested on a Debian 12 Ryzen 7 Pro 2700, a Windows 10 i5-4670, a windows 11 i5-8265U, and a Devuan 6 i5-8265U.  Squashed down to a single commit.

kk7ds:
Okay, any changes to the driver or just rebasing? github makes it stupid hard to just see what happened when there was a rebase. I'll take your word for it if nothing changed, else I can check the diffs manually.

srw777:
No changes since my last real commit.  (Fixing up the MEM_FORMAT stuff.)

PR #1549, opened 2026-04-15
tk760g: TK-372G correct variants & names. Issue #12446

ojfd:
Last byte is xfb, not xff. Verified using Kenwood KPG56D software. Naming changed to reflect naming in TK-372G service manual. Only 4 variants exist, all with K country code. There are no TK-372G models with M, E or other country codes

### DIFF
--- a/chirp/drivers/tk760g.py
+++ b/chirp/drivers/tk760g.py
@@ -1581,20 +1581,20 @@ class TK378G_Radios(Kenwood_Series_60G):
 
 @directory.register
 class TK372G_Radios(Kenwood_Series_60G):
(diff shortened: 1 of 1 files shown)


kk7ds:
This is not really rescue-able in its current state, but I'm guessing [698fc75](https://github.com/kk7ds/chirp/pull/1549/commits/698fc751027938117f26b1ed82698f41007055cb) is the meat here. I will cherry-pick that in a new PR (but with proper issue linkage). It also looks like you maybe created this in the web view or something because it doesn't have an actual/proper git user so I won't be able to credit you.

PR #1552, opened 2026-04-16
radtel_rt900: Fix FHSS Code blank handling, pickle, and code_flag byte

rhaist:
Follow-up to #1543. Addresses the issues @TrimbleSoftware raised after testing the merged FHSS Code support against RT-900BT and RT-920 hardware and the OEM CPS.

## Summary

* **Pickle / clipboard copy** — `validate_fhss_code` was a nested closure inside `get_memory()`. Nested functions are unpicklable, so `Ctrl+C` on any memory containing the FHSS Code setting raised `can't pickle local object` in the wxui memedit clipboard path. Hoisted to a module-level `_validate_fhss_code`.
* **Blank / null Code (0xFFFFFF)** — the prior code masked `_mem.code & 0x7FFFFF`, so an unset raw 0xFFFFFF rendered as `7FFFFF` in the UI. The validator now accepts blank input, the UI shows blank when the raw value is 0xFFFFFF, and `set_memory` writes 0xFFFFFF back when the field is cleared. Matches OEM CPS behavior for an unset Code.
* **`code_flag` byte (was `unknown6`)** — OEM CPS writes 0xA0 into this byte whenever a Code is set and 0xFF when the Code is cleared. Renamed the field and drove it from `set_memory` so CHIRP-saved images no longer diverge from CPS-saved images on that byte.

## Why not `set_apply_callback`

@TrimbleSoftware also suggested switching `set_memory`'s conditional to the standard `if setting.has_apply_callback(): setting.run_apply_callback()` dispatch (issue 3 in the review). I kept the conditional because `set_apply_callback` wraps the supplied callback in a `lambda`, which is itself unpicklable. Using it in a `mem.extra` setting would re-introduce the exact pickle bug we're fixing here. The conditional dispatch in `set_memory` keeps the same behavior without bringing back an unpicklable closure.

## Verification

* Empirically round-tripped an FHSS Code on a real RT-900BT: configured via CHIRP, uploaded, re-downloaded, setting persists.
* Exercised blank, in-range, out-of-range, and non-hex inputs through the validator.
* Confirmed `pickle.dumps(mem)` round-trips a memory whose `extra` contains the new setting (the path Ctrl+C uses).

## Test plan

- [x] `tox -e style` — clean (`cpep8` + `mypy`)
- [x] `tox -e fast-driver -- -k "Radtel_RT-9 or Radioddity_GS-10B"` — **213 passed**, including `test_memory_extra_flat`, `test_memory_extra_frozen`, and `test_copy` across all six variants that inherit from `RT900BT`
- [x] Empirical round-trip on physical RT-900 BT

Thanks to @TrimbleSoftware for the thorough hardware review of #1543.

### DIFF
--- a/chirp/drivers/radtel_rt900.py
+++ b/chirp/drivers/radtel_rt900.py
@@ -53,6 +53,37 @@ import struct
 
 LOG = logging.getLogger(__name__)
 
+# FHSS Code is a 24-bit per-channel value. The OEM CPS represents an
+# unset code as 0xFFFFFF in raw memory and as a blank field in the UI.
+# When a code is set, the OEM also writes 0xA0 into the adjacent flag
+# byte; when the code is cleared, that flag byte is restored to 0xFF.
+FHSS_CODE_NULL = 0xFFFFFF
+FHSS_CODE_FLAG_ACTIVE = 0xA0
+FHSS_CODE_FLAG_NULL = 0xFF
+
+
+def _fhss_code_to_text(raw_code):
+    raw = int(raw_code)
+    if raw == FHSS_CODE_NULL:
+        return ""
+    return "%06X" % (raw & 0x7FFFFF)
+
+
+def _validate_fhss_code(value):
+    s = str(value).strip()
+    if s == "":
+        return value
+    try:
+        v = int(s, 16)
+    except ValueError:
+        raise InvalidValueError(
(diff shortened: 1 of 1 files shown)


TrimbleSoftware:
Just tested this latest code mod. Looks good now. Thanks for the quick fix up.

kk7ds:
Thanks both!

PR #1557, opened 2026-04-18
Bug 12454 - Fix 935/XS20 new channel creation

mrtjr1:
Fixes #12454
Clear factory bits on channel creation to prevent incorrect values. Fix power level wattage settings for XS20 series

# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/kg935g.py
+++ b/chirp/drivers/kg935g.py
@@ -1660,16 +1660,16 @@ class KG935GRadio(chirp_common.CloneModeRadio,
         _nam = self._memobj.names[number]
 
         if mem.empty:
-            _mem.set_raw("\x00" * (_mem.size() // 8))
+            _mem.fill_raw(b'\x00')
             self._memobj.valid[number] = 0
-            self._memobj.names[number].set_raw("\x00" * (_nam.size() // 8))
+            self._memobj.names[number].fill_raw(b'\x00')
             return
 
         # If it's a NEW channel, wipe it to all 0s first
         # This prevents "ghost" data from previous radio use or factory junk
         if was_empty:
-            _mem.set_raw("\x00" * (_mem.size() // 8))
-            _nam.set_raw("\x00" * (_nam.size() // 8))
+            _mem.fill_raw(b'\x00')
+            _nam.fill_raw(b'\x00')
 
         _mem.rxfreq = int(mem.freq / 10)
         if mem.duplex == "off":
--- a/chirp/drivers/kg935g.py
+++ b/chirp/drivers/kg935g.py
@@ -1694,7 +1694,7 @@ class KG935GRadio(chirp_common.CloneModeRadio,
         if mem.power is not None:
             _mem.power = self._set_power(mem)
         else:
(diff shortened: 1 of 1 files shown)


chirp/drivers/kg935g.py
```diff
@@ -1662,6 +1665,12 @@ def set_memory(self, mem):
             self._memobj.names[number].set_raw("\x00" * (_nam.size() // 8))
             return
 
+        # If it's a NEW channel, wipe it to all 0s first
+        # This prevents "ghost" data from previous radio use or factory junk
+        if was_empty:
+            _mem.set_raw("\x00" * (_mem.size() // 8))
```
kk7ds: `fill_raw(b'\x00')`

chirp/drivers/kg935g.py
```diff
@@ -1685,7 +1694,7 @@ def set_memory(self, mem):
         if mem.power is not None:
             _mem.power = self._set_power(mem)
         else:
-            _mem.power = True
+            _mem.power = 0  # default to low power if not set
```
kk7ds: This is not related to "new channel creation" right? Just another fix, but unrelated correct?

chirp/drivers/kg935g.py
```diff
@@ -2504,8 +2513,8 @@ class KGXS20G(KG935GRadio):
     _model = b"KG-UV8D-A"
     _record_start = 0x79
     config_map = config_map_x20h
-    POWER_LEVELS = [chirp_common.PowerLevel("L", watts=0.5),
-                    chirp_common.PowerLevel("H", watts=5.5)]
+    POWER_LEVELS = [chirp_common.PowerLevel("L", watts=5.0),
+                    chirp_common.PowerLevel("H", watts=20.0)]
```
kk7ds, line 2517: This too, although probably related-enough to the power fix above to put in the same commit...

mrtjr1: Yeah - I agree.  Will look to do that.

chirp/drivers/kg935g.py
mrtjr1: It is related -   when creating a new channel this code will default to True (1) if not user specified, which is not a valid value for the XS20 series due to its silly power mapping of 0 and 2 for Low and High.   So the 1 has to be corrected later on in the remap function.   It helps provide consistency to what the power value will be, instead of sometimes Medium, sometimes High, sometimes Low. Will look to move to its own commit

chirp/drivers/kg935g.py
mrtjr1: ok

chirp/drivers/kg935g.py
```diff
@@ -1689,12 +1689,13 @@ def set_memory(self, mem):
         _mem.iswide = int(mem.mode == "FM")
         # set the tone
         self.tone_model.set_tone(mem, _mem)
+
```
kk7ds: Unrelated whitespace damage here.. I'll fix before I merge, thanks!

PR #1559, opened 2026-04-19
baofeng_uv17Pro: add boot screen image upload for BF5RM / AR-5RM

campbell-r-e:
## What this does

Adds a **Boot Screen** settings group to the `BF5RM` driver (Baofeng 5RM / AR-5RM) so users can upload a custom 160×128 boot image directly from CHIRP without needing a separate tool.

A single **Upload to Radio** click uploads the boot image first, then proceeds with normal channel programming.

## Protocol

- 115200 baud 8N1, DTR+RTS asserted
- Raw handshake: `PROGRAMBFNORMALU` → `0x06` → `0x44`
- Framed packets: `[0xA5][cmd][id_hi][id_lo][len_hi][len_lo][payload…][crc_hi][crc_lo]`  
  CRC-16/CCITT (poly 0x1021) over bytes 1…(5+len)
- Sequence: `HANDSHAKE(0x02)` → `ERASE(0x04)` → `SET_ADDRESS(0x03)` → `WRITE(0x57)`×40 → `OVER(0x06)`
- Flash base `0x0C0000`, erase package ID 17668, 40 × 1024-byte chunks
- HANDSHAKE is retried up to 5× to allow for firmware startup time after mode switch

## Image handling

- **BMP**: parsed natively, no extra dependencies
- **PNG / JPEG**: auto-scaled via Pillow if installed; graceful error if not
- Nearest-neighbour scaling to 160×128, RGB565 little-endian encoding

## UI changes

- Adds `RadioSettingValueFile` to `chirp/settings.py` — a `RadioSettingValueString` subclass that carries a wildcard filter string
- `ChirpSettingGrid._add_items()` in `chirp/wxui/common.py` detects `RadioSettingValueFile` and renders a `wx.propgrid.FileProperty` (native Browse button) instead of a plain text field

## Test plan

- [ ] Upload boot screen image (BMP, PNG, JPEG) on AR-5RM / Baofeng 5RM
- [ ] Verify normal channel upload still works after boot screen upload
- [ ] Verify graceful error if image file is missing or path is blank
- [ ] Verify other BF5RM subclasses (MaxTalker P15, K5-Plus) unaffected
- [ ] Verify no regression in settings grid for other radios using `RadioSettingValueString`

### DIFF
--- a/chirp/drivers/baofeng_uv17Pro.py
+++ b/chirp/drivers/baofeng_uv17Pro.py
@@ -15,16 +15,25 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
+import os
+import time
 
 from chirp.drivers import baofeng_common as bfc
 from chirp import chirp_common, directory, memmap, bandplan_na
 from chirp import bitwise
 from chirp.settings import RadioSetting, \
     RadioSettingValueBoolean, RadioSettingValueList, \
-    RadioSettingValueString, \
+    RadioSettingValueString, RadioSettingValueFile, \
     RadioSettings, RadioSettingGroup
 import struct
-from chirp import errors
+from chirp import checksum, errors
+
+try:
+    from PIL import Image as _PILImage
+    _PIL_AVAILABLE = True
+except ImportError:
+    _PILImage = None
+    _PIL_AVAILABLE = False
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -218,6 +220,352 @@ def _upload(radio):
...
+
+def _bs_crc16(data, offset, count):
+    """CRC-16/CCITT (poly 0x1021) over data[offset:offset+count]."""
+    crc = 0
+    for i in range(count):
+        crc ^= data[offset + i] << 8
+        for _ in range(8):
+            crc = ((crc << 1) ^ 0x1021) if (crc & 0x8000) else (crc << 1)
+        crc &= 0xFFFF
+    return crc
```
kk7ds, line 246: I believe we have this in `chirp/crc.py` already but if not, it should go there please.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -218,6 +220,352 @@ def _upload(radio):
...
+    frame[6 + n] = (crc >> 8) & 0xFF
+    frame[6 + n + 1] = crc & 0xFF
+    return bytes(frame)
+
+
+def _bs_recv(pipe, timeout=5.0):
+    """Read one framed response; returns (cmd_byte, payload) or raises."""
+    deadline = time.time() + timeout
+    buf = bytearray()
+    pipe.timeout = 0.05
```
kk7ds, line 270: This kind of pattern is very much discouraged as it leads to very inconsistent behavior on different platforms. Is there some reason you're doing this instead of reading specific lengths with a reasonable timeout?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -218,6 +220,352 @@ def _upload(radio):
...
+        raise errors.RadioError("Boot screen: not a valid BMP file")
+    px_off = struct.unpack_from('<I', raw, 10)[0]
+    width = struct.unpack_from('<i', raw, 18)[0]
+    height = struct.unpack_from('<i', raw, 22)[0]
+    bpp = struct.unpack_from('<H', raw, 28)[0]
+    if bpp != 24:
+        raise errors.RadioError(
+            "Boot screen: BMP must be 24-bit RGB (got %d-bit). "
+            "Re-save as 24-bit BMP or use PNG/JPEG with "
+            "Pillow installed." % bpp)
```
kk7ds, line 321: 99% of  CHIRP users are on windows with no idea what pillow is or any ability to control installing it into the frozen environment. I'd prefer something like "Unsupported format" (with those details) over "please install this extra library" (my words).

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -218,6 +220,352 @@ def _upload(radio):
...
+                     w, h, _BS_WIDTH, _BS_HEIGHT)
+            rows = _bs_scale_nn(rows, w, h, _BS_WIDTH, _BS_HEIGHT)
+        data = _bs_rows_to_rgb565(rows)
+    else:
+        try:
+            from PIL import Image
+        except ImportError:
+            import subprocess
+            import sys
+            LOG.info("Boot screen: Pillow not found, attempting install...")
```
kk7ds, line 386: Please don't do this. Even though this looks like an open-source python project, almost all of the users are installing binaries where this won't work. Even if it did, I wouldn't want software I use doing this.

I'm amenable to just putting PIL in the requirements (although I need to get some ducks in a row to do so). Can we change this to just not advertise the setting if PIL is not available?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -218,6 +220,352 @@ def _upload(radio):
...
+            img = img.resize((_BS_WIDTH, _BS_HEIGHT), Image.LANCZOS)
+        pixels = img.load()
+        buf = bytearray(_BS_WIDTH * _BS_HEIGHT * 2)
+        for y in range(_BS_HEIGHT):
+            for x in range(_BS_WIDTH):
+                r, g, b = pixels[x, y]
+                px = ((r >> 3) << 11) | ((g >> 2) << 5) | (b >> 3)
+                i = (y * _BS_WIDTH + x) * 2
+                buf[i] = px & 0xFF
+                buf[i + 1] = (px >> 8) & 0xFF
```
kk7ds, line 409: Some comments about what you're doing here would be much appreciated. As the maintainer of this for the last almost-20 years, please think of me and pretend you're likely to disappear in a year and that I have to continue fielding bug reports :)

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -218,6 +220,352 @@ def _upload(radio):
...
+    return False
+
+
+def _upload_boot_screen(radio, img_path):
+    """
+    Upload a boot screen image to the radio using the dedicated picture
+    protocol.  Called from BF5RM.sync_out() before the normal channel upload.
+    Uses the already-open radio.pipe serial port.
+    """
+    img_data = _bs_image_to_rgb565(img_path)
```
kk7ds, line 455: Can you make it do all this on `set_settings()` instead of during upload? That's the time when the user is giving you the input and expecting an error if it's not okay. AFAICT, the workflow here would be:

1. Provide a path to a file
2. Try to upload
3. Find out it's not the right format when it fails, loop back to 1

That's going to be frustrating for people... It's also less ideal to do that in the clone thread I think.

kk7ds (commented):
Thanks, this is cool, but let's please just bring PIL into the requirements and make it more of a "missing if PIL is not available" sort of thing.

I don't think I have any of the radios that support this - which platforms could you test builds on to make sure we get this right and that it works on all the ones we support? The builds are generated on older versions of all the platforms in order to support people running on older stuff, so we need to make sure that we have suitable constraints for PIL on the places it builds.

chirp/drivers/baofeng_uv17Pro.py
campbell-r-e: Switched to `checksum.crc16_xmodem()` from `chirp/checksum.py` — there's no `crc.py` in the tree but `checksum.py` has exactly what's needed.

chirp/drivers/baofeng_uv17Pro.py
campbell-r-e: Rewrote `_bs_recv()` to read specific lengths: `pipe.read(5)` for the fixed header, then `pipe.read(n + 2)` for payload + CRC, all under a single deadline-driven timeout. No more unbounded reads.

chirp/drivers/baofeng_uv17Pro.py
campbell-r-e: Error now reads: *"Boot screen: unsupported format — BMP must be 24-bit RGB (got %d-bit). Re-save as a 24-bit BMP."* The PIL-absent path says: *"unsupported format — only 24-bit BMP files are supported. Convert the image to a 24-bit BMP and try again."* Neither path mentions Pillow to the user.

chirp/drivers/baofeng_uv17Pro.py
campbell-r-e: Removed the `subprocess`/`pip install` block entirely. Pillow is now in `requirements.txt`; availability is checked once at import time. When absent, the file picker is BMP-only and the label reflects that. Happy for you to tighten the version constraint once your build targets are sorted.

chirp/drivers/baofeng_uv17Pro.py
campbell-r-e: Added numbered step comments to `_upload_boot_screen()` for each protocol phase: 1) raw handshake, 2) framed PROGRAM handshake, 3) erase flash, 4) set write address, 5) write chunks, 6) signal done. Let me know if anything still needs more detail.

chirp/drivers/baofeng_uv17Pro.py
campbell-r-e: Moved validation out of the upload thread. `_apply_boot_image` is wired as an apply callback in `get_settings()` — the framework calls it when `set_settings()` runs. It validates the path and converts to RGB565 immediately, storing the result in `self._boot_image_data`. `sync_out()` just checks `if self._boot_image_data` and sends the pre-converted bytes — no file I/O or format checks happen in the upload thread.

campbell-r-e:
PIL is now in `requirements.txt` with the version constraint left open — happy for you to tighten that once your build targets are sorted. Availability is checked once at import time; when absent the file picker falls back to BMP-only with no mention of Pillow in any user-facing message.

For platform testing: I can cover macOS 15.7 and Fedora 43. I don't have a Windows environment, so if you spot anything platform-specific in review I'm happy to iterate on it.

chirp/drivers/baofeng_uv17Pro.py
kk7ds: Sorry, we just renamed this file, so checksum is what I meant :)

kk7ds:
Thanks, I can probably test the windows part enough now that it happens in the validate step.

I'll have another look at this soon.

PR #1561, opened 2026-04-24
[mp31] Add comments and fix two bugs in get_settings

davejbuilds:
## Summary

- Document all `MEM_FORMAT`/`H777_SETTINGS2` struct fields, constants, and functions with docstrings and inline comments explaining encoding conventions (BCD frequencies, CTCSS×10, DTCS flag bits, inverted BCL/beatshift/squelch)
- Fix `get_settings` crash when `squelchlevel` EEPROM byte is `0xFF` (uninitialized): `9 - 255 = -246` is out of the 0–9 range, causing the Settings panel to silently drop the squelch control; fix clamps the stored value to 0–9 before inverting
- Fix `FutureWarning` in timeout `RadioSettingValueList` by switching from deprecated `current=` to `current_index=`

## Test plan

- [x] Loaded `tests/images/Baofeng_MP31.img` and verified all 38 channels read correctly
- [x] `get_settings` / `set_settings` round-trip passes with no warnings
- [x] `get_memory` / `set_memory` round-trip on all 38 channels passes
- [x] `get_features` returns correct bounds and band


kk7ds (changes requested):
Looks like this is is a copy and modification of the h777 driver. Please see this:

https://chirpmyradio.com/projects/chirp/wiki/DevelopersAdd_a_Radio#Do-NOT-just-copy-a-driver

This should be a subclass with only the changes necessary to handle the differences.

Thanks!

PR #1562, opened 2026-04-25
generic_csv: Accept power labels and fall back gracefully (#11959)

JWhiteUX:
## Summary

- Fixes the CSV import failure described in [issue 11959](https://chirpmyradio.com/issues/11959): a CSV exported from a radio whose `PowerLevel` objects use labels like `Low`/`High` (rather than wattages) cannot be re-imported. `parse_power()` raises `ValueError` on each row, every row is dropped, and the importer surfaces the result as `InvalidDataError("No channels found")` — the file is rejected outright.
- In `chirp/drivers/generic_csv.py`, `_parse_csv_data_line` now special-cases the `Power` column: case-insensitive labels `low`/`med`/`medium`/`mid`/`high` are mapped to `PowerLevel` objects with handheld-typical wattages (1W / 2.5W / 5W). This is best-effort — without the source radio's power table we can't know exact watts — but it preserves the label and avoids dropping the row.
- Unknown values still fall back to `DEFAULT_POWER_LEVEL` (50W), but the warning is now self-explanatory: it names the expected formats *and* the literal text the user will see in the GUI's Power column, instead of a terse "using default" message.

### Error message

**Before** — terse `Invalid power specification: ''` with no guidance on what was expected or what CHIRP did:

Before error

**After** — names the accepted formats *and* the literal value the user will see in the GUI's Power column:

After error

### Import outcome

**Before** — every row rejected, import fails with `No channels found`:

Before result

**After** — rows import successfully with HT-typical wattage applied for known labels (`Low` here on a UV-5R):

After result

## Test plan

- [x] `pytest tests/unit/test_csv.py` — 23/23 pass.
- [x] Updated `test_parse_unknown_power` to assert non-numeric labels no longer drop the row (default applied).
- [x] New `test_round_trip_label_power` covers `Low` / `HIGH` / `medium` (case-insensitive) round-tripping to `Low` / `High` / `Med` PowerLevels.
- [x] New `test_unrecognized_power_falls_back_to_default` covers the truly bogus value path.
- [x] Manual: opened a UV-5R `.img`, exported to CSV, re-imported — rows now load with `Low (1W)` / `High (5W)` instead of being rejected (see screenshots above).

## Notes / out of scope

A fuller fix would round-trip the *exact* radio-specific wattage by embedding the source radio's `valid_power_levels` in the CSV (or matching labels against the destination driver's power table on import). That requires a CSV format change and is left for a follow-up. The mapping in this PR handles the common HT case and unblocks the reported bug.

### DIFF
--- a/chirp/drivers/generic_csv.py
+++ b/chirp/drivers/generic_csv.py
@@ -23,6 +23,31 @@ from chirp import chirp_common, errors, directory
 LOG = logging.getLogger(__name__)
 DEFAULT_POWER_LEVEL = chirp_common.AutoNamedPowerLevel(50)
 
+# Handheld-typical wattages applied when a CSV's Power column contains a
+# label (e.g. "Low"/"High") instead of a numeric value. Without the source
+# radio's power table we can't know exact watts, so these are best-effort
+# defaults matching common HT (handheld transceiver) ranges.
+POWER_LABEL_DEFAULTS = {
+    'low':    ('Low', 1.0),
+    'high':   ('High', 5.0),
+    'med':    ('Med', 2.5),
+    'medium': ('Med', 2.5),
+    'mid':    ('Mid', 2.5),
+}
+
+
+def _interpret_power_label(value):
+    """If @value is a known power label (case-insensitive), return a
+    PowerLevel with HT-typical wattage. Otherwise return None."""
+    if not value:
+        return None
+    key = value.strip().lower()
+    spec = POWER_LABEL_DEFAULTS.get(key)
+    if spec is None:
+        return None
+    name, watts = spec
+    return chirp_common.PowerLevel(name, watts=watts)
+
 
 class OmittedHeaderError(Exception):
     """Internal exception to signal that a column has been omitted"""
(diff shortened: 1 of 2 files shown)


JWhiteUX (commented):
Woke up thinking about this and want to revise a few things before this gets reviewed.

**Switch label matching from a dict to a regex list**

The current dict only matches exact normalized strings. Real-world CSVs that have been round-tripped through Excel, Numbers, or hand-edited in text editors routinely contain variants like `LOW`, `lo`, `HI`, `Hi`, `med`, `MED`, `medium`. The CHIRP wiki itself acknowledges that general-purpose spreadsheet programs let users enter values that break the format, and most users hitting this bug are doing exactly that. A regex list catches the variants without bloating the lookup table:

```python
_POWER_LABEL_PATTERNS = [
    (re.compile(r'^hi(gh)?$',   re.IGNORECASE), 'High', 5.0),
    (re.compile(r'^lo(w)?$',    re.IGNORECASE), 'Low',  1.0),
    (re.compile(r'^med(ium)?$', re.IGNORECASE), 'Med',  2.5),
    (re.compile(r'^mid$',       re.IGNORECASE), 'Mid',  2.5),
]
```

Patterns require at least two characters to avoid ambiguous single-letter matches.

**Tighten log messaging**

Current warning is verbose and buries the actionable part. Reworded for clarity:

- Match path (now `LOG.warning`, was `LOG.info`): `"Power %r mapped to ~%.1fW (HT default); actual wattage depends on source radio"`
- No-match path: `"Power %r: invalid label; expected wattage (e.g. '5W') or label (Low/Med/High). Applying default (%r)"`

Promoting the match path to `warning` because a silent wattage substitution is exactly the kind of thing a user needs to know happened — the current `info` level means most users will never see it.

**Tests**

Adding cases for the regex variants (`lo`, `LO`, `LOW`, `hi`, `HI`, `HIGH`, `med`, `MED`, `medium`) and negative cases for single-character inputs (`l`, `h`, `m`) to confirm they fall through to the default path.

**Rationale and scope note**

The hardcoded HT-typical wattages are a known limitation — “Low” is 0.5W on an IC-V80, 1W on a UV-5R, 1.5W on a UV-K5, and 5W on an AnyTone 5888UV. This PR can’t resolve that at the CSV layer without radio context, which `get_features()` doesn’t have at parse time. The accurate label-to-watt mapping needs to happen at the copy-paste step into a destination radio image, where `valid_power_levels` is available. Happy to take that on as a follow-on PR — referenced in [#10283](https://chirpmyradio.com/issues/10283) — but wanted to keep this one scoped to “don’t drop the row on import.”

JWhiteUX:
Closed pending further testing

PR #1563, opened 2026-04-27
h777v4: Fix detected models

kk7ds:
A recent change to h777v4 re-added the enter-programming-mode step
to do_download() which is correct for non-detected subclasses but
not for those that did detect_from_serial() first.

Make the driver do this only if we're not a detect-or or detect-ee.

### DIFF
--- a/chirp/chirp_common.py
+++ b/chirp/chirp_common.py
@@ -1508,6 +1508,10 @@ class DetectableInterface:
         return extra + list(detected)
 
     @classmethod
+    def is_detected(cls):
+        return getattr(cls, '_DETECTED_BY', None) is not None
+
+    @classmethod
     def detect_model(cls, detected_cls):
         detected_attr = class_detected_models_attribute(cls)
         if getattr(cls, detected_attr, None) is None:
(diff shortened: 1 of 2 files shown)


kk7ds:
From the comment in https://github.com/kk7ds/chirp/commit/ba77073722c41a0eded7a87e84abbf5f345fda22#r183472724

@KC9HI @w5alc

N5RR:
I went ahead and modified my source files with the patch code in the pull request.

I can confirm the fix works properly and can now download and upload as intended. Thank you for the quick response and the great work on this awesome software.

kk7ds:
Great, thanks. We need to hear from @KC9HI to make sure the original models that use the detection also still work after this before I can merge.

KC9HI:
I have to figure out where I have stored these radios among the hundreds of others that I have. I will try to find them and get this done today.

KC9HI:
Radios located. Download, upload and loading of saved files works.

I did notice a spelling error in the comment. I believe that "opne" was intended to be "one".

Jim

kk7ds:
Thanks much for the test and typo spot Jim.

My radio librarian hasn't shown up for work in years, so I know how it is :)

PR #1564, opened 2026-04-28
baofeng_uv17Pro: add boot screen image upload for BF5RM / AR-5RM

campbell-r-e:
Continuation of #1559 (accidentally closed).

## What this does

Adds a **Boot Screen** settings group to the `BF5RM` driver (Baofeng 5RM / AR-5RM) so users can upload a custom 160×128 boot image directly from CHIRP without needing a separate tool.

A single **Upload to Radio** click uploads the boot image first, then proceeds with normal channel programming.

## Protocol

- 115200 baud 8N1, DTR+RTS asserted
- Raw handshake: `PROGRAMBFNORMALU` → `0x06` → `0x44`
- Framed packets: `[0xA5][cmd][id_hi][id_lo][len_hi][len_lo][payload…][crc_hi][crc_lo]`  
  CRC-16/CCITT (poly 0x1021) over bytes 1…(5+len)
- Sequence: `HANDSHAKE(0x02)` → `ERASE(0x04)` → `SET_ADDRESS(0x03)` → `WRITE(0x57)`×40 → `OVER(0x06)`
- Flash base `0x0C0000`, erase package ID 17668, 40 × 1024-byte chunks
- HANDSHAKE is retried up to 5× to allow for firmware startup time after mode switch

## Image handling

- **BMP**: parsed natively, no extra dependencies
- **PNG / JPEG**: auto-scaled via Pillow; PIL added to `requirements.txt`
- Nearest-neighbour scaling to 160×128, RGB565 little-endian encoding

## UI changes

- Adds `RadioSettingValueFile` to `chirp/settings.py` — a `RadioSettingValueString` subclass that carries a wildcard filter string
- `ChirpSettingGrid._add_items()` in `chirp/wxui/common.py` detects `RadioSettingValueFile` and renders a `wx.propgrid.FileProperty` (native Browse button) instead of a plain text field

## Test plan

- [ ] Upload boot screen image (BMP, PNG, JPEG) on AR-5RM / Baofeng 5RM
- [ ] Verify normal channel upload still works after boot screen upload
- [ ] Verify graceful error if image file is missing or path is blank
- [ ] Verify other BF5RM subclasses (MaxTalker P15, K5-Plus) unaffected
- [ ] Verify no regression in settings grid for other radios using `RadioSettingValueString`

### DIFF
--- a/chirp/drivers/baofeng_uv17Pro.py
+++ b/chirp/drivers/baofeng_uv17Pro.py
@@ -15,16 +15,25 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import logging
+import os
+import time
 
 from chirp.drivers import baofeng_common as bfc
 from chirp import chirp_common, directory, memmap, bandplan_na
 from chirp import bitwise
 from chirp.settings import RadioSetting, \
     RadioSettingValueBoolean, RadioSettingValueList, \
-    RadioSettingValueString, \
+    RadioSettingValueString, RadioSettingValueFile, \
     RadioSettings, RadioSettingGroup
 import struct
-from chirp import errors
+from chirp import checksum, errors
+
+try:
+    from PIL import Image as _PILImage
+    _PIL_AVAILABLE = True
+except ImportError:
+    _PILImage = None
+    _PIL_AVAILABLE = False
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


kk7ds:
There's really never any reason to re-open a PR like this, even if you accidentally close it. It's really best to keep the conversation in one place.

That said, I started work on the system to make some test builds with pillow to validate this but it now includes a ton of other driver stuff that I have issues with. Can you rebase this back on master and just include the boot screen stuff? If not, we can wait until that other stuff is merged, but I'd rather not.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -1760,6 +2157,8 @@ class BaofengUV5RMPlus(BF5RM):
     _has_pilot_tone = True
     _has_send_id_delay = True
     _has_scramble = True
+    _has_gps = True
+    _has_voxsw = True
```
kk7ds: Is this relevant to the boot screen thing?

kk7ds:
Hey, I'm sorry I haven't gotten back to this. The blocker is that I'm going to have to build a new macos build worker with something more recent than 10.11 and I just haven't had a chance to get it all working yet. I might just get a mac mini to do the builds as it's really laborious to maintain the current VM on KVM I'm using currently.

Can you re-open this so I can get back to testing it when I have something to work on?

PR #1567, opened 2026-04-29
tdh8 - Fixes 12466: TD-H3, H8 Gen 2 and H8 Gen 3 Adds VFO A&B Offset Direction. Also fixes VFO A&B BCL setting persistence

TrimbleSoftware:
tdh8 - Fixes 12466: TD-H3, H8 Gen 2 and H8 Gen 3 Adds VFO A&B Offset Direction. Also fixes VFO A&B BCL setting persistence.

This fixes two **long-standing** bugs in the TD-H3, TD-H8 Gen 2 and TD-H8 Gen 3 drivers:

0) VFO A and B Offsets and TX Frequencies were completely broken. The radio firmware expects a VFO A or B to have a TX Frequency in a certain memory location. The driver did not have a setting for the Offset direction value (+, - or Off). The direction is need to calculate the TX Frequency from the RX Frequency, the Offset value and the Offset Direction. The first commit address this problem. The driver was never even calculating or storing a TX Frequency at all. This fix required refactoring some of the memory variable names to eliminate duplicates names.

1) The VFO A and B settings had a setting for BCL (Busy Lock) that did not work properly as the setting was getting the value from the wrong place in memory. Naturally, this was causing the BCL value to appear to not persist between radio downloads/uploads as  the value was being written to the correct place, but being read from the wrong place. This bug was found in testing the bug reported in Issue 12466 by the reporting user.

2) Removed some duplicate redundant settings value lists and refactored the settings code to clean it up some.

User testing and results here:

https://chirpmyradio.com/issues/12466

It's **way-about-time** for this driver to get a total rewrite as it is very hard to maintain in its current state!

### DIFF
--- a/chirp/drivers/tdh8.py
+++ b/chirp/drivers/tdh8.py
@@ -106,8 +106,10 @@ struct {
 } settings;
 
 //#seekto 0x0CB8;
-lbcd offseta[4];
-lbcd offsetb[4];
+struct{
+    lbcd a[4];
+    lbcd b[4];
+}vfo_offset;
 
 #seekto 0x0CD8;
 struct{
@@ -148,7 +150,7 @@ struct {
      lowpower:2,
      wide:1,
      unused8:1,
-     offset:2;
+     dira:2;
   u8 unused10;
 } vfoa;
 
@@ -171,7 +173,7 @@ struct {
      lowpowerb:2,
      wideb:1,
      unused8:1,
-     offsetb:2;
+     dirb:2;
   u8 unused10;
 } vfob;
 
@@ -365,8 +367,10 @@ struct {
 } settings;
 
 //#seekto 0x0CB8;
-lbcd offseta[4];
-lbcd offsetb[4];
+struct{
+  lbcd a[4];
+  lbcd b[4];
+}vfo_offset;
 
 #seekto 0x0CD8;
 struct{
@@ -472,7 +476,7 @@ struct {
      lowpower:2,
      wide:1,
      unused8:1,
-     offset:2;
+     dira:2;
   u8 unused10;
 } vfoa;
 
@@ -495,7 +499,7 @@ struct {
      lowpowerb:2,
      wideb:1,
      unused8:1,
-     offsetb:2;
+     dirb:2;
   u8 unused10;
 } vfob;
 
@@ -589,8 +593,10 @@ struct {
 } settings;
 
 //#seekto 0x0CB8;
-lbcd offseta[4];
-lbcd offsetb[4];
+struct {
+lbcd a[4];
+lbcd b[4];
+}vfo_offset;
 
 #seekto 0x0CD8;
 struct{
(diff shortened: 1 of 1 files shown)


chirp/drivers/tdh8.py
```diff
@@ -365,8 +367,10 @@
 } settings;
 
 //#seekto 0x0CB8;
-lbcd offseta[4];
-lbcd offsetb[4];
+struct{
+  lbcd a[4];
+  lbcd b[4];
+}vfo_offset;
```
kk7ds, line 373: This is the same as above. Just define it once:
```
struct vfo_offset { ... };
```
and then use it multiple times:
```
struct vfo_offset offseta;
struct vfo_offset offsetb;
```
That said, aren't the `vfoa` and `vfob` the same structure? If so, then _they_ should also be shared and avoid the need to modify both of them and with differing names for that same thing.

kk7ds (commented):
This driver needs some serious memory structure cleanup.

Can I just ask though - why (on earth) do people care about setting their VFO current state from desktop software? I know it's there in the driver, I know it's there in the OEM software, but why (again, on earth) would anyone want to do this?

It also seems like it would be much better to just implement this as a special channel _if_ there's some good reason to do it.

Could we not just rip out the VFO setting here?

TrimbleSoftware:
> This driver needs some serious memory structure cleanup.
> 
> Can I just ask though - why (on earth) do people care about setting their VFO current state from desktop software? I know it's there in the driver, I know it's there in the OEM software, but why (again, on earth) would anyone want to do this?
> 
> It also seems like it would be much better to just implement this as a special channel _if_ there's some good reason to do it.
> 
> Could we not just rip out the VFO setting here?

If I do the driver rewrite I'd definitely implement the VFO A and VFC B as Special channels. However, that is a bit-out-of-scope for a bug fix.

The biggest problem with this bug is that it could leave the radio in an inoperable state that the users could not correct with out using the factory CPS to fix the bad values. Plus, it leaves the user totally confused as why the Offset Value is not being used for anything useful.

I'm not inclined to make that big of a change until such a time as the driver is rewritten...

chirp/drivers/tdh8.py
TrimbleSoftware: Again, not really in the scope for a simple bug fix. Of course this type of thing would be addressed within the total driver rewrite.

kk7ds:
Yeah, the rewrite is not for this bug for sure, although you are introducing duplicative definitions (on top of existing ones).

But still I feel like offering programming for the VFOs is just silly anyway. I understand the bug reporter apparently cares about this but it seems like a pointless feature with code that can (and clearly does) have a maintenance burden.

chirp/drivers/tdh8.py
kk7ds: Ack, I see these are in the two different H8 and H3 memory maps, so fair enough. Please just acknowledge my OCD-inspired "duplication is bad" tendencies :)

PR #1569, opened 2026-04-30
HA2 @issuse 12468: Set default DTMF System, TOT, Squelch Level

Tommy83033:
…elch Level 4

# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/retevis_ha2.py
+++ b/chirp/drivers/retevis_ha2.py
@@ -325,8 +325,17 @@ SIDE_KEY_LIST = [
     {"name": "Prog PTT", "id": 31}
 ]
 
+DTMF_SIGNALINGLIST_DISABLED = 15
+TOTTIME_DISABLED = 36
+SQUELCHLEVEL_DISABLED = 4
+
 
 def _get_memory(radio, mem, _mem, ch_index, isvfo=False):
+    was_empty = False
+    if not isvfo:
+        ch_index_dict = retevis_ha1g.get_ch_index(radio)
+        was_empty = ch_index not in ch_index_dict
+
     mem.extra = RadioSettingGroup("Extra", "extra")
     mem.extra.append(
         RadioSetting(
@@ -335,7 +344,8 @@ def _get_memory(radio, mem, _mem, ch_index, isvfo=False):
             RadioSettingValueList(
                 retevis_ha1g.get_namedict_by_items(
                     retevis_ha1g.TIMEOUTTIMER_LIST),
-                current_index=(_mem.tottime - 1))))
+                current_index=(
+                  was_empty and TOTTIME_DISABLED - 1 or (_mem.tottime - 1)))))
     mem.extra.append(
         RadioSetting(
             "totpermissions",
@@ -346,8 +356,10 @@ def _get_memory(radio, mem, _mem, ch_index, isvfo=False):
         RadioSetting(
             "rxsqlmode",
             "Squelch Level",
-            RadioSettingValueList(retevis_ha1g.SQUELCHLEVEL_LIST,
-                                  current_index=_mem.rxsqlmode)))
+            RadioSettingValueList(
+                retevis_ha1g.SQUELCHLEVEL_LIST,
+                current_index=(
+                    was_empty and SQUELCHLEVEL_DISABLED or _mem.rxsqlmode))))
     mem.extra.append(
         RadioSetting(
             "alarmlist",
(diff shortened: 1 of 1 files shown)


chirp/drivers/retevis_ha2.py
```diff
@@ -327,6 +327,14 @@
 
 
 def _get_memory(radio, mem, _mem, ch_index, isvfo=False):
+    if not isvfo:
+        ch_index_dict = retevis_ha1g.get_ch_index(radio)
+        if ch_index not in ch_index_dict:
+            _mem.dtmfsignalinglist = 15
+            _mem.tottime = 36
+            _mem.rxsqlmode = 4
+            mem.freq = 0
```
kk7ds: This line is dead code. It will always be overwritten by L399

chirp/drivers/retevis_ha2.py
```diff
@@ -327,6 +327,14 @@
 
 
 def _get_memory(radio, mem, _mem, ch_index, isvfo=False):
+    if not isvfo:
+        ch_index_dict = retevis_ha1g.get_ch_index(radio)
+        if ch_index not in ch_index_dict:
+            _mem.dtmfsignalinglist = 15
+            _mem.tottime = 36
+            _mem.rxsqlmode = 4
```
kk7ds: These are not okay to do. It causes us to modify the radio memory in the `get_memory()` path. We should never modify the radio memory until/unless `set_memory()` is called (and try to only modify the relevant bits for the channel in question). If you think about opening an image, as soon as we get all the memories, this will have changed a whole lot of bits all over the image. I know you probably want that because it will "fix" a bunch of the bad memories in existing images, but it makes it very difficult for debugging when CHIRP "sees" the memory as different than it is stored in the image when the user hasn't intended to make any changes.

I think what you want to do is put the defaulting logic below with the extra elements themselves. So here you could do:
```
if not isvfo:
    ch_index_dict = retevis_ha1g.get_ch_index(radio)
    was_empty = ch_index not in ch_index_dict
else:
    was_empty = False
```
and then make the suggested changes below.

chirp/drivers/retevis_ha2.py
```diff
@@ -459,6 +459,9 @@ def _set_memory(radio, mem, _mem, ch_index, isvfo=False):
 
     _mem.fill_raw(b"\x00")
     if mem.empty:
+        _mem.dtmfsignalinglist = 15
+        _mem.tottime = 36
+        _mem.rxsqlmode = 4
```
kk7ds: This is good, because it sets the values for these to sane defaults for when this memory is used again. It would also be good to have these either defined somewhere (i.e. `DTMF_SIGNALINGLIST_DISABLED = 15`).

Also, does the radio delete memories by marking their freq as `0` or `0xF` ? If the latter, CHIRP should do that here in the same way. Asking because you're checking for either/both values above in `get_memory()`.

chirp/drivers/retevis_ha2.py
kk7ds: Here we do something like:
```
was_empty and 15 or _mem.dtmfsignalinglist
```
so that if the channel is going from empty to non-empty, we use the default value of 15 instead of the zero that is stored in the empty channel memory.

Make sense?

chirp/drivers/retevis_ha2.py
```diff
@@ -388,25 +396,17 @@ def _get_memory(radio, mem, _mem, ch_index, isvfo=False):
...
-            return mem
-
     mem.freq = int(_mem.rxfreq)
     mem.name = radio.filter_name(str(_mem.alias).rstrip())
     if isvfo:
         mem.immutable += ["name"]
-    tx_freq = int(_mem.txfreq)
 
-    if mem.freq == 0 or mem.freq == 0xFFFFFFFF:
+    if mem.freq == 0 or _mem.rxfreq == 0xFFFFFFFF:
```
kk7ds: It is better to always examine the `_mem` values (rx and tx frequencies) here since that's what you really need to know (i.e. is the channel in the radio's memory empty or not). So I think both of these checks should be against `_mem.rxfreq`.

kk7ds (changes requested):
Thanks Tommy, I  understand the problem now. Just a couple comments about how this should be done slightly differently to be more robust.

kk7ds:
Tommy, sorry for the delay in responding to this. Github has had some outages recently and apparently my comments did not get committed when I made them before.

chirp/drivers/retevis_ha2.py
Tommy83033: Okay, thank you very much for providing the correct example. I think I roughly understand how to set the default value.

chirp/drivers/retevis_ha2.py
Tommy83033: The radio doesn't delete channels purely by setting frequency to 0 or 0xFFFFFFFF. It mainly implements channel deletion by adjusting the channel quantity channeldata.chnum and removing the corresponding index from the channel index list channeldata.chindex.

chirp/drivers/retevis_ha2.py
Tommy83033: Yeah! You're right. We just need to check _mem.rxfreq.

kk7ds:
These commits should be squashed, but I'll do that for you so we can merge this. Thanks!

PR #1574, opened 2026-05-05
Updated flatpak metainfo file to solve warnings from flathub linter

PolyCatDev:
There were a bunch of warnings and errors from the flathub linter regarding the flatpak metainfo file. This commit fixes them.


AsciiWolf:
Looks good!

PR #1575, opened 2026-05-05
#12481  ic2730: Add IC-2730E (European variant) support

benjaminchodroff:
Fixes #12481 - Add IC-2730E support

The IC-2730E speaks the initial clone handshake at 19200 baud (not 9600 like the IC-2730A) and only responds to the CLONE_ID query when the payload contains the radio's own model code with the last byte zeroed, instead of the all-zeros payload sent by every other Icom clone-mode driver. It also returns "Icom Inc.4D" as the end-of-clone frame instead of ".4E".

Add a thin IC2730ERadio subclass with these three overrides, leaving the existing IC-2730A driver untouched (no A variant hardware available for testing). Behavior is gated by a new opt-in flag _id_query_with_model on IcomCloneModeRadio with a default of False, so the 25 other Icom clone-mode drivers are unchanged. icf.IcomCloneModeRadio.sync_out now reads self.BAUD_RATE instead of hardcoding 9600; this is a no-op for every existing clone-mode driver (24 inherit the 9600 default; IcomM710Radio overrides sync_out itself).

Diagnosed by capturing the Icom CS-2730 protocol exchange under Parallels and isolating each variable; the radio responds only when both baud=19200 AND the query payload contains the model code.

Tested with a real IC-2730E: full clone-from-radio (21312 bytes) and clone-to-radio both succeed.


chirp/drivers/icf.py
```diff
@@ -335,8 +335,22 @@ def start_hispeed_clone(radio, cmd):
...
 
+def _id_query_payload(radio):
+    """Build the payload for the initial CLONE_ID query.
+
+    Most Icom clone-mode radios reply to a query with all-zero payload.
+    A few (e.g. IC-2730E) only respond when the query carries the radio's
+    own model code with the last byte zeroed. Drivers opt into the
+    latter by setting `_id_query_with_model = True`.
+    """
+    if getattr(radio, '_id_query_with_model', False):
```
kk7ds: Can you make this a flag on the top-level `IcomCloneModeRadio` class, defaulted to False, to avoid this need to test if the attr is missing or not? Makes it better for the next radio that needs to do this.

kk7ds (commented):
Really bizarre - icom radios are so consistent it seems weird that this one is different like this. Thanks for including the receipts for why you think (know) this is necessary.

benjaminchodroff:
@kk7ds Not sure why that check didn't pass, but hope this works for you!

kk7ds:
Failed initially because github. This needs rebasing anyway, so I'll pull it and see what's going on with the latest run.

PR #1576, opened 2026-05-06
Mp31 comments fixes

davejbuilds:
Adds Baofeng MP31 Radio as a subclass of H777Radio with only the changes
necessary to handle differences from the base driver.

Overrides:
- MP31_MEM_FORMAT: 38 channels starting at 0x0000 (vs 16 at 0x0010)
- settings2 at 0x026B (vs 0x03C0 in H777)
- _ranges and _memsize capped at 0x0400 (radio does not respond above this)
- get_features: memory bounds (1, 38)
- process_mmap: clamps uninitialized 0xFF settings2 fields on load

Also includes Pxton px999s alias (same hardware).

Tested on live hardware (macOS, CH340 cable, USB VID:PID=1A86:7523).
Radio identifies as P3107\xf7\x00\x00. All 38 channels, tones, power
levels, squelch, VOX, and timeout timer verified round-trip correctly.

Fixes #11505


chirp/drivers/h777.py
```diff
@@ -655,6 +655,97 @@ def match_model(cls, filedata, filename):
...
+    u8 voicelanguage;
+    u8 scan;
+    u8 vox;
+    u8 voxlevel;
+    u8 voxinhibitonrx;
+    u8 lowvolinhibittx;
+    u8 highvolinhibittx;
+    u8 alarm;
+    u8 fmradio;
+} settings;
```
kk7ds: Since this is mostly duplicated from the base class can you split out `settings2` like the rest of the subclasses do so you can reuse the main part of `MEM_FORMAT` like they do?

chirp/drivers/h777.py
```diff
@@ -655,6 +655,97 @@ def match_model(cls, filedata, filename):
...
+    VENDOR = 'Pxton'
+    MODEL = 'px999s'
+
+
+@directory.register
+class MP31Radio(H777Radio):
+    """Baofeng MP31"""
+    VENDOR = "Baofeng"
+    MODEL = "MP31"
+    ALIASES = [Pxtonpx999s]
```
kk7ds: This is the old way of doing aliases:

https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Aliases

However, for a completely identical clone like this with not even another brand sibling, the best option is to just add this to `chirp/share/model_alias_map.yaml`. That will get it represented on the front page of the site without adding another vendor to the clone dialog.

chirp/drivers/h777.py
```diff
@@ -655,6 +655,97 @@ def match_model(cls, filedata, filename):
...
+
+    def get_settings(self):
+        # Clamp settings2 values that may be 0xFF in uninitialized images
+        s2 = self._memobj.settings2
+        if int(s2.squelchlevel) > 9:
+            s2.squelchlevel = 0
+        if int(s2.timeouttimer) >= len(TIMEOUTTIMER_LIST):
+            s2.timeouttimer = 0
+        if int(s2.scanmode) >= len(self.SCANMODE_LIST):
+            s2.scanmode = 0
```
kk7ds: This is modifying memory in a "get" operation which you should not do:

https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Do-not-modify-input-in-read-paths

That said, in each of these cases the right way to handle it is to let the user just "fix it in post" with the "out of range" handling in the UI. For these they should show as yellow when they load bad values and the user can pick a sane value to set.

Was that not working for these three or did you just think it best to avoid that?

chirp/drivers/h777.py
```diff
@@ -655,6 +655,97 @@ def match_model(cls, filedata, filename):
...
+    def process_mmap(self):
+        self._memobj = bitwise.parse(MP31_MEM_FORMAT, self._mmap)
+        # Clamp settings2 fields that may be 0xFF in uninitialized images
+        s2 = self._memobj.settings2
+        if int(s2.squelchlevel) > 9:
+            s2.squelchlevel = 0
+        if int(s2.timeouttimer) >= len(TIMEOUTTIMER_LIST):
+            s2.timeouttimer = 0
+        if int(s2.scanmode) >= len(self.SCANMODE_LIST):
+            s2.scanmode = 0
```
kk7ds: Again, this is modifying memory in a read path, so please don't do that.

If it's causing problems, the best thing to do is modify/clamp what you _return_ from `get_settings()` but don't actually modify the radio memory.

chirp/drivers/h777.py
```diff
@@ -703,6 +794,28 @@ def test_encode_tone_none(self):
...
+
+    def test_mp31_ident(self):
+        radio = MP31Radio(None)
+        self.assertIn(b"P3107\xf7\x00\x00", radio.IDENT)
+
+    def test_mp31_alias(self):
+        radio = MP31Radio(None)
+        vendors = [a.VENDOR for a in radio.ALIASES]
+        self.assertIn('Pxton', vendors)
+
```
kk7ds: These should actually be in `tests/unit` somewhere, but they're pre-existingly wrong, so that's fine. I'll try to follow up and fix that for all these.

.gitignore
```diff
@@ -40,3 +40,5 @@ chirp/locale/.files
 result/
 # for nvim editor
 .nvim.lua
+venv/
+.DS_Store
```
kk7ds, line 43: the venv thing probably doesn't belong here but the .DS_Store one is fine, it should just be in its own commit.

kk7ds (commented):
YES, much better thanks very much for following up on this :)

Just a few things left, let me know if you need clarification or help making these changes.

Thanks!

chirp/drivers/h777.py
davejbuilds: I tried appending the full MEM_FORMAT but got Array cardinality mismatch. Instead I split MEM_FORMAT into MEM_FORMAT_CHANNELS (memory array only) and MEM_FORMAT_SETTINGS (the settings struct at 0x02B0), with MEM_FORMAT = MEM_FORMAT_CHANNELS + MEM_FORMAT_SETTINGS so existing subclasses are unaffected. MP31Radio.process_mmap then uses MP31_MEM_FORMAT + MP31_SETTINGS2 + MEM_FORMAT_SETTINGS to reuse the shared settings block. All tests pass locally — does this approach look right to you?

chirp/drivers/h777.py
davejbuilds: The yellow out-of-range handling was not triggering — it was a hard crash with ValueError: None is not in list when loading an image with 0xFF in those fields.

I've moved the clamping out of process_mmap and into get_settings(). It now saves the original values, temporarily sets clamped values for display, calls super().get_settings(), then restores the originals so radio memory is never modified.

chirp/drivers/h777.py
davejbuilds: Moved the Pxton px999s alias from a Python class in h777.py to chirp/share/model_alias_map.yaml as suggested. Also removed the Pxtonpx999s class and ALIASES line from MP31Radio. The YAML entry points to Baofeng MP31

chirp/drivers/h777.py
davejbuilds: Removed clamping from process_mmap entirely.

.gitignore
davejbuilds: Will split the .gitignore change into its own separate commit.

davejbuilds:
The unit test failure is the pre-existing AR-5 duplicate in ArcshellAR5 that exists in upstream master before my changes. Confirmed locally by checking out master and running the test

kk7ds:
No, the unit test is not failing on master. You can see it passed just a couple hours ago on #1577. Checked out master locally and confirmed it passes locally. It runs on every commit before merge, so it really can't be broken.

```
  unit: OK (28.50=setup[6.47]+cmd[22.02] seconds)
  congratulations :) (28.56 seconds)
$ gwhat
1ff87630 (HEAD -> work, origin/master, origin/HEAD) ic2730: Add IC-2730E (European variant) support
```

You've added your entire venv here (~1700 files) which might be related but definitely needs to be fixed:
image

chirp/drivers/h777.py
kk7ds: Modifying and restoring is still modifying. Once you get this cleaned up, I'll pull it down and figure out why you're not getting the out of range handling and fix it for you.

chirp/drivers/h777.py
kk7ds: Excellent, thanks

chirp/drivers/h777.py
kk7ds: I don't see why you need to do that when the others don't but the three-way split is also an option yeah. I'll have a look when you get this fixed.

chirp/drivers/h777.py
```diff
@@ -656,6 +659,80 @@ def match_model(cls, filedata, filename):
...
+    u8 scanmode;
+    u8 beep;
+    u8 sidekey;
+    u8 rxemergency;
+} settings2;
+"""
+
+
+@directory.register
+class MP31Radio(H777Radio):
```
kk7ds, line 697: The failed unit test is because you're inheriting from `H777Radio`, which has `ALIASES` set. Thus registering this class re-registers all those aliases. You need to do what the other subclasses in this file are doing, which is declare no legacy aliases. This is a good demonstration for why we don't do these this way anymore.

chirp/drivers/h777.py
kk7ds, line 635: Like this, see below.

chirp/share/model_alias_map.yaml
```diff
@@ -376,3 +376,6 @@ Quansheng:
 Powerful:
 - alt: Quansheng UV-K5
   model: UV-S5
+Pxton:
+- alt: Baofeng MP31
+  model: px999s
```
kk7ds: I do not see any reference to the model being `px999s` so I think this should be `999S`.

chirp/.DS_Store
kk7ds, line 1: This should not be added.

.gitignore
```diff
@@ -12,7 +12,6 @@ ignored/
 patch_queue/
 archived_builds/
 .tox/
-.venv/
```
kk7ds, line 15: Please don't make this change to the file.

kk7ds (commented):
Your first commit adds 1700 files from venv, your second one removes them all. You're also removing an existing line from gitignore but not adding the one your commit claims to be (`.DS_Store`). If you need help with the git part please say so.

davejbuilds:
Went ahead and addressed the out-of-range crash rather than leaving it for you. `super().get_settings()` logs the error but doesn't crash — it just leaves the affected `RadioSetting` values uninitialized (`_current = None`). The fix overrides `get_settings()`, walks the returned settings, and calls `set_value(0)` on any uninitialized ones to set a safe default. `self._memobj` is never touched. All driver tests now pass.

chirp/drivers/h777.py
```diff
@@ -708,6 +708,16 @@ def process_mmap(self):
...
             MP31_MEM_FORMAT + MP31_SETTINGS2 + MEM_FORMAT_SETTINGS, self._mmap)
 
+    def get_settings(self):
+        settings = super().get_settings()
+        for setting in settings.walk():
+            if setting.get_name() in ('settings2.squelchlevel',
+                                      'settings2.timeouttimer',
+                                      'settings2.scanmode'):
+                if not setting.value.initialized:
+                    setting.value.set_value(0)
```
kk7ds: This is not the right thing to do, and it doesn't fix it for the base class.

kk7ds:
Looks like you fixed the venv add, I'll pull this down and look at the proper fix for the range issues.

kk7ds:
Um, no, it's working just fine:
image
That's exactly what the UI is supposed to do if you have out-of-range values.. it flags them and lets the user pick a new valid value to clean up the image. Maybe the reason you think it's not working is because the driver tests fail? That's because your image has invalid values in it. The solution is to fix the image that we commit to the tree for testing.

So I will nuke your last commit and fix the image and push that back up here.

PR #1580, opened 2026-05-16
 baofeng_uv17Pro: Fixes #12251. Allow UV-5R Mini, Mini 5 and UV-5G Mini to upload to radio over a BLE serial connection

TrimbleSoftware:
baofeng_uv17Pro: Fixes #12251. Allow UV-5R Mini, Mini 5 and UV-5G Mini to upload to radio over a BLE serial connection

This PR has two commits that address  the long-standing issue where a Bluetooth Low Energy (BLE) serial connection could not be used to upload to a Baofeng UV-5R Mini, UV-5G Mini or the newer Mini 5 radios.

These two commits are a substantial refactor of a previous PR ([1463](https://github.com/kk7ds/chirp/pull/1463)) submitted by @Marcotic some months ago.

I'd like to give full credit here to him for discovering a solution to the problem and figuring out the logic to implement it. That PR seems to have gone stale and has not seen any activity for months now. Hence the reason for the refactoring and the submission of a new one.

The following will help describe the fix:

0) The radio firmware supports downloading and uploading  from/to the radio over a BLE connection
1) The BLE connection supports a block size of **0x40** for **downloads**
2) The BLE connection supports a block size of **0x80** for **uploads**
3) The upload blocks must be an even multiple of the size of the memory region being written to during a BLE  upload
4) Partial blocks (blocks not a full 0x80 bytes long) must be padded to a full BLE upload block size with \xff bytes
5) if a partial block is padded with any values that are not equal to \xff it will corrupt the radio memory
6) When a partial block has been padded an entry is added to the serial trace log file.

7) A serial connection over BLE has substantially more latency than a USB serial connection. During the **ident** process that latency is used during the "magic sending" exchange to measure that latency and use that OR  the port name to determine if the connection is over USB or BLE.

During testing it was discovered that the USB Serial connection cannot use a block size of 0x80 for uploads. When that was tried it would corrupt the radio memory and only 2 out or each 4 memories would be displayed.

A helper method has also been added to chirp_common to help determine of the current serial connection is over a BLE serial connection and can be used by other drivers that need to determine if they are operating over a BLE connection.

8) A new setting for Enabling/Disabling Bluetooth as also been added.

This has been tested on Windows, Linux and MacOS and is working as it should for both connections over USB Serial and BLE Serial.

See CHIRP Issues tracker: https://chirpmyradio.com/issues/12251

### DIFF
--- a/chirp/drivers/baofeng_uv17Pro.py
+++ b/chirp/drivers/baofeng_uv17Pro.py
@@ -19,6 +19,7 @@ import logging
 from chirp.drivers import baofeng_common as bfc
 from chirp import chirp_common, directory, memmap, bandplan_na
 from chirp import bitwise
+from chirp import platform
 from chirp.settings import RadioSetting, \
     RadioSettingValueBoolean, RadioSettingValueList, \
     RadioSettingValueString, \
@@ -158,7 +159,7 @@ def _download(radio):
                           radio.BLOCK_SIZE):
             frame = radio._make_read_frame(addr, radio.BLOCK_SIZE)
 
-            radio.pipe.log('Sending request for %04x' % addr)
+            radio.pipe.log('Sending request for 0x%04x' % addr)
             bfc._rawsend(radio, frame)
 
             d = bfc._rawrecv(radio, radio.BLOCK_SIZE + 4)
@@ -173,7 +174,6 @@ def _download(radio):
 
             # UI Update
             status.cur = len(data) // radio.BLOCK_SIZE
-            status.msg = "Cloning from radio..."
             radio.status_fn(status)
     return data
 
@@ -494,7 +494,7 @@ class UV17Pro(bfc.BaofengCommonHT):
 
     def _make_frame(self, cmd, addr, length, data=""):
         """Pack the info in the header format"""
-        frame = cmd + struct.pack(">i", addr)[2:] + struct.pack("b", length)
+        frame = cmd + struct.pack(">i", addr)[2:] + struct.pack("B", length)
         # add the data if set
         if len(data) != 0:
             frame += data
@@ -2358,7 +2358,10 @@ class UV5RMini(UV17Pro):
 
     MEM_TOTAL = 0x8240
 
+    _is_on_ble = False
+
     _has_support_for_banknames = False
+    _has_bt = True  # allow BT setting
 
     _idents = [MSTRING_UV17PROGPS]
     _mem_size = MEM_TOTAL
@@ -2384,6 +2387,81 @@ class UV5RMini(UV17Pro):
                    _uhf_range,
                    _uhf_rx2_range]
     MODES = UV17Pro.MODES + ['AM']
+
+    BLE_UP_BLOCK_SIZE = 0x80  # uploading to radio over BLE uses 0x80 blocksize
+
+    def sync_in(self):
+        self._is_on_ble = platform.get_platform().is_ble_serial(self.pipe)
+        LOG.debug('Detected BLE: %s' % self._is_on_ble)
+        return super().sync_in()
+
+    def sync_out(self):
+        self._is_on_ble = platform.get_platform().is_ble_serial(self.pipe)
+        LOG.debug('Detected BLE: %s' % self._is_on_ble)
+        return super().sync_out()
+
+    def _upload(radio):
+        """Upload to UV5R Mini (over BLE or USB/Serial)"""
+        # Put radio in program mode and identify it and
+        # determine if on BLE or USB/Serial connection
+        _do_ident(radio)
+
+        if radio._is_on_ble:  # BLE upload needs a diff blocksize
+            _blocksize = radio.BLE_UP_BLOCK_SIZE
+        else:
+            _blocksize = radio.BLOCK_SIZE
+
+        data = b""
+
+        # UI progress
+        status = chirp_common.Status()
+        status.cur = 0
+        status.max = radio.MEM_TOTAL // _blocksize
+        status.msg = "Cloning to radio on %s/Serial..." % \
+            ('BLE' if radio._is_on_ble else 'USB')
+        radio.status_fn(status)
+
+        data_addr = 0x00
+        radio_mem = radio.get_mmap()
+        radio.pipe.log('Uploading to radio using block size of 0x%04x'
+                       % _blocksize)
+        for i in range(len(radio.MEM_SIZES)):
+            MEM_SIZE = radio.MEM_SIZES[i]
+            MEM_START = radio.MEM_STARTS[i]
+            for addr in range(MEM_START, MEM_START + MEM_SIZE,
+                              _blocksize):
+                # calc min byte count to keep from reading too many bytes
+                byte_count = min(_blocksize, (MEM_START + MEM_SIZE) - addr)
+                data = radio_mem[data_addr:data_addr + byte_count]
+
+                if byte_count < _blocksize:  # does block need padding?
+                    radio.pipe.log('Padding partial block with \xff '
+                                   'by 0x%02x bytes' %
+                                   (_blocksize - len(data)))
+                    # pad partial block with \xff
+                    data += (b'\xff' * (_blocksize - len(data)))
+
+                if radio._uses_encr:
+                    data = _crypt(radio._encrsym, data)
+
+                data_addr += byte_count  # advance read pointer
+
+                frame = radio._make_frame(b"W", addr, _blocksize, data)
+                radio.pipe.log('Sending address %04x' % addr)
+                bfc._rawsend(radio, frame)
+
+                ack = bfc._rawrecv(radio, 1)
+                if ack != b"\x06":
+                    msg = "Bad ack writing block %04x" % addr
+                    raise errors.RadioError(msg)
+
+                # UI Update
+                status.cur = data_addr // _blocksize
+                radio.status_fn(status)
+
+        return data
+
+    upload_function = _upload
 
     _end_fmt = """
     // #seekto 0x8220;
--- a/chirp/platform.py
+++ b/chirp/platform.py
@@ -182,6 +182,9 @@ class UnixPlatform(Platform):
 
         return ver
 
+    def is_ble_serial(self, serial):
+        return serial.port.startswith('/tmp/ttyBLE')
+
 
 class Win32Platform(Platform):
     """A platform module suitable for Windows systems"""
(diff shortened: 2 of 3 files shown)


chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -2385,6 +2389,93 @@ class UV5RMini(UV17Pro):
                    _uhf_rx2_range]
     MODES = UV17Pro.MODES + ['AM']
 
+    BLE_UP_BLOCK_SIZE = 0x80  # uploading to radio over BLE uses 0x80 blocksize
+
+    # any serial latency greater than this is most likely a BLE connection
+    BLE_MIN_LATENCY = 0.19
```
kk7ds: Sorry, but this will be far too error-prone. Years (decades) ago, the CHIRP code had a lot more non-deterministic code that depended on timing because people felt like it was required and all it did was yield tons of bugs of the variety "driver X works on windows but not linux" or "driver X works with one USB serial device but not another". For example, I do a lot of my driver development with serial-over-lan devices for easier cable management and for better tracing when running OE software on Windows. Such an arrangement will surely trip this.

So, we need to figure out a better way to do this. Something deterministic, likely per-platform I guess. I don't think I have one of these devices to be able to examine what they look like on each platform. We might also want to work with the ble-serial developer to try to figure out how we can detect these.

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -494,7 +495,7 @@ def _make_read_frame(self, addr, length):
 
     def _make_frame(self, cmd, addr, length, data=""):
         """Pack the info in the header format"""
-        frame = cmd + struct.pack(">i", addr)[2:] + struct.pack("b", length)
+        frame = cmd + struct.pack(">i", addr)[2:] + struct.pack("B", length)
```
kk7ds, line 497: Just a comment, not your fault, but this code is already wrong. It should be a single `struct.pack(">ib", addr >> 16, length)` or the like. The point of `struct.pack` is to build structures, not just format individual data elements. Would be good if we could go ahead and fix that with this change.

chirp/chirp_common.py
```diff
@@ -2256,3 +2256,21 @@ def is_split(bands, freq1, freq2):
...
+
+
+def is_ble_conn(radio, latency, min_latency):
+    """Determine if the radio is using a BLE serial connection"""
+    # Serial connection latency is measured during a certian write & read to
+    # see how long the radio takes to respond to that sequence.
+    # Any latency greater than min_latency is considerd a virtual port over BLE
+    # OR if the port name is in the list of ble-serial BLE port substrings
+    # that match (not case sensitive):
+    # Any: port with BLE in name, Linux: /tmp/ttyBLE,
```
kk7ds: I think this is legit....

chirp/chirp_common.py
```diff
@@ -2256,3 +2256,21 @@ def is_split(bands, freq1, freq2):
...
+    # MacOS: /dev/ttys, Windows: COM9 - COM15
```
kk7ds: But these are not. Some regular serial drivers in macos will use the `/dev/ttys` pattern, and of course Windows can assign anything to any of the `COMn` names.

I'm not sure about macos options at the moment, but can you look at the system information output on windows when these are connected and see if there's some indication that `COMn` is a temporary BLE device such that we can detect it that way?

chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -2385,6 +2389,93 @@ class UV5RMini(UV17Pro):
...
+                     "latency of %6.4f on port %s" % (_serial_latency,
+                                                      radio.pipe.portstr))
+        else:
+            LOG.info("Radio is using UBS/Serial cable connection with "
+                     "latency of %6.4f on port %s" % (_serial_latency,
+                                                      radio.pipe.portstr))
+
+        return ack
+
+    def _upload(radio):
```
kk7ds, line 2403: Isn't this a full copy of the base upload function? If the only difference is the block size, then we should just make that dynamic in the base implementation and only ever let it be different in this case. Otherwise, these two things can (and will) diverge over time.

kk7ds (commented):
Sorry for the delay.. lots of "life" is getting in the way right now.

Thanks for picking this up, I really do appreciate it. My many scars from working on CHIRP for nearly 20 years at this point won't allow me to bless the timing-based approach (especially since it will obviously break with various other non-BLE cases as noted). I'm definitely willing to help try to figure out something deterministic so we can tell the difference.

kk7ds:
```python
import winreg
import sys

key = winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, r'HARDWARE\DEVICEMAP\SERIALCOMM')
devices = {}
for i in range(255):
    try:
        device, name, _ = winreg.EnumValue(key, i)
        devices[name] = device
    except EnvironmentError:
        break

try:
    device = devices.get('BLE')
except KeyError:
    print('No BLE device found')
    sys.exit(1)

pair = list(device)
pair[-2] = '1'
pair = ''.join(pair)
print('Found BLE device %s, guessing pair is %s' % (device, pair))
try:
    pairdev = dict(((v, k) for k, v in devices.items()))[pair]
except KeyError:
    print('Pair device not found!')
else:
    print('Pair device is %s' % pairdev)
```
Gives me this:
```
PS C:\Users\Dan> python .\detect_ble.py
Found BLE device \Device\com0com24, guessing pair is \Device\com0com14
Pair device is COM99
```

Should work deterministically for windows right? Linux we can guess by the port name. I guess I'm surprised it's not the same on macos... ?

kk7ds:
Yeah, looks like `/tmp/ttyBLE` on macos as well. So I think between that for the POSIX platforms and the code above for Windows, we can make this deterministic and reliable right?

chirp/drivers/baofeng_uv17Pro.py
TrimbleSoftware: The gross portion of the latency is caused by the time the radio takes to respond to the ident value being sent and reading the response. In all my testing on Windows and Linux I've found that it is pretty predictable  to use that to see if the connection is slow enough in response to be a BLE serial connection. It is usually an order of magnitude slower to make the round trip than a USB serial connection.

That being said, of course it would be best to find a hard-cut way to determine the connection type. I think the best way would be for CHIRP to fully integrate ble-serial/blink into the port selection process so the user doesn't have to figure out how to install or use ble-serial. It would be a part of CHIRP and users could select BLE devices from the port listbox and use them that way instead of having to run a separate command line process. Maybe something for the future, but clearly that would be a better way to support BLE connections in CHIRP.

chirp/drivers/baofeng_uv17Pro.py
TrimbleSoftware: It is not a "full-copy" of the base upload function. It has partial block _padding_ logic in it allow the memory segments that are shorter that the full block size to be padded inorder to be uploadable. In other words: if  the memory region is not evenly divisible by the block size,  the last remaining block has to be padded to the full size of the block and only with b'\xff' values or the radio memory will be corrupted on upload. A partial length block can not be uploaded.

For USB block size uploads the memory segments are always an even multiple of the block size of (0x40), but when doubling the block size to 0x80 for BLE, the memory lengths are no longer evenly divisible by the BLE block size and the partial blocks have to be padded. This is the entire gist of this solution.

These memory sizes are even divisible by 0x40, but not be 0x80:

    `MEM_SIZES = [0x8040, 0x0040, 0x01C0]` 

It might be possible to make the partial block padding logic _conditional_ in the base upload function, but it might be ugly...

TrimbleSoftware:
> Yeah, looks like `/tmp/ttyBLE` on macos as well. So I think between that for the POSIX platforms and the code above for Windows, we can make this deterministic and reliable right?

Yes, that should work. Do you want to code up a replacement for `chirp_common.is_ble_conn()` method and remove the latency parameters?

chirp/drivers/baofeng_uv17Pro.py
kk7ds: Okay, we can put a pin in that for the moment, but I want to come back to it.

kk7ds:
> Yes, that should work. Do you want to code up a replacement for chirp_common.is_ble_conn() method and remove the latency parameters?

Just making sure there's no missing "me" in there.. you're asking if I'll do that on top of this? If so, I can sure. I know I shamefully still have the lock on the field PR, so my track record here is bad. But, that is "nice to have" and this is more impactful so I'll try to prioritize it.

Should be in `chirp/platform.py` not `chirp_common.py` I think. Lemme take a crack at it...

kk7ds:
Almost completely untested (I don't have one of these to test with anyway). Kept as a separate commit on top so you can see what I changed, but we should squash it in if it works.

TrimbleSoftware:
> Almost completely untested (I don't have one of these to test with anyway). Kept as a separate commit on top so you can see what I changed, but we should squash it in if it works.

Tested your changes on Win11. BLE Serial detection works good _when_ it runs. The only way to get the __init__() to run to do the detection is to Download from the radio first. If you just open up a standing .img file __init__() never runs and is_ble_serial() is never run, so  the detection is never carried out and the upload will fail if the attempt is over a BLE serial connection. Close, but not quite there yet...

kk7ds:
Ah right, need to put it in sync_in/out.. I don't have any way to test it other than synthetically...

TrimbleSoftware:
> Ah right, need to put it in sync_in/out.. I don't have any way to test it other than synthetically...

Yes, it needs to run on each and every upload AND download. You'll never know when a user picks a different connection on the fly...

TrimbleSoftware:
> > Ah right, need to put it in sync_in/out.. I don't have any way to test it other than synthetically...
> 
> Yes, it needs to run on each and every upload AND download. You'll never know when a user picks a different connection on the fly...

Tested again and that change had fixed that problem. Good work!

chirp/platform.py
```diff
@@ -182,6 +182,9 @@ def os_version_string(self):
 
         return ver
 
+    def is_ble_serial(self, serial):
+        return serial.port == '/tmp/ttyBLE'
```
TrimbleSoftware: I think this would be better as:

        `return 'BLE' in serial.port.upper()`

That would allow Linux/MacOS setups to have more that one ble-serial connection at a time. /tmp/ttyBLE0, /tmp/tty/BLE1, etc.

kk7ds: That seems too loose to me, but perhaps `startswith('/tmp/ttyBLE')` would be enough

kk7ds:
Have you been able to test this on Linux or macos? I think either is enough but it would be good to confirm if possible.

TrimbleSoftware:
> Have you been able to test this on Linux or macos? I think either is enough but it would be good to confirm if possible.

I'm away from my Linux box right now, but I'll spin up a VM and test with that. Thanks for the heads-up on testing,,,

kk7ds:
Okay, I think it's good enough to not break anything current, so let's just merge it and we can circle back if it's not activating properly on POSIX platforms. It's pretty simple and hard to be wrong there anyway. Thanks!

chirp/platform.py
TrimbleSoftware: Agreed!

Marcotic:
Thanks for all the work, and for the credit!!

TrimbleSoftware:
> Okay, I think it's good enough to not break anything current, so let's just merge it and we can circle back if it's not activating properly on POSIX platforms. It's pretty simple and hard to be wrong there anyway. Thanks!

I finally got BLE working on my Debian 13 VM and tested the UV-5R Mini BLE upload and it worked great over the BLE Serial connection. So we're good to go!

PR #1585, opened 2026-05-27
Ha1g/ha2 boot zone scan alarm channel update

Tommy83033:
# CHIRP PR Guidelines

### DIFF
--- a/chirp/drivers/retevis_ha1g.py
+++ b/chirp/drivers/retevis_ha1g.py
@@ -447,6 +447,11 @@ class HA1GBankModel(chirp_common.BankModel):
 
         if not channels_in_bank and bank.index in self.get_used_zone_index():
             # disable bank
+            settings = self._radio._memobj.settings
+            if settings.homepoweronzone_1 == bank.index:
+                settings.homepoweronzone_1 = 0xFFF
+            if settings.homepoweronzone_2 == bank.index:
+                settings.homepoweronzone_2 = 0xFFF
             _zone_index = [x for x in self.get_used_zone_index()
                            if x != bank.index]
             self.zinfo.zoneindex = (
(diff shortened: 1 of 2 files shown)


chirp/drivers/retevis_ha1g.py
```diff
@@ -1294,6 +1298,69 @@ def set_ch_index(self, ch_index_list):
     self._ch_cache = ch_index_list.copy()
 
 
+def UpdatePowerOnChannels(self, ch_index, ch_index_list):
```
kk7ds: Please don't use CamelCase for function names. They should be lowercase with underscores.

Also, as I have mentioned, this needs to be an object method if it takes self. Do not pass self as `self` to module-level functions. This seems best to be a method on the radio object.

chirp/drivers/retevis_ha1g.py
```diff
@@ -1171,6 +1171,10 @@ def _set_memory(self, mem, _mem, ch_index):
         ch_index_dict.append(ch_index)
     elif ch_index in ch_index_dict and rx_freq <= 0:
         ch_index_dict.remove(ch_index)
+        UpdatePowerOnChannels(self, ch_index, ch_index_dict)
+        UpdateZoneChannels(self, ch_index)
+        UpdateScanChannels(self, ch_index, ch_index_dict)
+        UpdateAlarmChannels(self, ch_index)
```
kk7ds: Can you combine this into a single helper? The individual functions are fine, but there should be one that calls them all. Like:
```
def update_channel_indexes(self, ch_index, ch_index_dict):
 ...
```
and then just call that.

chirp/drivers/retevis_ha1g.py
Tommy83033: Okay, I've made the changes.

chirp/drivers/retevis_ha1g.py
Tommy83033: Thanks for the advice. I've done as suggested.

kk7ds:
Thanks, looks good now. This fixes a bug, right? I don't see the bug number in the commits...

Tommy83033:
> Thanks, looks good now. This fixes a bug, right? I don't see the bug number in the commits...

I found this issue during use but didn’t file a bug report. My teammate also encountered the same problem, so I’ve implemented a fix.

kk7ds:
Okay, just wanted to make sure we tagged it, if there was one.

PR #1587, opened 2026-05-30
tools/check_commit: enforce whole-word six/future detection

jose-d:
Fix word-boundary matching in check_commit six/future checks.

Now we match `six` word but not eg. `sixth` etc.


kk7ds:
This is a non-required check to try to catch some common things. It doesn't have to be perfect, and six/future are not really part of the picture anymore, so we can probably just remove those checks. But... sure :)

PR #1589, opened 2026-06-05
Adding Baofeng 5RH Pro with GPS - v2 firmware

deeja:
# DRAFT while I set up some testing, etc

### DIFF
--- a/chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
+++ b/chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
@@ -1,0 +1,900 @@
+import time
+import logging
+import random
+
+from chirp import bitwise
+from chirp import chirp_common
+from chirp import directory
+from chirp import errors
+from chirp import memmap
+from chirp.settings import (
+    RadioSettings, RadioSettingGroup, RadioSetting,
+    RadioSettingValueList, RadioSettingValueBoolean, RadioSettingValueString,
+)
+
+LOG = logging.getLogger(__name__)
+
+# Protocol constants
+HEADER_SYNC = bytearray(b"PROGRAM\x00")
+HEADER_SYNC_PIC = b"Picture\xff"  # boot-image handshake (no seed, no XOR)
+HEADER_INFO = b"INFORMATION"
+END_INFO = b"END\x00"
+
+# Accepted boot-image dimensions (width, height)
+BOOT_IMAGE_SIZES = [(160, 128), (240, 240), (240, 320)]
+T_INFO = bytearray(16)
+for _i in range(12, 16):
+    T_INFO[_i] = 0xFF
+
+# Data layout
+DATA_LEN = 49152
+CHN_SIZE = 48
+CHN_MAX = 640
+
+TONES = chirp_common.TONES
+DTCS = chirp_common.ALL_DTCS_CODES
+
+POWER_LEVELS = [
+    chirp_common.PowerLevel("Low", watts=0.5),
+    chirp_common.PowerLevel("High", watts=5),
+]
+
+DUPLEX = ["", "-", "+", "split"]
+MODES = ["FM", "NFM"]
+
+# Settings option lists (labels mirror the CPS general-settings form).
+# For these the stored byte equals the option index.
+SQL_LIST = ["OFF"] + [str(i) for i in range(1, 10)]
+TOT_LIST = [str(i * 15) for i in range(15)]          # 0..210s
+PRETOT_LIST = [str(i) for i in range(11)]            # 0..10s
+APO_LIST = ["OFF", "30", "60", "120", "240", "480"]  # minutes
+SAVE_LIST = ["OFF", "1:1", "1:2", "1:4"]
+DISP_LIST = ["Frequency", "Name", "Number", "Frequency+Name"]
+DUAL_LIST = ["Single band single watch", "Dual band dual watch",
+             "Dual band single watch"]
+MAINBAND_LIST = ["A", "B"]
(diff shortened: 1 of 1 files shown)


deeja:
Adding this as a draft for anyone interested in testing it. I will follow it up with some test files, etc

deeja:
@campbell-r-e I saw you were working on something similar to this. Maybe you could have a look over it.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,977 @@
...
+    def get_memory(self, number):
+        _mem = self._memobj.memory[number - 1]
+        mem = chirp_common.Memory()
+        mem.number = number
+
+        if not self._get_valid(number - 1):
+            mem.empty = True
+            return mem
+
+        # Frequency (BCD-in-hex encoded 32-bit)
```
kk7ds, line 812: Then it should be `lbcd` type. Do not add your own BCD parsing.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,977 @@
...
+
+        Matches CPS ConvertChnValidFlg (DataProtocol.cs:883).
+        """
+        flags = int(self._memobj.valid_flags[index // 8])
+        return ((flags >> (index % 8)) & 1) == 0
+
+    def _set_valid(self, index, valid):
+        byte_i = index // 8
+        bit = index % 8
+        flags = int(self._memobj.valid_flags[byte_i])
```
kk7ds, line 796: Use a `bit` array type in your memory map.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,977 @@
...
+            return (None, None, None)
+        pol = "R" if (dath & 0xC0) == 0xC0 else "N"
+        val = ((dath & 0x07) << 8) | datl
+        code = int("%X" % val)  # hex digits as decimal: 0x023 -> 23
+        return ("DTCS", code, pol)
+    if dath == 0:
+        return (None, None, None)
+    val = (dath << 8) | datl
+    freq = int("%X" % val)  # hex digits as decimal: 0x1000 -> 1000
+    return ("Tone", freq / 10.0, None)
```
kk7ds, line 194: This looks like the standard encoding that most of these manufacturers ripped off of Kenwood. So please use the `kenwood_tone.py` routines to parse it.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,977 @@
...
+        model_str = "".join(chr(b ^ seed) if b != 0xFF else "" for b in model_data).strip()
+        LOG.info(f"Radio model: {model_str}")
+
+    # Send direction (0x52='R' for read, 0x57='W' for write)
+    direction = 0x57 if is_write else 0x52
+    port.write(bytes([seed ^ direction]))
+
+    # H6: Receive encrypted 0x41
+    deadline = time.time() + 3.0
+    while time.time() < deadline:
```
kk7ds, line 299: Don't do this. Set (or inherit) a reasonably long timeout, read the number of bytes you're expecting. Don't try to implement a non-blocking read routine in python. It _will_ end up with frustrated users across platforms, USB drivers, and hardware.

deeja:
@kk7ds  - Hey thanks for the review. I'll take a look over it when I've got some time but it might not be for a while. If anyone wants to have a go at this PR then they are welcome to.

McNeight:
So I've been taking a crack at this, and merging in:
* A nearly complete driver by @campbell-r-e at https://github.com/campbell-r-e/chirp/tree/5rh-pro-clean
  * Also includes a new RadioSettingValueFile class, for picking out a boot image to upload.
* The memory map contributed by Mike Iacovacci at https://chirpmyradio.com/issues/11669 (specifically https://chirpmyradio.com/attachments/13201).
* Some of the settings information from Jonathan Deeds at https://chirpmyradio.com/issues/11669 (specifically https://chirpmyradio.com/attachments/13931).
* The I/O routines (handshake, xor read and write, etc.) from @deeja .
* The suggestions above by @kk7ds .
* Ignoring for now, but leaving open the possibility of taking changes from the Claude-influenced patch by "unit m" at https://chirpmyradio.com/issues/11669 (specifically https://chirpmyradio.com/attachments/15492).

I don't have all of the pieces in place yet, but I know the following needs work:
* "There are a ton of drivers with negative and unnecessary #seekto directives in their memory maps."
  * I have no idea what this looks like, and if it's already existing in this particular driver code.
* The combination of memory map and parsing of settings, at this point, might be redundant? I'm still wrapping my brain around that part.
* Is there one driver that can be held up as a "gold standard" of what driver code should look like?

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
McNeight: Almost.

Instead of a uint16, tone frequencies are stored as a 4 digit BCD and then divided by 10. Meaning a value of 0x1000 is a tone of 100.0 Hz, and not 409.6 Hz.

I've modified kenwood_tone to add a "bcd=False" parameter to the public functions, while passing the bcd flag on to the private functions. There should be no impact to existing code.

steinbrueckri:
see also https://github.com/kk7ds/chirp/pull/1625

PR #1590, opened 2026-06-07
Update memory map for RT-900 series radios, add support to RT-920 for FM, AM and HF and new model Bajeton BJ7800

TrimbleSoftware:
This is a substantial update of the Radtel 900 series radio driver and updates/adds the following:

0) **All RT-900 series radios:** Edit memory map to support recent firmware changes. Added ToolTips to all settings. Some models such as the RT-900 now have some of the PF Key memory locations shifted so that had to be handled by conditionals in get_settings(). Increased DTMF Code strings from 5 chars to 6 chars per recent firmware changes.

1) **RT-920:** Added named banks/zones and changed zone geometry to match recent firmware updates (10 zones of 99 frequencies). Zone names can be edited in the Banks tab now by double-clicking on the Zone/Bank name:

image

Zone Mode or Flat Mode switching is now supported in the driver in the settings.

2) **RT-920**: Added FM, AM and HF sub-devices complete with VFO channel in memory grid:

image

3) **Bajeton BJ7800**: Add new model as a sub-class of RT-920, but with static banks with different geometry (15 zones of 64 frequencies) and no named banks. Also added test image for it.

4) **RT-920**: Updated test image for FM, AM and HF sub-devices

CHIRP Issue tracker link: https://chirpmyradio.com/issues/12445


TrimbleSoftware:
NOTE: These changes do not alter the .img file size and DO NOT break compatibility with existing .img files. The previously unused locations in the .img file are now being used and the exact same areas of the image are still being downloaded from the radio. In some cases some additional areas are now being uploaded to propagate the new driver functionality to the radio.

chirp/drivers/radtel_rt900.py
```diff
@@ -40,13 +41,13 @@
     RadioSettingValueString,
     RadioSettingValueMap,
     MemSetting,
-    InvalidValueError
+    InvalidValueError,
```
kk7ds, line 44: I know this is annoying of me, but I really hate changing lines for no reason in a given patch. Trailing commas are better, but I'd much prefer them in a cleanup commit instead of one that has functional changes. There's a lot of it in this set, which makes a lot of these changes look like they're related to functional changes. If we go back and try to bisect a regression, we have to look at all of these things to see if they were related to a bug. If they're in a "gardening" patch that has no effect other than beautification, we can ignore it in such a search.

chirp/drivers/radtel_rt900.py
```diff
@@ -1207,7 +1421,7 @@ def get_memory(self, number):
         if mem.freq == 0:
             mem.empty = True
         # tx freq can be blank
-        if _mem.txfreq.get_raw() == b"\xFF\xFF\xFF\xFF":
+        if _mem.txfreq.get_raw() == b"\xff\xff\xff\xff":
```
kk7ds, line 1424: This and the `fill_raw()` changes are also "gardening" (unless you've changed some of those sizes in this patch).

chirp/drivers/radtel_rt900.py
```diff
@@ -1601,48 +1850,461 @@ class RT920(RT900BT):
...
+            return mem
+
+        mem.freq = freq
+
+        if mem.number > self._upper:
+            mem.immutable += ["name"]
+        else:
+            mem.name = str(_name.name).rstrip("\xff ")
+
+        mem.mode = "Auto"
```
kk7ds: I really regret adding "Auto" to the list of valid modes so I hate to use it when not required. What is the actual point/meaning of this here if it's immutable but the device claims to support USB, LSB, CW, etc?

kk7ds (commented):
Two grumpy complaints I won't hold you to, but legit question in the third about the mode=Auto thing.

chirp/drivers/radtel_rt900.py
TrimbleSoftware: I really would have rather "hidden" the Mode column as  all the HF memories' modulation are controlled by a "modulation setting" in a new settings panel:

image

And we did have this same discussion back in March about the Mode column in this driver:
https://lists.chirpmyradio.com/hyperkitty/list/developers@lists.chirpmyradio.com/thread/FITKH5M5RTPDZDWUWRGWBFGYJRWTO3JG/

So the Mode column is not useful, except for copy/paste ops.

So in retrospect I would say that the Mode is _sort-of-automatic_ as it is controlled by a sub-device wide setting.  

The HF sub-device has three different modulations (LSB, USB & CW) and only one can be in effect for all 15 of the HF frequencies at a time. There is no support for each frequency to have a different modulation. This behavior is patterned after the factory CPS.

chirp/drivers/radtel_rt900.py
TrimbleSoftware: Fair enough, I'll refrain from making these _beatification_ changes mixed in with other functional changes...

chirp/drivers/radtel_rt900.py
TrimbleSoftware: Ditto!

chirp/drivers/radtel_rt900.py
kk7ds: > I really would have rather "hidden" the Mode column as all the HF memories' modulation are controlled by a "modulation setting" in a new settings panel:

Hmm, are you saying that these memories are just stored in the radio's memory and the behavior of them is controlled *en masse* by the "how to interpret HF memories" setting? That's pretty ... crazy, of course, and doesn't really fit the model very well. I really want CHIRP to be the swiss-army-knife but I also don't really want to jump through insane hoops in every case and this feels pretty much like they phoned it in so they could put "HF" on the feature list...

> So the Mode column is not useful, except for copy/paste ops.

It's not just useful for copy-paste, it's required for correctness. Copying something out of here and pasting it into, say, an Icom HF radio will not work with Auto.

Some options:

1. Set these all to CW or USB - something generally universal and make mode immutable for these memories.
2. Make mode immutable, and return whatever mode is set in the global setting at the point at which we return them. If the radio is set to USB, then all the HF memories are returned as USB, with mode immutable. That way they at least sort of copy/paste between radios as expected.
3. Investigate allowing hiding the Mode column in combination with either of the above.

In general, I really (really) do not want to break the ability to sanely copy and paste between radios like this (which is one reason why Auto is problematic).

TrimbleSoftware: Just forced pushed a commit that gets the HF Memory grid Mode from the HF SSB Modulation Mode setting. If had to be conditionalized to filter out FM and AM modulation as that is not applicable for HF memories. No more auto Mode!

I think this is more what you're looking for.

This works relatively well, but I would like to be able to refresh the HF memories editor to update the Mode column value when the Setting changes. The setting set_volatile(True)  flag doesn't seem cause a refresh of the memory editors...

kk7ds: I need to add this for another driver too. I think I had some prototype code laying around for it. Either way, "last value" is better than "auto" IMHO, so thanks!

kk7ds:
FWIW, I also really like to avoid any individual commit that doesn't pass tests making it into the permanent history. Meaning, "Fix broken test" patches really shouldn't be in there. But, I've complained about a lot here and you've got a bunch of commits split out that probably makes squashing that one back in non-trivial, so I'll let you slide AGAIN :D

TrimbleSoftware:
> FWIW, I also really like to avoid any individual commit that doesn't pass tests making it into the permanent history. Meaning, "Fix broken test" patches really shouldn't be in there. But, I've complained about a lot here and you've got a bunch of commits split out that probably makes squashing that one back in non-trivial, so I'll let you slide AGAIN :D

Thanks, and I had always planned on a rebase with some squishes. So I was prepared to do that. Next time ask and I'd be glad to do it.

kk7ds:
> Thanks, and I had always planned on a rebase with some squishes. So I was prepared to do that. Next time ask and I'd be glad to do it.

Oh okay. I was trying to avoid a "Local man stabbed to death by his own github contributor, story at 11" sort of situation :)

PR #1591, opened 2026-06-12
Bank renumbering

eliggett:
For radios where each assigned channel into a bank requires a unique channel index to be assigned, this change adds an option to right-click on a bank's column header and choose "Renumber". The renumber feature numbers the channels sequentially and without any duplicate indices.

Claude was used to figure this out. But it was also checked out and fussed with by a human (me) and tested on my Icom ID-5100 which really needed this bank numbering feature.

PR checks pass. This PR is a single addition and only changes one file. 

--E
de W6EL

# CHIRP PR Guidelines


chirp/wxui/bankedit.py
```diff
@@ -179,6 +181,37 @@ def row2mem(self, row):
...
         return mem - self._features.memory_bounds[0]
 
+    def _colheader_right_click(self, event):
+        col = self._grid.XToCol(event.GetX(), event.GetY())
+        if col < self._meta_cols:
+            return
+        if not isinstance(self._bankmodel,
+                          chirp_common.MappingModelIndexInterface):
+            return
+        bank = self._bankmodel.get_mappings()[self.col2bank(col)]
```
kk7ds, line 191: Technically this could be an expensive operation and/or a query to the radio in the live case. We can stash a list of bank names ahead of time, get it from the column header label, or just not  make the context menu item have the name in it and make this a lot simpler right?

kk7ds (commented):
Thanks for working on this, thanks for being upfront about claude, and thanks for checking claude's work :)

I really don't care about bank stuff much myself and thus I need to look at this a little more deeply than just with my Saturday morning coffee to remind myself how all this works (and try it out). But, it looks sane. One comment inline about the query to the radio in the right-click handler. I think technically more of the bank editor should be done in the background so it's already probably pretty laggy for a slow driver, but a call in a right-click handler would be particularly annoying.

Thanks!

chirp/wxui/bankedit.py
eliggett: Good point. I think the bank name really isn't even needed come to think of it, especially as the user has already right-clicked on the column label. 

Re: GUI performance lag, I have another patch which makes the scrolling and listing of channels significantly quicker, but it's a bit more involved and I figured I would start with something simple first.

PR #1592, opened 2026-06-16
Add Baofeng BF-5RH as alias of UV-5RH

Yundin:
## Summary

The Baofeng BF-5RH is the same radio as the Baofeng UV-5RH, but it does not appear in the CHIRP model list. Expose it as an alias of the existing UV-5RH driver.

### DIFF
--- a/chirp/share/model_alias_map.yaml
+++ b/chirp/share/model_alias_map.yaml
@@ -143,6 +143,8 @@ Baofeng/Pofung:
   model: UV-5RG, RK, RQ, RS, RT, RU
 - alt: Baofeng 5RM
   model: UV-5RH Pro Max
+- alt: Baofeng UV-5RH
+  model: BF-5RH
 - alt: BF-F8HP
   model: UV-5RHP
 - alt: Radioddity UV-5RX3


chirp/drivers/baofeng_uv17Pro.py
```diff
@@ -2032,11 +2032,18 @@ def get_features(self):
...
+    VENDOR = "Baofeng"
+    MODEL = "BF-5RH"
+
+
 @directory.register
 class UV5RH(UV17Pro):
     VENDOR = "Baofeng"
     MODEL = "UV-5RH"
 
+    ALIASES = [BF5RHAlias]
```
kk7ds: This is the old/deprecated way of noting aliases. Please see this: https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Aliases

kk7ds: Since this is a direct/exact clone with just a slightly different name, please add this to the `model_alias_map.yaml` file instead of adding it to the code. Thanks!

Yundin: Got it, thanks! The mistake was that I didn’t notice the alias map on the front page and expected to see the exact model in the dropdown.

PR #1593, opened 2026-06-16
Normalize memory name for CSV output

amcdawes:
Normalize memory name for CSV output without using __setattr__ to handle immutable fields.

This fixes an issue in CSV export of FT60 img file where the original raised an error due to immutable fields.

### DIFF
--- a/chirp/drivers/generic_csv.py
+++ b/chirp/drivers/generic_csv.py
@@ -329,6 +329,9 @@ class CSVRadio(chirp_common.FileBackedRadio):
 
     def set_memory(self, newmem):
         newmem = newmem.dupe()
+        # unset immutable to allow name rstrip
+        newmem.immutable = []
+        newmem.name = newmem.name.rstrip()
         if newmem.power is None:
             newmem.power = DEFAULT_POWER_LEVEL
         else:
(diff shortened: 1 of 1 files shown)


chirp/drivers/generic_csv.py
```diff
@@ -329,6 +329,9 @@ def _grow(self, target):
 
     def set_memory(self, newmem):
         newmem = newmem.dupe()
+        # Normalize name for CSV output without going through __setattr__,
+        # since callers may pass memories with immutable fields.
+        newmem.__dict__["name"] = newmem.name.rstrip()
```
kk7ds: This doesn't make much sense to me. Please file or reference a bug with a reproducer/trace so I can see what the proper fix would be.

amcdawes: I apologize, I didn't see the full instructions for how to file an issue and jumped right to a PR (and an ugly one at that). This issue is filed as Bug #12526. Same error as #10437 but not resolved by that fix. I've also tested using
```
newmem.immutable = []
newmem.name = newmem.name.rstrip()
```
but don't want to be too hasty to file another PR without knowing the best fix

kk7ds: Okay, thanks, the 10437 is the bug number we want to reference in the fix.

The proper solution here is to unset the immutable list on the duplicate, yep.

Please don't create another PR. Force push to this one a single commit with the above fix and reference the bug in the commit message per the [PR guidelines](https://github.com/kk7ds/chirp/blob/master/.github/pull_request_template.md)

Thanks!

PR #1595, opened 2026-06-17
Bulk edit

macikm:
This pull request adds a **Bulk Edit** feature to the memory editor grid, enabling users to update specific fields across multiple selected memories in a single action. Currently, modifying channel attributes requires updating them one-by-one, which is tedious for large configuration changes.

### Changes
1. **Memory Editor Enhancements** (`chirp/wxui/memedit.py`):
   - Added robust row/cell/block selection retrieval to capture all selected rows on right-click.
   - Added a `Bulk Edit...` option to the context menu, enabled when multiple memories are selected.
   - Implemented `ChirpBulkEditDialog`, a dialog displaying all editable columns (choice menus or text fields) next to checkboxes that "unlock" the field for editing.
   - Implemented undo context wrapping so all bulk edits can be reverted in a single undo step.
2. **Localization Updates** (`chirp/locale/*.po`):
   - Re-extracted translation templates and updated catalog files for all languages.
   - Completed Czech (`cs.po`) translations for the new strings.

### How to Use
1. Select multiple rows or blocks of cells in the memory list grid (using `Shift` or `Ctrl`, or dragging).
2. Right-click the selection and select **Bulk Edit...** (Hromadná změna...).
3. In the dialog, check the box next to the attribute you want to change (e.g. *SList / Scanlists* or *Power*). This will unlock the control.
4. Set the new value and click **Update** (Aktualizovat). All selected memories will be updated simultaneously.


chirp/wxui/memedit.py
```diff
@@ -2839,6 +2912,90 @@ def __init__(self, radio, dvmemory):
...
+
+    def get_updates(self):
+        updates = {}
+        for coldef, chk in self.checkboxes.items():
+            if chk.IsChecked():
+                updates[coldef] = self.editors[coldef].GetValue()
+        return updates
+
+
 class ChirpMemPropDialog(wx.Dialog):
```
kk7ds, line 2999: This class right below yours literally does the same thing. Perhaps your AI coding assistant led you astray? What am I missing?

macikm: When I tried this function, I had a problem that the same frequency was copied to all marked channels, for example, even though I only changed the inclusion in the "scan list"
so I went with a dialog where the user checks which channel attributes they want to change

kk7ds: So you thought you'd just add another duplicate one that works almost the same way? Weird.

The existing dialog tracks changes automatically and shows in bold the ones that will be applied. It uses a bunch of other existing prop-editing machinery for a consistent behavior across platforms.

PR #1597, opened 2026-06-22
a11y: fix screen reader access to memory grid, settings grid, and radio info pages

denizsincar29:
## Summary

This PR fixes screen reader (NVDA/JAWS) accessibility across three areas of the wx GUI that were previously silent or unusable for blind/low-vision users:

1. **Memory grid** (`memedit.py`) — accessible name for the grid window, keyboard-reachable context menu (Apps key / Shift+F10), and per-cell announcement on navigation via [prismatoid](https://github.com/ethindp/prism).
2. **Settings grid** (`common.py`, `memedit.py`) — `wx.propgrid.PropertyGrid`-based Settings/Other Settings/Advanced Settings/Work Mode Settings pages now announce property name, value, type, and locked/unspecified state on navigation, speak pending value changes while editing (combo box / checkbox), and expose setting descriptions on F1.
3. **Radio Info PropertyGrids** (`radioinfo.py`, `common.py`) — the same row-level announcement is extended to the read-only Features/Driver/Icom/Image Metadata pages in the Info tab via a new lightweight helper that doesn't depend on `RadioSetting`.

I'm blind and use NVDA on Windows daily; this is based on real usage and confirmed against NVDA's developer-info dumps to identify exactly which widgets were reporting raw/unhelpful accessibility info. I've tested all three changes manually with NVDA.

## Why no test image / new driver

No new radio drivers are added; this is purely GUI accessibility plumbing, so the "new drivers need a test image" requirement doesn't apply.

## Not covered (left as follow-up)

The Browser tab's raw memory editor (`developer.py`) uses a hand-rolled `wx.lib.scrolledpanel.ScrolledPanel` layout rather than `PropertyGrid`/`Grid`, so it needs separate treatment — it's a developer-facing tool rather than a regular Settings page, so I left it out of scope here.

## Testing

- Manually tested with NVDA on Windows across all three grids (memory grid, settings grid incl. live edit announcements, radio info grids).
- `tools/cpep8.py` and `mypy --config-file .mypy.ini` pass clean on all changed files.
- Branch is rebased on current `master`, no merge commits.

There's no existing CHIRP issue tracker ticket for this yet — happy to file one if maintainers prefer that before merge.

### DIFF
--- a/chirp/wxui/accessibility.py
+++ b/chirp/wxui/accessibility.py
@@ -1,0 +1,429 @@
+# Copyright 2026 Deniz Sincar <denizsincar29@gmail.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Screen reader accessibility helpers for wx.grid.Grid and
+wx.propgrid.PropertyGrid.
+
+Both controls paint their own rows/cells instead of using one native
+child window per item, so under MSAA/IAccessible they expose a single
+ROLE_SYSTEM_CLIENT/PANE object for the whole control with no children:
+arrow-key navigation between rows is otherwise completely silent for
+screen reader users.
+
+An earlier version of this fix (see PR #1597) spoke each row directly
+through prismatoid (ethindp/prism) whenever the row changed. That
+worked, but it bypassed the OS accessibility layer entirely, so it
+fought with the narration wx already produces natively for the
+surrounding UI (tab control, panel labels, etc) -- two voices talking
+over each other -- and, on macOS, spoke constantly for every user
+regardless of whether a screen reader was even running, since nothing
+gated it on assistive technology actually being active.
+
+This version does the same job the way any native control does: by
+implementing wx.Accessible so the grid answers IAccessible queries
+(name/value/role/state/focused child) truthfully, and by calling
+wx.Accessible.NotifyEvent() when the focused row/cell changes so
+Windows re-queries it. No screen-reader-specific package or manual
+speech call is involved anywhere in this module -- wx.Accessible and
+NotifyEvent are both part of wxPython itself.
+
+Caveat: wx.Accessible is only implemented for MSW (see the wxWidgets
+docs for wx.Accessible -- "Availability: Only available for MSW").
+On other platforms attach_grid_accessible()/attach_propgrid_accessible()
+still set a plain accessible name (which GTK/Cocoa do surface) but
+install no per-row bridging, so this specifically fixes Windows/NVDA
+and does not attempt -- and does not risk regressing anything -- on
+Linux or macOS. That matches the double/forced-narration report in
+PR #1597, which was on macOS, a platform this module now does nothing
+extra on.
+"""
+
+import logging
+
+import wx
+
+from chirp.wxui import config
+
+LOG = logging.getLogger(__name__)
+
+CONF = config.get()
+
+
+def _accessible_supported():
+    """Probe whether wx.Accessible actually works on this platform.
+
+    wx.Accessible exists as a class attribute on every platform (it's
+    part of the Phoenix bindings), but it's only *implemented* for
+    MSW: on GTK/Cocoa builds ``wx.Accessible()`` raises
+    NotImplementedError from the underlying C++ the moment you try to
+    construct one, even though ``hasattr(wx, 'Accessible')`` is True
+    there too. So we have to actually try instantiating it rather than
+    checking for the attribute.
+    """
+    if CONF.get_bool('disable_accessibility', 'state', default=False):
+        return False
+    accessible_cls = getattr(wx, 'Accessible', None)
+    if accessible_cls is None:
+        return False
+    try:
+        accessible_cls()
+    except NotImplementedError:
+        return False
+    return True
+
+
+# Real per-row screen reader support (wx.Accessible) only works on
+# Windows builds of wxWidgets/wxPython; see _accessible_supported().
+HAS_ACCESSIBLE = _accessible_supported()
+
+
+def _acc_state(*names):
+    """OR together wx.ACC_STATE_SYSTEM_<NAME> flags looked up by name.
+
+    Looked up defensively (rather than hardcoded as integers) since
+    only a handful of these are documented for wxPython and the exact
+    set available can vary by version; a name that isn't present in
+    this build is simply skipped instead of raising.
+    """
+    value = 0
+    for name in names:
+        value |= getattr(wx, 'ACC_STATE_SYSTEM_' + name, 0)
+    return value
+
+
+def _notify(event_name, window, child_id):
+    """Tell the OS a virtual child's focus/selection/value changed.
+
+    This is the piece that makes screen readers re-query our
+    wx.Accessible object on their own, instead of us pushing speech
+    ourselves. No-op if wx.Accessible isn't available on this platform
+    or this wxPython build doesn't expose the named event constant.
+    """
+    if not HAS_ACCESSIBLE:
+        return
+    event_type = getattr(wx, event_name, None)
+    if event_type is None:
+        LOG.debug('wx.%s not available in this wxPython build', event_name)
+        return
+    try:
+        wx.Accessible.NotifyEvent(event_type, window,
+                                  wx.OBJID_CLIENT, child_id)
+    except Exception:
+        LOG.debug('NotifyEvent failed', exc_info=True)
+
+
+def _notify_focus(window, child_id):
+    _notify('ACC_EVENT_OBJECT_FOCUS', window, child_id)
+    _notify('ACC_EVENT_OBJECT_SELECTION', window, child_id)
+
+
+def _notify_value_change(window, child_id):
+    _notify('ACC_EVENT_OBJECT_VALUECHANGE', window, child_id)
+
+
+class GridAccessible(wx.Accessible):
+    """wx.Accessible for a wx.grid.Grid's grid window.
+
+    Exposes one virtual child per cell (1-based childId, row-major, as
+    IAccessible requires) whose name is built from the column header
+    and current cell text, so a screen reader reads it the same way it
+    would any other table/list control.
+    """
+
+    def __init__(self, grid):
+        super().__init__()
+        self.grid = grid  # the wx.grid.Grid, not GetGridWindow()
+
+    def _row_col(self, child_id):
+        cols = self.grid.GetNumberCols()
+        if child_id < 1 or cols <= 0:
+            return None
+        row, col = divmod(child_id - 1, cols)
+        if row >= self.grid.GetNumberRows():
+            return None
+        return row, col
+
+    def GetChildCount(self):
+        return (wx.ACC_OK,
+                self.grid.GetNumberRows() * self.grid.GetNumberCols())
+
+    def GetChild(self, childId):
+        if childId == 0:
+            return wx.ACC_OK, self
+        # A plain int childId with no wx.Accessible of its own is a
+        # "simple element" per the wx.Accessible docs.
+        return wx.ACC_OK, None
+
+    def GetName(self, childId):
+        if childId == 0:
+            return wx.ACC_OK, self.grid.GetGridWindow().GetName()
+        cell = self._row_col(childId)
+        if not cell:
+            return wx.ACC_FAIL, ''
+        row, col = cell
+        header = self.grid.GetColLabelValue(col)
+        value = self.grid.GetCellValue(row, col) or _('empty')
+        return wx.ACC_OK, _('%s %s, row %d') % (header, value, row + 1)
+
+    def GetValue(self, childId):
+        cell = self._row_col(childId)
+        if not cell:
+            return wx.ACC_OK, ''
+        row, col = cell
+        return wx.ACC_OK, self.grid.GetCellValue(row, col)
+
+    def GetRole(self, childId):
+        if childId == 0:
+            return wx.ACC_OK, wx.ROLE_SYSTEM_TABLE
+        return wx.ACC_OK, wx.ROLE_SYSTEM_CELL
+
+    def GetState(self, childId):
+        state = _acc_state('FOCUSABLE', 'SELECTABLE')
+        if childId == 0:
+            return wx.ACC_OK, state
+        cell = self._row_col(childId)
+        if not cell:
+            return wx.ACC_FAIL, 0
+        if cell == (self.grid.GetGridCursorRow(),
+                    self.grid.GetGridCursorCol()):
+            state |= _acc_state('FOCUSED', 'SELECTED')
+        return wx.ACC_OK, state
+
+    def GetFocus(self):
+        row = self.grid.GetGridCursorRow()
+        col = self.grid.GetGridCursorCol()
+        cols = self.grid.GetNumberCols()
+        if row < 0 or col < 0 or cols <= 0:
+            return wx.ACC_OK, self
+        return wx.ACC_OK, row * cols + col + 1
+
+    def GetLocation(self, elementId):
+        window = self.grid.GetGridWindow()
+        if elementId == 0:
+            return wx.ACC_OK, window.GetScreenRect()
+        cell = self._row_col(elementId)
+        if not cell:
+            return wx.ACC_FAIL, wx.Rect()
+        row, col = cell
+        rect = self.grid.CellToRect(row, col)
+        rect.SetPosition(window.ClientToScreen(rect.GetTopLeft()))
+        return wx.ACC_OK, rect
+
+
+def attach_grid_accessible(grid, name=None):
+    """Give a wx.grid.Grid a real per-cell accessible tree.
+
+    Binds EVT_GRID_SELECT_CELL to notify Windows of the new focused
+    cell so screen readers announce cell navigation on their own --
+    no speech is pushed by this code.
+    """
+    window = grid.GetGridWindow()
+    if name:
+        window.SetName(name)
+    if not HAS_ACCESSIBLE:
+        return None
+
+    try:
+        accessible = GridAccessible(grid)
+    except NotImplementedError:
+        # Belt-and-suspenders: _accessible_supported() already probed
+        # this at import time, but if it's ever wrong for some build
+        # we still shouldn't crash grid construction over it.
+        return None
+    window.SetAccessible(accessible)
+
+    def _on_select_cell(event):
+        event.Skip()
+        cols = grid.GetNumberCols()
+        if cols <= 0:
+            return
+        child_id = event.GetRow() * cols + event.GetCol() + 1
+        _notify_focus(window, child_id)
+
+    grid.Bind(wx.grid.EVT_GRID_SELECT_CELL, _on_select_cell)
+    return accessible
+
+
+class PropertyGridAccessible(wx.Accessible):
+    """wx.Accessible for a wx.propgrid.PropertyGrid.
+
+    Exposes one virtual child per visible row (1-based childId, in
+    display order) so screen readers get a real name/value/role/state
+    for each property instead of the single opaque pane
+    wx.propgrid.PropertyGrid otherwise reports.
+
+    speech_fn(prop) -> str, if given, overrides the default
+    "<label> <value>" rendering (e.g. to add type/locked/unspecified
+    wording); it is only used for GetName(), the accessible name is
+    what screen readers read on focus.
+    """
+
+    def __init__(self, pg, speech_fn=None):
+        super().__init__()
+        self.pg = pg
+        self.speech_fn = speech_fn
+
+    def _visible_props(self):
+        it = self.pg.GetIterator()
+        props = []
+        while not it.AtEnd():
+            props.append(it.GetProperty())
+            it.Next()
+        return props
+
+    def _prop_for_child(self, childId):
+        if childId < 1:
+            return None
+        props = self._visible_props()
+        if childId > len(props):
+            return None
+        return props[childId - 1]
+
+    def GetChildCount(self):
+        return wx.ACC_OK, len(self._visible_props())
+
+    def GetChild(self, childId):
+        if childId == 0:
+            return wx.ACC_OK, self
+        return wx.ACC_OK, None
+
+    def GetName(self, childId):
+        if childId == 0:
+            return wx.ACC_OK, self.pg.GetName()
+        prop = self._prop_for_child(childId)
+        if prop is None:
+            return wx.ACC_FAIL, ''
+        if self.speech_fn:
+            return wx.ACC_OK, self.speech_fn(prop)
+        if prop.IsCategory():
+            return wx.ACC_OK, _('Group: %s') % prop.GetLabel()
+        label = prop.GetLabel() or prop.GetName()
+        return wx.ACC_OK, '%s %s' % (label, prop.GetValueAsString())
+
+    def GetValue(self, childId):
+        prop = self._prop_for_child(childId)
+        if prop is None or prop.IsCategory():
+            return wx.ACC_OK, ''
+        return wx.ACC_OK, prop.GetValueAsString()
+
+    def GetRole(self, childId):
+        if childId == 0:
+            return wx.ACC_OK, wx.ROLE_SYSTEM_LIST
+        prop = self._prop_for_child(childId)
+        if prop is not None and prop.IsCategory():
+            return wx.ACC_OK, wx.ROLE_SYSTEM_GROUPING
+        if isinstance(prop, wx.propgrid.BoolProperty):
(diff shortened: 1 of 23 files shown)


chirp/wxui/common.py
```diff
@@ -39,6 +39,88 @@
...
+# unaffected. This is shared by memedit.py (memory grid) and the
+# wx.propgrid.PropertyGrid-based editors below (settings, radio info,
+# memory "Other"/"Advanced"/"Work Mode" tabs), since neither widget
+# exposes a usable accessibility tree on its own.
+try:
+    import prism as _prism
+    _prism_ctx = _prism.core.Context()
+    _sr = _prism_ctx.acquire_best()
+except Exception:
+    _sr = None
```
kk7ds: Can you put this in a separate module instead of `wxui.common`? I think it'd be best if we have this kept apart from everything else, unless there's a reason. Perhaps `wxui.accessibility`?

Also, is this prism library installable from pypi? In order to accept this, it needs to be integrate-able into the builds we provide for download and not just something that works if you install from source. Our builders are kinda old by design in order to support as old a system as possible, so I need to go check if/that it's installable and/or what I need to do to make it so. In other words, what needs to go into `requirements.txt`?

If it's easy and available, I will probably want to remove this defensive import. If not, we should log something when this is not available so we know from debug information if/when it's not.

kk7ds (commented):
Thanks for doing this. I'll be honest, this is a topic that I have avoided for a long time, mostly because the people that want (no, DEMAND) it are extremely intolerant of answers like "I don't have the time or desire to work on this". It has been a touchy subject for years to the point that I'm almost loathe to accept this because I'll have to maintain it going forward. That said, it's pretty understandable as it is, so if it really works and the libraries required remain available, it seems like a win.

My ask for putting it into a separate module are partially because I think it makes sense and partially because I want to make sure it's easily rip-out-able if it falls into disrepair.

kk7ds:
Okay, installed `prismatoid` locally on my mac dev environment. The voiceover "works" but I'm pretty sure not properly. There's a loud voice that already announces a lot of the regular UI elements (the tab, the memory/settings tab group, etc) and a second quieter voice that announces your new stuff. The primary voice is loud, the second voice is soft, and they talk all over the top of each other. If you wait for the main voice to stop and just move around in the memory grid, you can get some useful navigation help. But it's quite a struggle when they start talking over each other. Is that how it's supposed to be?

denizsincar29:
@kk7ds Hello! I'll answer to your both comments at once.
First, separate module does it harder to maintain, it will need to be paralelly developed.
Second, I will try not to rely on prism to make accessibility because WX is already a native gui and only grid and some other widgets are non-native and require something different.
My other solution will be to try to implement wx.accessible class on top of the grid.
Regarding MacOS, it has a very different style of keyboard navigation, so my implementation won't just work.
My suggestion will be to slightly adjust a few elements to be natively accessible by the OS and not do a dll for a screenreader.
I'll try to do this in a few days.
I use windows and linux.
Thanks for understanding.

denizsincar29:
@kk7ds Thanks for being honest about it. I can understand why this has become a sensitive topic over the years, and I appreciate that you're still willing to consider merging it.

However, I think making accessibility a separate module may actually increase the maintenance burden rather than reduce it.

An accessibility fork or separate implementation effectively creates two applications to maintain. Any future UI change would need to be implemented twice, and the two versions would inevitably drift apart over time.

The goal of these changes is not to introduce an alternative UI, but to make the existing one expose information in a way that screen readers can consume. For sighted users, nothing changes.

Keeping accessibility integrated also avoids user confusion ("which CHIRP should I install?") and reduces the chance that accessibility breaks because a new feature was added to the main UI but not to the accessibility module.

Ironically, making accessibility a separate module may increase the risk of it falling into disrepair, because it becomes something that can be forgotten, skipped, or left unsynchronized with the main UI.

From a maintenance perspective, I believe a small amount of accessibility code living alongside the existing UI is actually less work than maintaining a separate accessibility implementation.

I'm also happy to keep the changes isolated, well documented, and as non-invasive as possible so they're easy to understand and maintain in the future.

kk7ds:
> First, separate module does it harder to maintain, it will need to be paralelly developed.

I'm not talking about something separate from this repo, I mean a different python module (file) in the repo. So `chirp/wxui/accessibility.py` instead of putting it inside `chirp/wxui/common.py`.

> Second, I will try not to rely on prism to make accessibility because WX is already a native gui and only grid and some other widgets are non-native and require something different. My other solution will be to try to implement wx.accessible class on top of the grid.

Sorry, I'm not sure I understand. You're planning to re-do this without prism? Why?

The double voicing I'm hearing sounds to me like your implementation is competing with the base implementation of the wx widgets that support accessible interfaces already. That may mean that the prism-based approach is always going to compete, so it may not be the appropriate way, but implementing the actual wx.accessible interfaces may be better. So if that's your plan, hopefully it will resolve the issue.

> Regarding MacOS, it has a very different style of keyboard navigation, so my implementation won't just work. My suggestion will be to slightly adjust a few elements to be natively accessible by the OS and not do a dll for a screenreader. I'll try to do this in a few days. I use windows and linux. Thanks for understanding.

I don't understand this. The issue I'm describing has nothing to do with keyboard navigation on macos - the same thing (double voicing) happens with mouse navigation. Either way, CHIRP is a cross-platform app, so anything that is merged here needs to work for all three platforms, which I'm sure is possible.

kk7ds:
> However, I think making accessibility a separate module may actually increase the maintenance burden rather than reduce it.
> 
> An accessibility fork or separate implementation effectively creates two applications to maintain. Any future UI change would need to be implemented twice, and the two versions would inevitably drift apart over time.

You're missing my point. A module in python is just a file, here in this repo.

> The goal of these changes is not to introduce an alternative UI, but to make the existing one expose information in a way that screen readers can consume. For sighted users, nothing changes.

Yes, I understand. The reason I'm asking for things that affect maintainability, libraries related to the builds we do, and for it to work on all three platforms is specifically for integration here. I'm NOT talking about a fork or alternative/second install.

> Ironically, making accessibility a separate module may increase the risk of it falling into disrepair, because it becomes something that can be forgotten, skipped, or left unsynchronized with the main UI.
> 
> From a maintenance perspective, I believe a small amount of accessibility code living alongside the existing UI is actually less work than maintaining a separate accessibility implementation.
> 
> I'm also happy to keep the changes isolated, well documented, and as non-invasive as possible so they're easy to understand and maintain in the future.

All I'm looking for is for you to move your changes from `wxui/common.py` to `wxui/accessibility.py`. It doesn't change anything other than the organization of where these routines live in the tree.

kk7ds:
Another comment here - with this patch applied the narration voice happens _all the time_ on macos, not just when enabled at the system level. That would obviously be a problem as it forces all the sighted users to hear constant narration during navigation (mouse or keyboard). Moving to a proper `wx.Accessible` mechanism should solve that as well.

denizsincar29:
Pushed an update addressing both points from the review:

- Moved all of the accessibility plumbing into a new `chirp/wxui/accessibility.py` module, out of `common.py`.
- Replaced the prismatoid-based speech push with a proper `wx.Accessible` implementation (`GridAccessible` for the memory grid, `PropertyGridAccessible` for the Settings/Other/Advanced/Work Mode and Radio Info PropertyGrids). These answer standard IAccessible name/value/role/state/focus queries per cell/row and call `wx.Accessible.NotifyEvent()` when the focused cell/row changes, so the screen reader re-queries the OS accessibility layer itself instead of us pushing text to it. No screen-reader-specific package is involved anywhere — `wx.Accessible`/`NotifyEvent` are both part of wxPython.
- `wx.Accessible` is MSW-only per the wxWidgets docs, so on Linux/macOS this now sets a plain accessible name and does nothing further — no competing/forced narration on those platforms at all, which should resolve the macOS behavior you saw.
- Dropped the F1-hijack for reading a setting's description; it's now exposed via `PGProperty.SetHelpString()` / `wx.Accessible.GetDescription()`, reachable through the screen reader's own "read description" command instead of a custom keybinding.
- Rebased onto current master and regenerated locale files.

I've tested this manually with NVDA on Windows (grid, settings grid incl. live combo-box/checkbox edits, radio info grids) and it announces correctly, in NVDA's own voice/phrasing rather than a second competing one. Let me know if you'd like anything else adjusted.

denizsincar29:
CI is green now (fixed two things: mypy rejected the `wx.Accessible if HAS_ACCESSIBLE else object` conditional as a base class expression — that's just a hard mypy limitation on dynamic base classes, silenced with a targeted `# type: ignore[misc]` on that line — and the unit tests' `sys.modules['wx'] = MagicMock()` shim broke on the module-level `import wx.grid` / `import wx.propgrid` I'd added, since those bypass the mock's `sys.modules` entries entirely; switched to referencing `wx.grid.*`/`wx.propgrid.*` as plain attributes at call time instead, matching how the rest of `chirp/wxui/` already does it).

On macOS/Linux: confirmed against the wxWidgets docs and a wxGTK maintainer's own comment on the wx forums — `wxAccessible`/`wx.Accessible` really is Windows-only (`wxUSE_ACCESSIBILITY` is only implemented for the MSW port; enabling it on other ports doesn't even build). GTK's native widgets get accessibility for free from GTK/ATK itself, but that doesn't help here since `wx.grid.Grid`/`wx.propgrid.PropertyGrid` paint their own rows inside one widget rather than using a real child widget per row — the same reason this was silent on Windows without `wx.Accessible` is also why GTK's native support doesn't see individual rows either.

So right now this PR is Windows-only by construction: `accessibility.py` checks `hasattr(wx, 'Accessible')` and no-ops (just sets a plain accessible name) everywhere it's False, which is what actually happens on Linux/macOS. No regression there, but no support either — arrow-key navigation through these grids stays silent for Orca/VoiceOver users, same as before this PR (and before prismatoid).

A real fix for Linux would mean talking to AT-SPI2 directly (there's no `NotifyEvent()`-equivalent shortcut; it'd mean implementing the AT-SPI `Accessible`/`Table`/`Value` D-Bus interfaces via `pyatspi`/`Atspi` GI bindings and registering a service, which is a project of its own, not a small addition to this one). macOS would mean `NSAccessibility` via PyObjC, similarly nontrivial. I don't think either belongs in this PR. Happy to open follow-up issues for both once this lands, if that's useful.

chirp/wxui/accessibility.py
```diff
@@ -0,0 +1,376 @@
...
+
+import logging
+
+import wx
+
+LOG = logging.getLogger(__name__)
+
+# wx.Accessible (and thus real per-row screen reader support) only
+# exists on Windows builds of wxWidgets/wxPython.
+HAS_ACCESSIBLE = hasattr(wx, 'Accessible')
```
kk7ds: This does not appear to be a legit way to determine if we can use this class:
```
>>> import platform
>>> platform.platform()
'macOS-26.5.2-arm64-arm-64bit'
>>> import wx
>>> wx.Accessible
<class 'wx._core.Accessible'>
```

chirp/wxui/accessibility.py
```diff
@@ -0,0 +1,376 @@
...
+    _notify('ACC_EVENT_OBJECT_FOCUS', window, child_id)
+    _notify('ACC_EVENT_OBJECT_SELECTION', window, child_id)
+
+
+def _notify_value_change(window, child_id):
+    _notify('ACC_EVENT_OBJECT_VALUECHANGE', window, child_id)
+
+
+class GridAccessible(
+        wx.Accessible if HAS_ACCESSIBLE else object):  # type: ignore[misc]
```
kk7ds: Please don't make this class' parent dynamic, that's really obscure and confusing. Also, because of the above, this succeeds on other platforms which defeats the point.

chirp/wxui/accessibility.py
```diff
@@ -0,0 +1,376 @@
...
+    """wx.Accessible for a wx.grid.Grid's grid window.
+
+    Exposes one virtual child per cell (1-based childId, row-major, as
+    IAccessible requires) whose name is built from the column header
+    and current cell text, so a screen reader reads it the same way it
+    would any other table/list control.
+    """
+
+    def __init__(self, grid):
+        super().__init__()
```
kk7ds, line 145: Here's where this fails on macOS:
```
ERROR: <function ChirpMain.open_file at 0x1195e93f0> raised unexpected exception
  File "/Users/dan/Documents/chirp.git/chirp/wxui/common.py", line 824, in run_safe
    return fn(*args, **kwargs)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/main.py", line 593, in open_file
    editorset = ChirpEditorSet(radio, filename, self._editors)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/main.py", line 202, in __init__
    edit = self.MEMEDIT_CLS(radio, self._editors)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/memedit.py", line 970, in __init__
    self._grid = ChirpMemoryGrid(self)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/memedit.py", line 120, in __init__
    accessibility.attach_grid_accessible(self, _('Memory List'))
  File "/Users/dan/Documents/chirp.git/chirp/wxui/accessibility.py", line 210, in attach_grid_accessible
    accessible = GridAccessible(grid)
  File "/Users/dan/Documents/chirp.git/chirp/wxui/accessibility.py", line 118, in __init__
    super().__init__()
NotImplementedError
```
So you'll need to do something different here to determine if this is available or not that attempts to actually instantiate `wx.Accessible`

chirp/wxui/memedit.py
```diff
@@ -1153,9 +1165,37 @@ def _keyboard_overrides(self, event):
             # wx.grid.Grid() can do it.
             # https://github.com/wxWidgets/wxWidgets/issues/22625
             self.cb_copy(cut=False)
+        elif event.GetKeyCode() == wx.WXK_WINDOWS_MENU or (
+                event.GetKeyCode() == wx.WXK_F10 and event.ShiftDown()):
```
kk7ds, line 1169: I'm not a windows user - is this a common paradigm? Shift-F10 pops open the right-click menu? I'm surprised this is not hard-wired into wx on MSW if so.

chirp/wxui/accessibility.py
```diff
@@ -0,0 +1,376 @@
...
+    cell so screen readers announce cell navigation on their own --
+    no speech is pushed by this code.
+    """
+    window = grid.GetGridWindow()
+    if name:
+        window.SetName(name)
+    if not HAS_ACCESSIBLE:
+        return None
+
+    accessible = GridAccessible(grid)
```
kk7ds: I think we can just do:
```
try:
    accessible = GridAccessible(grid)
except NotImplementedError:
    return
```
here to solve the above detection problem. Then just make sure any of the things that we pass accessible into from the main classes can tolerate accessible being None and make it a no-op.

I would also like a way to disable this stuff with a config knob for the time being until we prove it out a little more. Basically, if a user starts having problems with the code that is wired into memedit (et al) I can tell them to plop something in their config to disable all this stuff and have it be the equivalent of `HAS_ACCESSIBLE=False`. Once we're sure it's solid we can drop that but I think it'd be a good idea for the time being, especially since this code 100% breaks my ability to even start chirp on macOS right now :)

I can add that knob myself though.

kk7ds (commented):
I tried this last night on Windows and it seems to work pretty well, and is organized much more in the way I'd like, thanks. I will say that on windows the blue highlights that indicate what is being selected/spoken appear completely off of the CHIRP window for some reason, but that may be an artifact of running from source and sharing a parent window with the command line (not a critical issue either way, IMHO).

It's entirely broken on macOS in that CHIRP can't even start. Details inline about why and what to do about it. If it's not clear I can make some changes and push it up for you to have a look at.

Aside from the other comments inline, the issue above and the kill switch knob (which I can implement myself) I'm mostly okay with this now, thanks.

denizsincar29:
Pushed a fix for all the inline points:

- `HAS_ACCESSIBLE` no longer uses `hasattr(wx, 'Accessible')` (true everywhere, meaningless) — it now actually tries `wx.Accessible()` and catches `NotImplementedError`, which is what was crashing startup on macOS.
- Dropped the conditional base class; `GridAccessible`/`PropertyGridAccessible` statically subclass `wx.Accessible`.
- `attach_grid_accessible()`/`attach_propgrid_accessible()` now wrap the actual construction in `try/except NotImplementedError: return None`, exactly as you suggested, so a bad probe on some odd build still degrades to a no-op instead of crashing.
- Added a `disable_accessibility`/`state` config knob to kill all of this at runtime without a rebuild, in case it needs to be turned off while it proves itself out (figured I'd save you the trouble since you'd mentioned adding it).

On the Shift+F10 question: yes, that's the standard Windows keyboard equivalent of a right-click (opens the context menu at the focused item) - it's a system-wide convention most native controls get for free, but `wx.grid.Grid` doesn't wire it up automatically since it paints its own cells, hence the manual bind.

CI is green on the new commit. I don't have a Mac to verify the startup fix myself, so let me know how it behaves there when you get a chance.

kk7ds:
> Pushed a fix for all the inline points:
> 
> * `HAS_ACCESSIBLE` no longer uses `hasattr(wx, 'Accessible')` (true everywhere, meaningless) — it now actually tries `wx.Accessible()` and catches `NotImplementedError`, which is what was crashing startup on macOS.

Yep, works on macOS now and correctly sees it as disabled (and thus does not break things). Confirmed on Linux as well.

> * Dropped the conditional base class; `GridAccessible`/`PropertyGridAccessible` statically subclass `wx.Accessible`.
> * `attach_grid_accessible()`/`attach_propgrid_accessible()` now wrap the actual construction in `try/except NotImplementedError: return None`, exactly as you suggested, so a bad probe on some odd build still degrades to a no-op instead of crashing.
> * Added a `disable_accessibility`/`state` config knob to kill all of this at runtime without a rebuild, in case it needs to be turned off while it proves itself out (figured I'd save you the trouble since you'd mentioned adding it).

This toggle doesn't seem to work for me. If I set it and confirm it's being seen as =true right when you fetch it, I still get all the memedit grid screen reading that should be disabled.

> On the Shift+F10 question: yes, that's the standard Windows keyboard equivalent of a right-click (opens the context menu at the focused item) - it's a system-wide convention most native controls get for free, but `wx.grid.Grid` doesn't wire it up automatically since it paints its own cells, hence the manual bind.

Still surprised they didn't wire that into the grid, but cool, a benefit for even the sighted users.

> CI is green on the new commit. I don't have a Mac to verify the startup fix myself, so let me know how it behaves there when you get a chance.

Yep, aside from the toggle not working (which I can dig into further if need be) the only other thing I notice is that on Windows I always get this barfed out on the console:

`TypeError: invalid result from GridAccessible.GetFocus()`

I haven't looked to see where or why, but we need to figure out what's going on there and fix it.

Also, we need to squash the commits down to probably just one to avoid persisting all the not-yet-right stuff for history. I can do that for you before I merge though so that it happens the way I want and so that users on the mailing list will see it. We should also reference a bug in the main tracker system, but I can add that at the same time as the squashing.

Thanks for your work on this. I know a bunch of people will appreciate it.

kk7ds:
I made the cleanups here, and figured out why the toggle wasn't working. The `GetFocus()` routine still seems to be wrong, as it seems like that always needs to return a `Window` as the second item of the tuple. I will experiment with that soon, but if you have any information on why you did that, please let me know.

I will push the locale fixes in a sec and then will go ahead and merge this for this week's build.

kk7ds:
@denizsincar29 can you create a user account on the main site https://chirpmyradio.com/projects/chirp/wiki/Home so I can cc you on any bugs that are reported for this stuff? Thanks!

douglangley:
Hello, and thanks to both of you for this — @denizsincar29 for the work and
@kk7ds for taking it.

Some context that might save you time, or at least save some duplicated effort.

Back in May I started trying to make CHIRP usable with a screen reader. I took
the other road: rather than patching the existing UI I built a separate wxPython
front end over the CHIRP library (used unmodified, as a dependency), so I could
iterate without touching the main tree. I'm not proposing that as a direction for
CHIRP — you're right that in-tree is the correct place for this. But I had some of the
same problems you did, and I found a few things you may be able to use.

## The grid

`wx.grid.Grid` was the same dead end for me. What I ended up with is a drop-in
accessible grid built on **`wx.dataview.DataViewListCtrl`** rather than
`wx.grid.Grid`:

https://github.com/Community-Access/wx-accessible-grid

It's **MIT licensed**, so there's no licensing obstacle to pulling it in or
lifting code out of it.

**One important trap:** don't `pip install wx-accessible-grid`. PyPI's latest is
0.4.1, which is an old WebView-based version that I've since retired — it is not
what you want and would give you entirely the wrong impression. The current
version is 0.9.0, from git. I'm running it pinned at
`payown/wx-accessible-grid@494103f`, which is 0.9.0 plus the column-locked
cursor described below; that's PR #4 against Community-Access, not yet merged.

### Why the control choice matters for your macOS problem

This is the part I think is actually worth your time. You noted that
`wx.Accessible` is MSW-only, so arrow-key navigation stays silent for VoiceOver
and Orca. That's correct and there's no way around it in that direction.

`DataViewListCtrl` is a different situation, because it wraps a *real native
view* per platform — **`NSTableView` on macOS**, which carries Apple's own
NSAccessibility implementation for free, so VoiceOver reads the table, its rows,
and each cell without anything extra. That is the one route I know of to macOS
grid accessibility in wx.

This isn't theoretical. We've both run it on macOS under VoiceOver — @payown, who
maintains the fork linked above, did the more thorough pass, covering **grid
navigation and channel editing**. So it's been exercised on real hardware, not
just reasoned about.

Worth calling out that editing was part of it, because the two are connected:
because plain Left/Right reach the `NSTableView` and the cell cursor tracks which
column you're on, F2 can edit *the focused cell* rather than falling back to
asking the user which field they meant. That chain working on macOS is the bit I
would have expected to be fragile.

The honest caveat: on Windows `DataViewListCtrl` is wx's *generic*, custom-drawn
control, not a native common control. It's exposed to MSAA well enough that NVDA
reads rows, but it announces no per-cell cursor — which is exactly the gap your
`wx.Accessible` work fills.

So I don't think these are competing approaches. They compose:
`DataViewListCtrl` gets you rows on macOS where `wx.Accessible` can't reach, and
your `GridAccessible` gets you cells on Windows where the generic control is
quiet. If I were doing it in-tree, that's the combination I'd aim for.

I should also flag that the grid as written stores its rows rather than being
virtual. For CHIRP's memory grids you'd likely want `DataViewCtrl` with a custom
model instead — the underlying accessibility story is the same.

### On the double speech you heard

You asked why prism gave you two competing voices on macOS. I think the answer
is what the speech is layered on top of, not prism itself.

If you push speech while the control's own selection is moving, the screen reader
announces the row *and* your speech announces it — two voices, which is what you
heard. In my grid the cell cursor takes over all four arrow keys and
**deliberately does not move the native selection or current item while
arrowing**. It scrolls with `EnsureVisible()` and speaks; the native selection is
only synced at action time (edit, delete, context menu). Because the control's
selection never moves during navigation, the screen reader never fires its
automatic row read, so there is only ever one voice.

That matches what we actually get: neither payown nor I have run into the double
speech you describe — not on macOS under VoiceOver, and not on Windows under
NVDA. Same library, same platform you hit it on. I don't doubt what you heard;
I think it's a consequence of pushing speech over a control whose own selection
is still moving underneath, which is the thing the cell cursor avoids.

The side benefit is verbosity: Up/Down is column-locked, so moving down a column
reads "Channel 47, 146.520" rather than re-reading all fourteen fields each time.
That turned out to matter more day-to-day than I expected.

### On your other objection, which I think was the stronger one

You also said the narration happened all the time on macOS, not only when a
screen reader was enabled. That one I'd take seriously rather than argue with.
prism's `create_best()` picks the highest-priority *available* backend, and below
the screen readers its registry contains plain TTS engines — OneCore and SAPI on
Windows. So unless you gate on a screen reader actually running, it will happily
speak to someone who never asked for it. That's fixable by gating, but you were
right to flag it, and it's a real argument for your `NotifyEvent` design over any
push-speech approach.

For completeness, the other library in this space is Tolk
(https://github.com/dkager/tolk), which can detect whether a screen reader is
present. But it's Windows-only, so it doesn't help the macOS side at all, and
given you've just moved deliberately away from pushing speech I'm not sure it's
a direction you want.

### One place your approach is clearly better than mine

Because you notify through the OS layer, braille follows focus automatically.
Mine doesn't — by suppressing the native selection during arrowing to avoid the
double-read, I've also stopped a braille display from tracking the cursor, and I
don't currently send braille myself. That's a genuine advantage of doing it the
way you have, and it's on my list to fix.

## Where to look, if any of this is useful

The full app is at https://github.com/douglangley/vrp (GPLv3, so no licence
friction with CHIRP). It's incomplete in a lot of places, because I was rebuilding
a UI from scratch rather than covering CHIRP's whole feature set — but the grid
and channel editing are solid. Specific files, in rough order of usefulness:

| Path | Why |
|---|---|
| `docs/research/2026-06-24-native-grid-voiceover-feasibility.md` | Why a report-mode `wx.ListCtrl` is silent under VoiceOver and `DataViewListCtrl` isn't. Probably the most useful thing here even if you take no code. |
| `wx_accessible_grid/grid.py` (in the library) | The cell cursor itself: `_on_key_down`, `_move_row`, `sync_selection_to_cursor`. |
| `vrp/native/channel_grid.py` | How a host wires it up, including Shift+F10. |
| `vrp/native/grid_model.py` | The data/selection model, deliberately free of any wx import so it can be unit-tested headless — which given your `sys.modules['wx'] = MagicMock()` shim might be a useful pattern. |
| `vrp/native/announce.py` | The policy for what gets spoken versus what the screen reader is left to say. |

Please treat all of it as take-what-you-want. I'd much rather this ends up in the
main tree than maintained twice, and I have no interest in adding to your
maintenance burden — I read your note about how this topic has gone in the past,
and that's fair. If none of it fits, that's a perfectly good outcome too.

Happy to answer questions about any of it.

PR #1598, opened 2026-06-23
Add support for RA89R/RA89G model

Tommy83033:
# CHIRP PR Guidelines


kk7ds:
Hi Tommy, I assume you've seen the PR check failure. Looks like this copies a substantial amount of code from the other driver. Are you working on reducing that?

Tommy83033:
> Hi Tommy, I assume you've seen the PR check failure. Looks like this copies a substantial amount of code from the other driver. Are you working on reducing that?

Yes, I’m still refactoring the code. My biggest concern is breaking other models under TH_UV88. The RA89R/RA89G shares most of the logic with them. It only adds encryption/decryption for data transmission and introduces the VFO channel in the channel module.

kk7ds:
Okay then it would be best to be a direct subclass of that driver. Can you point out just the memory map changes that you need for the new driver to start with? Do you just need to add a few structures and everything else is the same? If so, show me what those are and I can help get you started.

Thanks for doing this and thanks for being careful not to break the other driver :)

Tommy83033:
> Okay then it would be best to be a direct subclass of that driver. Can you point out just the memory map changes that you need for the new driver to start with? Do you just need to add a few structures and everything else is the same? If so, show me what those are and I can help get you started.
> 
> Thanks for doing this and thanks for being careful not to break the other driver :)

1. The RA89R driver reuses the `chns` structure in `th_uv88.MEM_FORMAT`. Only three fields—`freqreverse`, `talkaround`, and `tone5pttid`—were added to unused free spaces without altering the original memory layout.

2. Storage space for VFO channels (sharing the same structure as regular memory channels) and custom scan frequency parameters has been added, along with changes to the basic settings.

If possible, I would appreciate it if you could review my current code; it might provide a clearer picture. The communication part has undergone significant changes, and the display of the VFO channels has also been added.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,414 @@
...
+            raise errors.RadioError(msg)
+
+        # UI Update
+        status.cur = addr
+        status.msg = "Cloning to radio..."
+        radio.status_fn(status)
+    th_uv88._exit_program_mode(radio)
+
+
+def _encode_data(byte_data_backup):
```
kk7ds, line 272: Please add a docstring that explains the logic/behavior here here. Some sort of byte stuffing/escaping scheme it looks like?

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,414 @@
...
+                val = (0xF0 + byte_data[index + 1] + 0x80) & 0xFF
+                decoded_data.append(val)
+                index += 2
+            else:
+                decoded_data.append(byte_data[index])
+                index += 1
+    return bytes(decoded_data)
+
+
+def lrc_cal(lrc_init_value: int, auch_msg: bytes) -> bytes:
```
kk7ds: This too. Looks like some sort of checksum?

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,414 @@
...
+        rf = super().get_features()
+        rf.valid_special_chans = sorted(SPECIAL_MEMORIES.keys())
+        return rf
+
+    def sync_in(self):
+        """Download from radio"""
+        try:
+            data = _download(self)
+            # with open('output.bin', 'wb') as f:
+            #  f.write(data)
```
kk7ds: This temporary code should be removed.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,414 @@
...
+                and mem.mode == 'AM'):
+            msgs.append(chirp_common.ValidationWarning(
+                _('Frequency in this range must not be AM mode')))
+        return msgs + super().validate_memory(mem)
+
+
+@directory.register
+class RA89G(RA89R):
+    VENDOR = "Retevis"
+    MODEL = "RA89G"
```
kk7ds: What is the point of this subclass if it doesn't do anything different? Is there code in the base class(es) that treats it differently by model name?

chirp/drivers/th_uv88.py
```diff
@@ -302,7 +308,7 @@ def _rawrecv(radio, amount):
         msg = "Generic error reading data from radio; check your cable."
         raise errors.RadioError(msg)
 
-    if len(data) != amount:
+    if not radio.is_ra89rg() and len(data) != amount:
```
kk7ds: This looks like it is covering up a bug to me. It should always be an error if we read less than we expect because that also means we've taken a timeout.

chirp/drivers/th_uv88.py
```diff
@@ -707,13 +715,17 @@ def _get_memory(self, mem, _mem, _name):
...
+            mem.name = ""
+            for i in range(6):   # 0 - 6
+                mem.name += chr(_mem.name[i])
+            for i in range(10):
+                mem.name += chr(_name.extra_name[i])
 
-        mem.name = mem.name.rstrip()    # remove trailing spaces
+            mem.name = mem.name.rstrip()    # remove trailing spaces
+        else:
+            mem.name = _name
```
kk7ds: In this case `_name` is something like `VFO A` or something right? If so, that's the wrong thing to use here. The `name` field is the user-controllable alpha/text label stored in the radio. Now there are certain radios with hard-coded names where this would make sense, but for a VFO it does not. Set the `extd_number` field to the `VFO A` (or whatever) string and the UI will display it as the index. You should set `name` empty here and set it immutable to indicate that it's not settable.

chirp/drivers/th_uv88.py
```diff
@@ -764,8 +776,36 @@ def _get_memory(self, mem, _mem, _name):
         step = RadioSetting("step", "Step",
                             RadioSettingValueList(LIST_STEPS,
                                                   current_index=_mem.step))
+
```
kk7ds: Please don't add this unrelated blank line.

chirp/drivers/th_uv88.py
```diff
@@ -764,8 +776,36 @@ def _get_memory(self, mem, _mem, _name):
...
+                    TONE5_PTTID_LIST, current_index=tone5_pttid_value))
+            mem.extra.append(tone5_pttid)
+            freqreverse_value = _mem.freqreverse
+            if _mem.talkaround == 1:
+                freqreverse_value = 2
+            mem.extra.append(
+                RadioSetting(
+                    "freqreverse", "Frequency Reverse",
+                    RadioSettingValueList(FREQ_REVERSE_LIST,
+                                          current_index=freqreverse_value)))
```
kk7ds: It would be better if you put this in the subclass' `get_memory()` method instead of adding this special case to the base class. Lots of other drivers do this wrong and it's unfortunate, but here it should be easy to do. After you call `self._get_memory()`, add the extra fields to it before you return it.

chirp/drivers/th_uv88.py
```diff
@@ -782,20 +822,18 @@ def _get_memory(self, mem, _mem, _name):
                     SCRAMBLE_LIST, current_index=scramble_value))
             mem.extra.append(scramble)
 
-        optsig = RadioSetting("signal", "Optional signaling",
-                              RadioSettingValueList(
-                                  OPTSIG_LIST,
-                                  current_index=_mem.signal))
+        optsig = RadioSetting(
+            "signal", "Optional signaling",
+            RadioSettingValueList(OPTSIG_LIST, current_index=_mem.signal))
```
kk7ds: Please don't change this code. as it looks like you're just reformatting it.

chirp/drivers/th_uv88.py
```diff
@@ -782,20 +822,18 @@ def _get_memory(self, mem, _mem, _name):
...
         mem.extra.append(optsig)
-
-        rs = RadioSetting("pttid", "PTT ID",
-                          RadioSettingValueList(PTTID_LIST,
-                                                current_index=_mem.pttid))
-        mem.extra.append(rs)
-
+        if not self.is_ra89rg():
```
kk7ds: This is fine to do because it's not your code, but this is why it's better to do these special cases in the subclasses instead of the base classes. So that the subclasses have the logic to include things instead of having the base class contain all that logic.

chirp/drivers/th_uv88.py
```diff
@@ -864,26 +902,38 @@ def _set_memory(self, mem, _mem, _name):
         elif txmode == "DTCS":
             _mem.txtone = int(str(mem.dtcs), 8)
 
-        _mem.wide = self.MODES.index(mem.mode)
+        if self.is_ra89rg() and mem.mode == "AM":
+            _mem.wide = 2
+        else:
+            _mem.wide = self.MODES.index(mem.mode)
+
```
kk7ds: This too.. please just make your `set_memory()` do the right thing for this after it calls this base class. So something like:

```
self._set_memory(...)
if mem.mode == 'AM':
  _mem.wide = 2
```

(or whatever it should be).

chirp/drivers/th_uv88.py
```diff
@@ -864,26 +902,38 @@ def _set_memory(self, mem, _mem, _name):
...
+            if self.is_ra89rg() and element.get_name() == "freqreverse":
+                val = FREQ_REVERSE_LIST.index(element.value)
+                if val == 2:
+                    _mem.talkaround = 1
+                    _mem.freqreverse = 0
+                else:
+                    _mem.talkaround = 0
+                    _mem.freqreverse = val
+            else:
+                setattr(_mem, element.get_name(), element.value)
```
kk7ds: Same here, handle this in your own `set_memory()`. If you need to make this code not crash for things it doesn't understand, then name the setting something different (like `freqreverse_ta`) and then make this code ignore things it doesn't find:
```
for element in mem.extra:
    try:
        setattr(_mem, element.get_name(), element.value)
    except KeyError:
        pass
```

chirp/drivers/th_uv88.py
```diff
@@ -864,26 +902,38 @@ def _set_memory(self, mem, _mem, _name):
...
-        _workmode = self._memobj.workmodesettings
         _openradioname = self._memobj.openradioname
 
         basic = RadioSettingGroup("basic", "Basic Settings")
         group = RadioSettings(basic)
 
+        is_ra89rg = self.is_ra89rg()
         # Menu 02 - TX Channel Select
-        if self._hasLCD:
+        if self._hasLCD and not is_ra89rg:
```
kk7ds: Just make the subclass set `_hasLCD = False` right? No need to make this model-specific.

chirp/drivers/th_uv88.py
```diff
@@ -911,7 +961,7 @@ def get_settings(self):
 
         # Menu 07 - LED Mode
         if self._hasLCD:
-            if self.MODEL == "RA89":
+            if self.MODEL == "RA89" or is_ra89rg:
```
kk7ds: Change this to:
```
if self.MODEL in ['RA89', 'RA89R', 'RA89G']:
```
to keep it consistent.

chirp/drivers/th_uv88.py
```diff
@@ -959,13 +1009,15 @@ def get_settings(self):
...
 
         # Menu 17 - Scan Type
-        if self.MODEL == "QRZ-1":
-            options = ["Time", "Carrier", "Stop"]
-        else:
-            options = ["TO", "CO", "SE"]
-        rx = RadioSettingValueList(options, current_index=_settings.scanType)
-        rset = RadioSetting("basicsettings.scanType", "Scan Type", rx)
-        basic.append(rset)
+        if not is_ra89rg:
```
kk7ds: This is stacking more types of checking for models that will just get messier. Really what you want here is to not run this for any model that doesn't have a `scanType` right? So change this to:
```
if 'scanType' in _settings:
    ...
```

chirp/drivers/th_uv88.py
```diff
@@ -981,7 +1033,7 @@ def get_settings(self):
 
         # Menu 20 - Intro Screen
         if self._hasLCD:
-            if self.MODEL == "RA89":
+            if self.MODEL == "RA89" or is_ra89rg:
```
kk7ds: Same as above.

chirp/drivers/th_uv88.py
```diff
@@ -1008,6 +1060,62 @@ def get_settings(self):
             rset = RadioSetting("basicsettings.disMode", "Display Mode", rx)
             basic.append(rset)
 
+        if is_ra89rg:
```
kk7ds: Can you change this to be the same style as the others? (i.e. `if self.MODEL in [...]`). I know it's more typing but consistency is better.

chirp/drivers/th_uv88.py
```diff
@@ -1057,7 +1167,7 @@ def _char_to_name(name):
 
             # On model others than RA89 store a truncated name1 into
             # basicsettings
-            if self.MODEL != "RA89":
+            if self.MODEL != "RA89" and not is_ra89rg:
                 rset.set_apply_callback(_name_apply, _settings, "introScreen1",
```
kk7ds: Here too.

chirp/drivers/th_uv88.py
```diff
@@ -1078,19 +1188,22 @@ def _char_to_name(name):
...
-            options.append(item_to_add)
-        # log unknown region codes greater than 4
-        if _settings2.region > 4:
-            LOG.debug("Unknown region code: {value}".
-                      format(value=_settings2.region))
-        rx = RadioSettingValueList(options, current_index=_settings2.region)
-        rx.set_mutable(False)
-        rset = RadioSetting("settings2.region", "Region", rx)
-        advanced.append(rset)
+        if not is_ra89rg:
```
kk7ds: And this.

chirp/drivers/th_uv88.py
```diff
@@ -1099,30 +1212,53 @@ def _char_to_name(name):
                            "Flashlight", "Power Level", "Alarm",
                            "Noise Cancelaton", "Temp Monitor", "FM Radio",
                            "Talk Around", "Frequency Reverse"]
+            elif is_ra89rg:
```
kk7ds: And this.

chirp/drivers/th_uv88.py
```diff
@@ -1099,30 +1212,53 @@ def _char_to_name(name):
...
+                           "Talk Around", "Frequency Reverse"]
             elif self.MODEL in ["P2", "P62"]:
                 options = ["None", "VOX", "ManDown Sw",
                            "Scan", "Moni", "1750 Tone",
                            "Power Level", "Alarm", "Noise Cancelaton",
                            "Temp Monitor", "FM Radio", "Talk Around",
                            "Frequency Reverse"]
+            options_long = (
+                options if not is_ra89rg
+                else options[:8] + ["Temporarily Moni"] + options[8:])
```
kk7ds, line 1187: Why not just put this in `options` above for this model?

chirp/drivers/th_uv88.py
```diff
@@ -1099,30 +1212,53 @@ def _char_to_name(name):
...
+
+            def _side_key_apply(setting, obj, atrb):
+                index = options_long.index(str(setting.value))
+                setattr(obj, atrb, index)
+
+            sidekey1_index = (
```
kk7ds, line 1193: What is this logic doing?

chirp/drivers/th_uv88.py
```diff
@@ -1137,10 +1273,101 @@ def _char_to_name(name):
             rset = RadioSetting("basicsettings.manDownTm", "ManDown Tm", rx)
             advanced.append(rset)
 
-        if self._hasLCD:
+        if is_ra89rg:
```
kk7ds: Also here.

chirp/drivers/th_uv88.py
```diff
@@ -1137,10 +1273,101 @@ def _char_to_name(name):
...
+            _scanfreq = self._memobj.scanfreq
+            scanb = RadioSettingGroup("scandioc", "Scan Settings")
+            group.append(scanb)
+
+            options = ["TO", "CO", "SE"]
```
kk7ds: Looks to me like you're duplicating what you excluded above in the scanType stuff. Why? And why is this not located up near that other code for contextual clarity?

chirp/drivers/th_uv88.py
```diff
@@ -44,12 +44,14 @@
   u8   power:2,
        wide:2,
        b_lock:2,
-       unk3:2;
+       freqreverse:1,
```
kk7ds: Can you comment these new fields as RA89 only?

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,414 @@
...
+    for item in byte_data_backup:
+        data_temp = (item + 0x80) & 0xFF
+        if data_temp > 0xF9:
+            byte_data.extend([0xFF, data_temp & 0x0F])
+        else:
+            byte_data.append(data_temp)
+    return bytes(byte_data)
+
+
+def _decode_data(byte_data):
```
kk7ds, line 292: Here too.

chirp/drivers/th_uv88.py
```diff
@@ -1057,7 +1167,7 @@ def _char_to_name(name):
 
             # On model others than RA89 store a truncated name1 into
             # basicsettings
-            if self.MODEL != "RA89":
+            if self.MODEL != "RA89" and not is_ra89rg:
```
kk7ds: Here too.

chirp/drivers/th_uv88.py
```diff
@@ -1137,10 +1273,101 @@ def _char_to_name(name):
...
+                                       current_index=_settings.btmiclevel)
+            rset = RadioSetting("basicsettings.btmiclevel", "Mic Gain", rx)
+            btb.append(rset)
+
+            rx = RadioSettingValueBoolean(_settings.localspkswitch)
+            rset = RadioSetting("basicsettings.localspkswitch",
+                                "Speak Switch", rx)
+            btb.append(rset)
+
+        if self._hasLCD and not is_ra89rg:
```
kk7ds: Again, why does this not just check `_hasLCD` and the RA89 drivers set it to False?

kk7ds (commented):
Yep, indeed Tommy, the latest version is headed in the right direction, thanks.

chirp/drivers/retevis_ra89r.py
Tommy83033: I compared the official software for these two models, and apart from the model names, I found no other differences.

chirp/drivers/th_uv88.py
Tommy83033: No, it's because RA89R/RA89G uses encrypted communication, and the length of the response bytes is not fixed, so I disabled the verification in this part.

kk7ds: Calling `pipe.read(n)` and getting back less than `n` bytes is not acceptable, whatever the reason and it will always mean that we have taken a timeout. That behavior is not very portable across platforms, so you must arrange for that not to happen.

chirp/drivers/retevis_ra89r.py
kk7ds: Okay, then please add the alias to `chirp/share/model_alias_map.yaml` and it will be added to the website on the next build so that users know that it's supported by the same class as the RA89R.

chirp/drivers/th_uv88.py
Tommy83033: Another solution is to fetch bytes in a loop with a timeout. We stop reading once the packet checksum passes verification. I don’t have any other better ideas for now.

Tommy83033:
> Yep, indeed Tommy, the latest version is headed in the right direction, thanks.

hi, based on some feedback from last time, I've made some optimizations, but I currently don't have a good solution for the time-blocking issue caused by the uncertainty of the download length of walkie-talkie data.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+STIMEOUT = 1
+BAUDRATE = 57600
+
+MEM_FORMAT = """
+#seekto 0x10BC;
+struct chns chan_vfo_mem[4]; // CHAN_NUM
+
+#seekto 0x1F00;
+struct {
+  u8 bitmap[26];    // one bit for each channel marked in use
```
kk7ds: This should probably be `lbit` or `bit`

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+  u8 unk17:1,
+     btappmodeswitch:1,
+     localspkswitch:1,
+     unk18:4,
+     btswitch:1;
+  u8 unk20:5,
+     miclevel:3;
+  u8 unk19;
+  u8 unk1:4,
+     sqlLevel : 4;
```
kk7ds: Please remove the spaces around the colon in these for consistency.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+
+        # Check for framing delimiter
+        if b == FRAME_END:
+            # Validate preamble \xfe\xfe and minimum frame size requirement
+            if (len(package_data) >= 5
+               and package_data.startswith(FRAME_PREAMBLE)):
+                return bytes(package_data)
+
+            # If 0xFD appeared in noise before \xfe\xfe,
+            # clear invalid prefix bytes up to this point
```
kk7ds: This almost always means you've gotten out of sync and can't recover. I would remove this. If you end up with a timeout before `0xFD` or receive `0xFD` before your preamble, raise an error.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+def _rawrecv(radio, max_frame_length=MAX_FRAME_LENGTH, timeout=2.0) -> bytes:
+    """Reads raw serial stream
+    until an Icom-style frame (\xfe\xfe ... \xfd) is received."""
+    package_data = bytearray()
+    start_time = time.time()
+    while (time.time() - start_time) < timeout:
+        chunk = radio.pipe.read(1)
+        if not chunk:
+            # Avoid high CPU utilization on non-blocking ports
+            time.sleep(0.002)
```
kk7ds: This is not necessary. Your `pipe.timeout` is 250ms here if you take the default. Sleeping for 2ms doesn't really make any sense after such a timeout. Also, if 250ms is too short, set it higher and remove the `while (time.time() - start_time) < timeout` stuff - that's being done for you at the serial layer in a platform-specific reliable way.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+    if len(data) != 0:
+        output += data
+    frame += _encode_data(output + lrc_cal(0, output))
+    frame += b"\xFD"
+    # Return the data
+    return frame
+
+
+def _do_start(radio, send_data):
+    th_uv88._rawsend(radio, send_data)
```
kk7ds: This is not a useful thing at all and also marked as private (prefixed with `_`) so please just use `radio.pipe.write()` here. You're already converting unexpected exceptions into `RadioError` in your `sync_in()` and `sync_out()` methods.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+    ack = _rawrecv(radio, 8)
+    if ack != b"\xFE\xFE\xEF\xEE\xE6\x80\x80\xFD":
+        th_uv88._exit_program_mode(radio)
```
kk7ds: This is also not very useful. But also, this is not the place to handle sending this to the radio. Put it in a `finally:` block in `sync_in()` and `sync_out()` so that it always gets sent to the radio, regardless of if we were successful or not.

https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Communication-Errors

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+        # Now we read data
+        """
+         Regular packet size is 1037 bytes.
+         Combined with encryption algorithm and port transmission monitoring,
+         packet length is not fixed.
+         Packet length is calculated as 512*2+13 to prevent data field loss.
+        """
+        d = _rawrecv(radio)
+
+        LOG.debug("Response Data= " + util.hexprint(d))
```
kk7ds: Don't log this in debug log because it's already logged in the trace file:

https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Communication-logging

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+
+        if not d.startswith(b"\xFE\xFE\xEF\xEE\xE4"):
+            LOG.warning("Incorrect start")
+        if not d.endswith(b"\xFD"):
+            LOG.warning("Incorrect end")
```
kk7ds: Both of these should be errors that we raise to stop cloning.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+                val = (0xF0 + byte_data[index + 1] + 0x80) & 0xFF
+                decoded_data.append(val)
+                index += 2
+            else:
+                decoded_data.append(byte_data[index])
+                index += 1
+    return bytes(decoded_data)
+
+
+def lrc_cal(lrc_init_value: int, auch_msg: bytes) -> bytes:
```
kk7ds: `auch_msg` looks like German. I hope you didn't copy this from some other source? Either way, english names please.

chirp/drivers/retevis_ra89r.py
```diff
@@ -0,0 +1,524 @@
...
+            RadioSettingValueList(
+                th_uv88.PTTID_LIST, current_index=tone5_pttid_value))
+        mem.extra.append(tone5_pttid)
+
+        freqreverse_value = _mem.freqreverse
+        if _mem.talkaround == 1:
+            freqreverse_value = 2
+        mem.extra.append(
+            RadioSetting(
+                "freqreverse", "Frequency Reverse",
```
kk7ds, line 442: What does "frequency reverse" mean?

chirp/drivers/th_uv88.py
```diff
@@ -302,7 +308,7 @@ def _rawrecv(radio, amount):
         msg = "Generic error reading data from radio; check your cable."
         raise errors.RadioError(msg)
 
-    if len(data) != amount:
+    if radio.MODEL != "RA89R" and len(data) != amount:
```
kk7ds: Surely this should be removed now?

chirp/drivers/th_uv88.py
kk7ds: You're doing what I said _if_ you need to make this not crash due to KeyError. However, this means there is something in mem.extra that you don't support. What is that and why is it there?

chirp/drivers/th_uv88.py
```diff
@@ -868,22 +880,22 @@ def _set_memory(self, mem, _mem, _name):
...
-        _settings2 = self._memobj.settings2
-        _workmode = self._memobj.workmodesettings
         _openradioname = self._memobj.openradioname
 
         basic = RadioSettingGroup("basic", "Basic Settings")
         group = RadioSettings(basic)
 
         # Menu 02 - TX Channel Select
-        if self._hasLCD:
+        if self._hasLCD and self.MODEL != "RA89":
```
kk7ds, line 896: Why are you doing this? Why not just make your models set `_hasLCD = False` ?

chirp/drivers/th_uv88.py
```diff
@@ -911,7 +923,7 @@ def get_settings(self):
 
         # Menu 07 - LED Mode
         if self._hasLCD:
-            if self.MODEL == "RA89":
+            if self.MODEL in ["RA89", "RA89R"]:
```
kk7ds, line 924: Same

chirp/drivers/th_uv88.py
kk7ds: This comment is still relevant

chirp/drivers/th_uv88.py
Tommy83033: You're right,  I'll clean it up.

chirp/drivers/th_uv88.py
Tommy83033: For the RA89R model under `_hasLCD`, some menu items are supported, hence I did not set it to false. Examples of supported menu items include Dual Wait/Standby, LED Display Mode, Background Light Color, Intro Line 1, Intro Line 2, etc.

chirp/drivers/th_uv88.py
Tommy83033: If possible, I’d like to decouple it from UV88. This way we can eliminate many of these conditional checks and override `get_settings` in RA89R.

chirp/drivers/th_uv88.py
Tommy83033: There are currently no unsupported fields; I have restored this part to its original syntax.

chirp/drivers/retevis_ra89r.py
Tommy83033: Frequency reversal refers to the function of instantaneously switching the transmit and receive frequencies in repeater mode. This naming is based on the official software's naming convention.

kk7ds:
Tommy, I see that you requested review, but there are still a bunch of open comments without resolution or explanations/answers.

chirp/drivers/retevis_ra89r.py
Tommy83033: Thanks for the suggestion. I've added the docstring to clarify the custom byte stuffing and escaping encoding scheme.

chirp/drivers/retevis_ra89r.py
Tommy83033: I've also added comments here

chirp/drivers/retevis_ra89r.py
Tommy83033: I've renamed it to data

chirp/drivers/retevis_ra89r.py
Tommy83033: Thanks for the suggestion; I have changed it to lbit.

chirp/drivers/retevis_ra89r.py
Tommy83033: Thanks for the reminder; I have removed all the spaces before and after the colons.

chirp/drivers/retevis_ra89r.py
Tommy83033: > “频率反转”是什么意思？

 in repeater mode, the transmit frequency differs from the receive frequency. Reverse‑frequency means swapping the transmit and receive frequencies. This naming follows the convention used in their official software.

chirp/drivers/th_uv88.py
Tommy83033: Because long press has one more function than short press: Temporarily Moni

chirp/drivers/th_uv88.py
Tommy83033: The "Side Key Long" function includes an additional "Temporarily Moni" feature compared to the standard "Side Key"; however, since the key values ​​for both correspond to the same "options_long" index, they require separate handling.

Tommy83033:
> Tommy, I see that you requested review, but there are still a bunch of open comments without resolution or explanations/answers.

I am very sorry; it was my oversight! I have just processed and resolved all the remaining discussion threads. Thank you for your patience!

chirp/drivers/retevis_ra89r.py
Tommy83033: Thanks for pointing that out! I've also added a docstring to the lrc_cal function

chirp/drivers/retevis_ra89r.py
Tommy83033: Okay, I have deleted it.

chirp/drivers/retevis_ra89r.py
Tommy83033: Thanks for the suggestion! I've added the alias to chirp/share/model_alias_map.yaml as requested.

chirp/drivers/th_uv88.py
Tommy83033: Thanks for the explanation. I've updated the code to use extd_number for the VFO designation and left name empty and immutable as suggested.

chirp/drivers/th_uv88.py
Tommy83033: Thanks for pointing that out. I've removed.

chirp/drivers/th_uv88.py
Tommy83033: Thanks for the suggestion. I've moved this logic into the subclass's get_memory() method as advised.

chirp/drivers/th_uv88.py
Tommy83033: Okay. I've reverted this change

chirp/drivers/retevis_ra89r.py
Tommy83033: Thank you for the suggestion; I have removed that part of the logic.

chirp/drivers/retevis_ra89r.py
Tommy83033: Thanks for the feedback. I've removed the  sleep and timeout handling, relying on the built-in serial layer timeout instead.

chirp/drivers/retevis_ra89r.py
Tommy83033: Okay, I have removed that private method and used `radio.pipe.write()` directly instead.

chirp/drivers/retevis_ra89r.py
Tommy83033: Okay, I have made the changes as you suggested, placing the logic for exiting program mode within the `finally` blocks of `sync_in()` and `sync_out()` to ensure the command is sent to the radio regardless of whether the operation succeeds or fails.

chirp/drivers/retevis_ra89r.py
Tommy83033: Okay, I've removed that debug log.

chirp/drivers/retevis_ra89r.py
Tommy83033: Okay, I've modified the logging to throw an error.

chirp/drivers/th_uv88.py
Tommy83033: There is nothing unsupported in `mem.extra`; I have placed the special logic within the `set_memory` function of the RA89R.

chirp/drivers/th_uv88.py
Tommy83033: Okay, I've modified this part; I'm reusing the previous version.

chirp/drivers/th_uv88.py
Tommy83033: Thank you for the feedback; I have moved the handling of these special cases to the subclass.

chirp/drivers/th_uv88.py
Tommy83033: Thank you for the feedback; I have moved the handling of these special cases to the subclass.

chirp/drivers/th_uv88.py
Tommy83033: For the RA89R model under _hasLCD, some menu items are supported, hence I did not set it to false. Examples of supported menu items include Dual Wait/Standby, LED Display Mode, Background Light Color, Intro Line 1, Intro Line 2, etc.

chirp/drivers/th_uv88.py
Tommy83033: Thanks for the suggestion. I've updated the condition to check if self.MODEL in ['RA89', 'RA89R', 'RA89G']

chirp/drivers/th_uv88.py
Tommy83033: I’ve reviewed this section again; the `scanType` configuration is indeed present, so I’ve updated the logic accordingly.

chirp/drivers/th_uv88.py
Tommy83033: Got it, I’ve made the changes.

chirp/drivers/th_uv88.py
Tommy83033: Got it, I’ve made the changes.

chirp/drivers/th_uv88.py
Tommy83033: Got it, I’ve made the changes.

chirp/drivers/th_uv88.py
Tommy83033: Got it, I’ve made the changes.

chirp/drivers/th_uv88.py
Tommy83033: Got it, I’ve made the changes.

chirp/drivers/th_uv88.py
Tommy83033: Got it, I’ve made the changes.

chirp/drivers/th_uv88.py
Tommy83033: Got it

chirp/drivers/th_uv88.py
Tommy83033: Got it, I’ve made the changes.

chirp/drivers/th_uv88.py
Tommy83033: For the RA89R model under _hasLCD, some menu items are supported, hence I did not set it to false. Examples of supported menu items include Dual Wait/Standby, LED Display Mode, Background Light Color, Intro Line 1, Intro Line 2, etc.

PR #1600, opened 2026-06-24
bulk_rename: Add Bulk Rename dialog for templated channel renaming

tgies:
Add a Bulk Rename dialog, reachable from the memory editor's right-click menu and Edit menu, that renames selected channels from a template. Templates support Memory-field tokens, some derived convenience tokens such as {freq_khz}, sequence counters with offset/step/format ({seq+10/2:03d}), arithmetic operators, and conditional value maps with negation and nested tokens. A live preview shows results before applying. Renaming skips immutable and empty channels and is blocked on read-only or busy radios.

### DIFF
--- a/chirp/wxui/memedit.py
+++ b/chirp/wxui/memedit.py
@@ -1309,6 +1309,10 @@ class ChirpMemEdit(common.ChirpEditor, common.ChirpSyncEditor):
         goto = common.EditorMenuItem(cls, '_goto', _('Goto...'))
         goto.SetAccel(wx.AcceleratorEntry(wx.MOD_CONTROL, ord('G')))
 
+        bulk_rename = common.EditorMenuItem(
+            cls, '_bulk_rename_from_menu',
+            _('Bulk Rename...'))
+
         expand_extra = common.EditorMenuItemToggle(
             cls, '_set_expand_extra', ('expand_extra', 'state'),
             _('Show extra fields'))
@@ -1328,6 +1332,7 @@ class ChirpMemEdit(common.ChirpEditor, common.ChirpSyncEditor):
                 goto,
                 move_up,
                 move_dn,
+                bulk_rename,
                 ],
             common.EditorMenuItem.MENU_VIEW: [
                 expand_extra,
@@ -2310,6 +2315,19 @@ class ChirpMemEdit(common.ChirpEditor, common.ChirpSyncEditor):
             menu.Append(diff_across_item)
             menu.Enable(diff_across_item.GetId(), len(selected_rows) == 1)
 
+        rename_item = wx.MenuItem(
+            menu, wx.NewId(), _('Bulk Rename...'))
+        self.Bind(
+            wx.EVT_MENU,
+            functools.partial(self._bulk_rename,
+                              selected_rows),
+            rename_item)
+        menu.Append(rename_item)
+        rename_item.Enable(
+            self.editable and not self.busy
+            and used_selected >= 2
+            and self._features.has_name)
+
         self.PopupMenu(menu)
         menu.Destroy()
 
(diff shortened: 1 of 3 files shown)


tgies:
I saw there was a PR for something like this the other day, so I figured I'd clean up the bulk renamer hack I've been using and present it here in sort of an "RfC" capacity.

chirp/wxui/memedit.py
```diff
@@ -1309,6 +1309,10 @@ def get_menu_items(cls):
         goto = common.EditorMenuItem(cls, '_goto', _('Goto...'))
         goto.SetAccel(wx.AcceleratorEntry(wx.MOD_CONTROL, ord('G')))
 
+        bulk_rename = common.EditorMenuItem(
+            cls, '_bulk_rename_from_menu',
+            _('Bulk Rename...'))
```
kk7ds, line 1314: I'm worried that this will lead even more people to file bugs asking for bulk-edit (which already exists) because they see "Bulk rename" and think only rename is editable in bulk. I dunno what to do about that, but I'm just commenting.

Recently a number of those bugs have been raised and I've been thinking about changing the long-standing "Properties" entry in the menu to "Edit %i memories" in hopes that it will be more obvious to people. Putting "Edit" and "Bulk rename" next to each other in the menu stack would probably be a good idea to avoid people seeing one and not the other.

chirp/wxui/memedit.py
```diff
@@ -2331,6 +2349,60 @@ def _mem_properties(self, rows, event):
 
         wx.PostEvent(self, common.EditorChanged(self.GetId()))
 
+    @common.error_proof()
+    def _bulk_rename(self, rows, event):
+        from chirp.wxui.bulk_rename import BulkRenameDialog
```
kk7ds, line 2354: Not sure why this import is here and not at the top, but it should be at the top. And also, please import the module not the thing inside the module.

chirp/wxui/memedit.py
```diff
@@ -2331,6 +2349,60 @@ def _mem_properties(self, rows, event):
...
+
+        if (not self._features.has_name
```
kk7ds, line 2356: Best to either not offer this for `not has_name` or grey it out of the menu I think.

chirp/wxui/memedit.py
```diff
@@ -2331,6 +2349,60 @@ def _mem_properties(self, rows, event):
...
+        rows = [r for r in rows
+                if self._grid.IsRowShown(r)]
+        memories = [
+            self.synchronous_get_memory(self.row2mem(row))
+            for row in rows]
+        memories = [m for m in memories
+                    if not m.empty
+                    and 'name' not in m.immutable]
+
+        if len(memories) < 2:
```
kk7ds, line 2371: Same here, like the other bulk ops, don't offer the option unless it makes sense.

kk7ds (commented):
I saw this come in last night and was thinking (from just the summary in the phone notification) that this was more integrated into the existing bulk-edit box. I also thought it was for pattern-based bulk editing of more than just the name (like for generating lists of memories 25kHz apart, etc), but I was kinda bummed to see this be yet-another bulk edit dialog. This makes more sense to be separate based on how it works and what it's doing, but I also kinda wish it was more integrated into the existing bulk dialog. Any thoughts about extending that to have your "preview" list and/or a pop-out for the tag help buttons, etc? Even if it only supports the formatting for the name, having it all integrated into a single bulk edit dialog would make me a lot happier.

Did you consider using something like jinja for this instead of rolling your own? We also already have a very tiny DSL in `wxui.memquery` which would be nice to either mimic more closely or integrate with a bit...

Either way, this definitely works and is cool so let's definitely do this in some fashion :)

PR #1604, opened 2026-07-02
Improve regexps to avoid some false positives

fthain:
Commit fbcca4a9d1 changed a grep command in an attempt to prevent a false positive, that is, to prevent grep from matching python code not actually using the banned 'past' module. Problem is, (?!e) is not valid syntax unless grep was compiled with support for Perl Compatible Regular Expressions. However, we can use extended regexp syntax to avoid the false positive. While we're at it, let's make similar improvements to the other regular expressions to prevent more false positives.

### DIFF
--- a/tools/check_commit.sh
+++ b/tools/check_commit.sh
@@ -19,7 +19,7 @@ echo
 git diff ${BASE}.. -- '*.py' | grep '^+' > added_lines
 git diff ${BASE}.. -- 'chirp/drivers/*.py' | grep '^+' > driver_lines
 
-if grep -E '(from|import).*\<six\>' added_lines; then
+if grep -E '\<(from|import)\>.*\<six\>' added_lines; then
     fail No new uses of six
 fi
 
@@ -27,7 +27,7 @@ if grep -E '\<six\>' added_lines; then
     fail No new uses of six
 fi
 
-if grep -E '(from|import).*builtins' added_lines; then
+if grep -E '\<(from|import)\>.*builtins' added_lines; then
     fail No new uses of future
 fi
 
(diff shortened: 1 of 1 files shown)


kk7ds:
I'll be honest and say I almost replied to that thread with "please don't". This script is really meant to run in CI, which is always linux, it works, is tested, is just optional advice, etc. I run gnu coreutils on my mac dev machine (and don't really know how anyone can stand not to) so I don't care.

Thanks for doing it, I'll merge, but please let's not obsess over this :)

fthain:
On Thu, 2 Jul 2026, Dan Smith wrote:

> kk7ds left a comment (kk7ds/chirp#1604)
> 
> I'll be honest and say I almost replied to that thread with "please 
> don't". This script is really meant to run in CI, which is always linux, 
> it works, is tested, is just optional advice, etc. I run gnu coreutils 
> on my mac dev machine (and don't really know how anyone can stand not 
> to) so I don't care.
> 

I don't know what Declan uses. PCRE is optional and it happens to be 
turned off on my Linux box.

$ grep -P
grep: Perl matching not supported in a --disable-perl-regexp build
$ grep --version
grep (GNU grep) 3.12

Same for git grep:

$ git grep -P --no-index '\s'
fatal: cannot use Perl-compatible regexes when not compiled with USE_LIBPCRE

If your concern is "don't fix what isn't broken", well, my patch only uses 
the techniques already in use in that script, including "grep -E". I'm 
just using them consistently.

> Thanks for doing it, I'll merge, but please let's not obsess over this 
> :)
> 

I agree that obsessing about these grep commands would be an exercise in 
futility, because regexp is fundamentally the wrong tool for the job, 
because it doesn't comprehend python's grammar. I don't know what tools 
are around that can do proper static analysis of python scripts but I'm 
sure someone somewhere is obsessed with solving that problem properly.

kk7ds:
Yep, it turned from a quick hack into more than that, but it's super simple to maintain, it really only runs in one place, and it checks more than just the python code. I looked at several of the available tools a few years ago and they all had some opinionated base set of things to enforce that I don't like. This is low sophistication but saves a lot of time, unless we obsess over it :)

PR #1606, opened 2026-07-08
kenwood_tone: Add BCD processing for Tone & DCS

McNeight:
Related to issue #11669 and PR #1589

First half of BCD work on kenwood_tone functions.

As part of the 5RH driver work, it was suggested to look at using the
kenwood_tone functions as the data appeared to be similar. A closer
look showed that they were similar, but not enough to be functional.

The existing Kenwood function puts the tone frequency * 10 into a 16-bit
integer, such that 100.0 Hz becomes 1000 which becomes 0x03E8.

For the 5RH, the tone frequency * 10 is stored as a BCD (which is
effectively hexadecimal with six digits ignored). In this case, 100.0 Hz
becomes 1000 which becomes 0x1000.

In the process of testing the code, I modified the h777 driver to use it,
and discovered that those radios also stored the DCS code as BCD. Meaning
that a DCS code of 031, which would be stored as 0x0019 in octal or 0x001F
in decimal, would instead be stored as 0x0031.

The addition of the tone_enc_base variable for processing tone frequencies
as BCD, and the ability to set dcs_enc_base to 16 (effectively processing
as BCD), should widen the pool of potential users for this function, and
cut down on the number of separate tone implemenations and their tests.
The changes made to the h777 driver are an example of this.

With Dan's suggestion, this PR has been split into two parts: implementing
the first half (decoding with _get_tone_val()) to test against existing
code for encoding. The next commit will then implement the encoding half.

All test cases were duplicated using two tone models; model_dec with
dcs_enc_base defaulted to 8 and tone_enc_base defaulted to 10, and
model_bcd with dcs_enc_base and tone_enc_base both set to 16.
Both models allow testing new inputs and outputs, as well as validating
that the new settings do not interfere with parsing when it is not
expected to.

### DIFF
--- a/chirp/kenwood_tone.py
+++ b/chirp/kenwood_tone.py
@@ -24,31 +24,53 @@ class KenwoodToneModel:
     other radio manufacturers.
 
     Typically CTCSS tones are stored as the frequency in Hertz, multiplied by
-    10, and sometimes with a flag bit set (tone_flag).
+    10, and sometimes with a flag bit set (tone_flag). Some radios store it as
+    a base-10 integer. For example:
+    103.5 Hz * 10 becomes 1035, which is then stored as 0x040B.
 
-    DCS codes are stored either as octal or decimal (dcs_enc_base), with one
-    or more flag bits (dcs_base) set to indicate that it's DCS. Reverse
-    polarity is indicated by another flag bit (pol_mask). Finally, a completely
-    disabled tone/code field is often indicated by 0x0000 or 0xFFFF
-    (tone_init).
+    Other radios store this as a binary-coded decimal (BCD) number, where each
+    decimal digit (0–9) is represented by a fixed 4-bit binary sequence. An
+    equivalent way of thinking of it is storing a number in hexadecimal and
+    ignoring digits A-F. For example:
+    103.5 Hz * 10 becomes 1035, which is then stored as 0x1035.
 
-    Note that this requires your radio's memory structure to have u16 fields
-    named exactly "rxtone" and "txtone".
+    The method of storage depends on numerical base (tone_enc_base) used for
+    encoding and decoding.
+
+    DCS codes are stored as either octal, decimal, or BCD (dcs_enc_base).
+    For example, DCS 031 can be stored as 0x0019 octal, 0x001F decimal, or
+    0x0031 BCD / hexadecimal.
+
+    There can be one or more flag bits (dcs_base) set to indicate that it's
+    DCS. Reverse polarity is indicated by another flag bit (pol_mask).
+    Finally, a completely disabled tone/code field is often indicated by
+    0x0000 or 0xFFFF (tone_init).
+
+    Note that this requires your radio's memory structure to have u16 or ul16
+    fields named exactly "rxtone" and "txtone".
 
     :param dcs_base: Base value for DCS tones (e.g., 0x4000, 0x2800, 0x8000)
     :param pol_mask: Bitmask for polarity (e.g., 0x2000, 0x8000, 0x4000)
     :param tone_init: Initial value for uninitialized tones (0x0000, 0xFFFF)
     :param tone_flag: Flag to indicate a CTCSS tone (0x0000 or 0x8000)
-    :param dcs_enc_base: Numerical base for DCS encoding (8 or 10)
+    :param dcs_enc_base: Numerical base for DCS encoding (8, 10, or 16)
+    :param tone_enc_base: Numerical base for Tone encoding (10 or 16)
     """
+
     def __init__(self, dcs_base, pol_mask, tone_init=0x0000, tone_flag=0x8000,
-                 dcs_enc_base=8):
-        self.dcs_base = dcs_base
-        self.pol_mask = pol_mask
-        self.tone_init = tone_init
-        self.tone_flag = tone_flag
-        assert dcs_enc_base in (8, 10), "DCS encoding base must be 8 or 10"
-        self.dcs_enc_base = dcs_enc_base
+                 dcs_enc_base=8, tone_enc_base=10):
+        self.dcs_base = int(dcs_base)
+        self.pol_mask = int(pol_mask)
+        self.tone_init = int(tone_init)
+        self.tone_flag = int(tone_flag)
+        assert dcs_enc_base in (8, 10, 16), (
+            "DCS encoding base must be 8, 10, or 16"
+        )
+        self.dcs_enc_base = int(dcs_enc_base)
+        assert tone_enc_base in (10, 16), (
+            "Tone encoding base must be 10 or 16"
+        )
+        self.tone_enc_base = int(tone_enc_base)
 
     def _get_tone_val(self, tone_val):
         """
@@ -60,15 +82,21 @@ class KenwoodToneModel:
         if tone_val == 0xFFFF or tone_val == 0x0000:
             return None, None
 
+        code = pol = None
         if (tone_val & self.dcs_base) == self.dcs_base:
             if self.dcs_enc_base == 8:
                 code = int("%03o" % (tone_val & 0x07FF))
             elif self.dcs_enc_base == 10:
                 code = int("%03i" % (tone_val & 0x07FF))
+            elif self.dcs_enc_base == 16:
+                code = int("%03x" % (tone_val & 0x07FF))
             pol = (tone_val & self.pol_mask) and "R" or "N"
-            return code, pol
-
-        return (tone_val & 0x7fff) / 10.0, None
+        else:
+            if self.tone_enc_base == 10:
+                code = (tone_val & 0x7FFF) / 10.0
+            elif self.tone_enc_base == 16:
+                code = float("%04x" % (tone_val & 0x7FFF)) / 10.0
+        return code, pol
 
     def _set_tone_val(self, code, pol):
         """
--- a/chirp/kenwood_tone.py
+++ b/chirp/kenwood_tone.py
@@ -114,8 +114,9 @@ class KenwoodToneModel:
             if pol == "R":
                 val += self.pol_mask
             return val
-
-        return int(code * 10) + self.tone_flag
+        else:
+            freq = int(round(code * 10))
+            return int("%i" % freq, self.tone_enc_base) + self.tone_flag
 
     def set_tone(self, mem, _mem):
         """
--- a/tests/unit/test_kenwood_tone.py
+++ b/tests/unit/test_kenwood_tone.py
@@ -46,172 +46,414 @@ class TestKenwoodToneModelInit(base.BaseTest):
 class TestGetToneVal(base.BaseTest):
     def setUp(self):
         super().setUp()
-        self.model_dec = kenwood_tone.KenwoodToneModel(
+        self.model_8_10 = kenwood_tone.KenwoodToneModel(
             dcs_base=0x4000, pol_mask=0x2000)
-        self.model_dcs_bcd = kenwood_tone.KenwoodToneModel(
-            dcs_base=0x4000, pol_mask=0x2000,
-            dcs_enc_base=16, tone_enc_base=10)
-        self.model_tone_bcd = kenwood_tone.KenwoodToneModel(
-            dcs_base=0x4000, pol_mask=0x2000,
-            dcs_enc_base=8, tone_enc_base=16)
-
-    def test_get_tone_val_zero(self):
-        code, pol = self.model_dec._get_tone_val(0x0000)
-        self.assertIsNone(code)
-        self.assertIsNone(pol)
-
-    def test_get_tone_val_zero_dcs_bcd(self):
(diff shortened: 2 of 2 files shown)


chirp/kenwood_tone.py
```diff
@@ -163,11 +198,11 @@ def get_tone(self, _mem, mem):
             mem.tmode = "DTCS"
         elif rxmode or txmode:
             mem.tmode = "Cross"
-            mem.cross_mode = "%s->%s" % (txmode, rxmode)
+            mem.cross_mode = f"{txmode}->{rxmode}"
```
kk7ds: Can you remove all this "gardening" please and just keep the edits to the functional change you're trying to make here? I'm not a fan of f-strings, but I'm really not a fan of non-functional changes mixed with functional ones as I have to eye-diff whether you're changing something or just enforcing your preferences.

chirp/kenwood_tone.py
```diff
@@ -39,100 +41,133 @@ class KenwoodToneModel:
...
+                code = int(f"{(tone_val & 0x07FF):03d}")
+            elif self.dcs_enc_base == 16:
+                code = int(f"{(tone_val & 0x07FF):03x}")
             pol = (tone_val & self.pol_mask) and "R" or "N"
             return code, pol
 
-        return (tone_val & 0x7fff) / 10.0, None
+        if tone_bcd:
+            # hex digits as decimal: 0x1000 -> 1000
+            h = int(f"{(tone_val & 0x7FFF):04x}")
```
kk7ds: Please don't do bit math in f-strings, it's much harder to grok, IMHO.

chirp/kenwood_tone.py
```diff
@@ -39,100 +41,133 @@ class KenwoodToneModel:
...
+            return int(code * 10) + self.tone_flag
 
-    def set_tone(self, mem, _mem):
+    def set_tone(self, mem, _mem, tone_bcd=False):
         """
         Set tone values for a memory channel based on UI settings.
 
         :param mem: Memory channel object from UI
         :param _mem: Radio memory object
+        :param tone_bcd: Tone frequency stored as a 4 digit BCD
```
kk7ds: This is not really a good parameter to pass to set/get_tone().. Set up the behavior in the model and then let these just be getter/setters. Either way, this needs to be more descriptive. It's really "BCD" vs just a direct scalar value, I think you're trying to say... 0x16 vs just 16, but please explain it better for posterity.

kk7ds (commented):
Thanks for this, but I'd ask you to please remove the gardening and clean up the commit history to remove the intermediates. Also it would be ideal if you would change just the code (ideally just the get or set side) in one commit with no test change and then flesh out the tests and the other side in another commit (or commits) so that they can be tested individually to prove that you haven't broken anything here. Changing the set and get code at the same time doesn't prove that this will continue working with existing images and all the other radios that use this and I want to be very careful about that of course.

chirp/kenwood_tone.py
```diff
@@ -57,18 +81,29 @@ def _get_tone_val(self, tone_val):
         :param tone_val: Binary tone value from radio memory
         :return: Tuple of (code, polarity) or (None, None) if no tone
         """
-        if tone_val == 0xFFFF or tone_val == 0x0000:
+
+        assert isinstance(tone_val, (bitwise.u16DataElement,
```
kk7ds: I really don't like the idea of making this have to be a bitwise element instead of just an integer partially because bitwise is (relatively) slow and because this should be generally useful and requiring bitwise makes it less so.

Is there some reason you're making this change, or just trying to tighten the expectations? Unless there's some compelling reason, I'd rather it continue to take "a u16like thing which could be an `int` or a `bitwise.DataElement`".

chirp/drivers/h777.py
```diff
@@ -19,7 +19,7 @@
 import unittest
 import logging
 
-from chirp import chirp_common, directory, memmap
+from chirp import chirp_common, directory, kenwood_tone, memmap
```
kk7ds: It would be good to also split this out into a separate commit just to keep things cleaner, but also because being in the first one leaves this driver broken, since the set side doesn't yet support what it needs.

So ideally, the get side refactor, the set side refactor, and then the "okay and now use this thing for h777". I can do that before I merge if you don't want to.

tests/unit/test_kenwood_tone.py
```diff
@@ -1,6 +1,5 @@
+from chirp import chirp_common, kenwood_tone
```
kk7ds: It's conventional to organize imports into three groups, each sorted alphabetically: stdlib imports, requirements lib imports, then local project imports. CHIRP should mostly be following that pattern (although I see we were missing a blank line between stdlib and project) so in general I'd appreciate it if you'd leave it that way.

tests/unit/test_kenwood_tone.py
```diff
@@ -18,385 +17,2059 @@ def test_init_valid_params(self):
...
-        super().setUp()
-        self.model = kenwood_tone.KenwoodToneModel(
-            dcs_base=0x4000, pol_mask=0x2000)
+    def test_round_trip_tone_16_16(self):
+        mem = chirp_common.Memory()
+        mem.tmode = "Tone"
+        mem.rtone = 100.0
+        result = self._round_trip_16_16(mem)
+        self.assertEqual(result.tmode, "Tone")
+        self.assertEqual(result.rtone, 100.0)
```
kk7ds, line 1784: Thanks for fleshing out all these cases. Might be a good opportunity for using `ddt` to squash this down a bit. I'm not always in favor of that approach as I feel like it can be hard to reason about failures sometimes. But for something where we're trying to test the full matrix of possibilities it probably makes sense. No need to change it now, just something to think about.

kk7ds (commented):
Thanks for splitting, that was mostly what I was looking for. To verify this, I pulled down just the first commit and:

1. Made sure it passed unit tests as-is
2. Patched out the unit test change to make sure that the existing tests worked with just the get side functionally changed
3. Patched out the h777 changes (since that driver is broken on the first commit) and made sure all the driver tests pass

So I'm mostly good with this now, although I'd like to remove the requirement to pass a `bitwise.DataElement` as noted, but I'll wait for your response. I can also fix up the sequencing things if you want.

Note my mention on #1609 that this will probably help over there as well.

Thanks!

chirp/drivers/h777.py
McNeight: I'm going to pull the h777 driver changes out of this branch all together. I've noticed that there is some duplication among 16 channel radio drivers of varying pedigrees, and I think I can merge all of that into a new driver. I'm focused on this, as the 5RM was a bit more than I could chew for a first-time driver. An advantage being that I can test the new driver side-by-side with the old drivers.

kk7ds: There is way too much duplication for sure. Unfortunately I can't de-dupe everything in my head all the time so I have to just let some of it go. If you're doing more than just obvious compaction, we need to get some testers to help, which I can try to round up. Especially for h777, which is _widely_ used... we really need to avoid breaking it.

chirp/kenwood_tone.py
McNeight: Just trying to tighten the expectations. Unfortunately, Pylance reports an error with trying to use the bitwise AND operator between a `bitwise.DataElement` and an `int`. So the options appear to be to leave it as-is or remove it altogether.

I get the feeling I'm trying to solve a problem that doesn't exist, so I'm going to go ahead and remove it.

chirp/drivers/h777.py
McNeight: I had noticed that even though the tests/model_stats.tsv file is three and a half years old, that the h777 driver ranks #2 in popularity.

chirp/kenwood_tone.py
```diff
@@ -57,18 +79,24 @@ def _get_tone_val(self, tone_val):
         :param tone_val: Binary tone value from radio memory
         :return: Tuple of (code, polarity) or (None, None) if no tone
         """
-        if tone_val == 0xFFFF or tone_val == 0x0000:
+        if tone_val == 0xFFFF or tone_val == 0x0000 or tone_val is None:
```
kk7ds: Under what circumstances would we get None here? Perhaps this is one of those places where the external static type checker is confused about python being a duck typed language and requires you to make non-sensical runtime changes to make it happy?

chirp/kenwood_tone.py
```diff
@@ -57,18 +79,24 @@ def _get_tone_val(self, tone_val):
...
                 code = int("%03i" % (tone_val & 0x07FF))
+            elif self.dcs_enc_base == 16:
+                code = int("%03x" % (tone_val & 0x07FF))
             pol = (tone_val & self.pol_mask) and "R" or "N"
-            return code, pol
-
-        return (tone_val & 0x7fff) / 10.0, None
+        else:
+            if self.tone_enc_base == 10:
+                code = float(tone_val & 0x7FFF) / 10.0
```
kk7ds: You don't need this cast to float here (or below) and adding it makes this harder to suss out the exact functional changes you're making.

chirp/kenwood_tone.py
```diff
@@ -78,6 +106,10 @@ def _set_tone_val(self, code, pol):
         :param pol: Polarity string ("N" or "R") or None for CTCSS
         :return: Binary tone value
         """
+
+        assert isinstance(code, (int, float, type(None))
```
kk7ds: Again, I don't think there's a reason to allow None here, but if I'm missing something  please point it out.

chirp/kenwood_tone.py
kk7ds: > I get the feeling I'm trying to solve a problem that doesn't exist, so I'm going to go ahead and remove it.

Yes, thanks. I appreciate that some of the hinting is helpful in an IDE but when it starts requiring runtime changes to make the static checker happy I draw the line.

kk7ds (commented):
I'm going to pull down #1609 and see if I can swap this in for the internal representation.

kk7ds:
Okay yep, works for #1609. So, let's argue about the None thing and then get this merged so I can rebase that on this.

chirp/kenwood_tone.py
McNeight: The `None` checking was less about making the type checker happy, and more about finding and fixing bugs that I encountered as I was testing. My original assumption was that it would compliment the None check in `_set_tone_val()` but I can see how `tone_val` here doesn't have a way of becoming None, unlike `_set_tone_val()`.

chirp/kenwood_tone.py
McNeight: OK. I can remove that, but I believe the parenthesis need to stay to make order of operation clear.

chirp/kenwood_tone.py
McNeight: There is. It's in one of the test cases, specifically `test_set_tone_val_none()`, and the existing code for `_set_tone_val()` in `kenwood_tone` reads:
```python
if code is None:
    return 0x0000
```
The assert is, again, the result of testing and debugging my code and is most likely overkill. But None is an allowable value.

kk7ds: Yep, yep, right.

chirp/kenwood_tone.py
kk7ds: Yeah, they were there before, so good to leave them

chirp/kenwood_tone.py
kk7ds: Roger that, thanks.

tests/unit/test_kenwood_tone.py
```diff
@@ -18,385 +18,2059 @@ def test_init_valid_params(self):
...
-    def test_get_tone_val_ffff(self):
-        code, pol = self.model._get_tone_val(0xFFFF)
+    def test_get_tone_val_ffff_16_16(self):
+        code, pol = self.model_16_16._get_tone_val(0xFFFF)
         self.assertIsNone(code)
         self.assertIsNone(pol)
 
-    def test_get_tone_val_ctcss(self):
-        code, pol = self.model._get_tone_val(885 + 0x8000)
+    def test_get_tone_val_none_8_10(self):
```
McNeight: I'm also deleting these six test cases, since `None` is no longer a valid input for `_get_tone_val()`.

chirp/kenwood_tone.py
McNeight: So... it turns out that I do need one of those float casts (for base16) in order to turn the string conversion back into a float.

My sense of symmetry wants to put back the cast to float on both lines, and slightly modifying the base10 output to match the formatting of the other outputs. It's completely unnecessary at a technical level, but the consistency with the other output lines (setting `code =`) looks better.

That said, "why waste time say lot word when few word do trick!"

kk7ds: Okay yeah, I see why you need the other because you're int->string->float'ing it. But not this one. I appreciate the symmetry but more stuff than necessary is more stuff to confuse what we're doing here IMHO. And following the git history it might make this commit look like it was changing the base-10 logic for reasons related to the rest of the patch.

PR #1610, opened 2026-07-24
docs: Baofeng UV-5R - more detailed descriptions for each radio setting

jasiek:
This is a documentation-only change, but changes driver code via `set_doc(...)`. Based on the english manual I got off the internet, I've written up user friendly descriptions of each setting.


kk7ds:
> This is a documentation-only change, but changes driver code via set_doc(...). Based on the english manual I got off the internet, I've written up user friendly descriptions of each setting.

These are not from the manual. I can tell because they're actually helpful. If they were from any OEM's manual they would be like "DTMF Code: Sets the DTMF Code" :)

Thanks!

PR #1612, opened 2026-07-25
Documentation: Baofeng BF-888S/Retevis H777

jasiek:
Auto-generated documentation for radio settings for the Baofeng BF-888S, based on the English manual.

The strings are **not** marked for translation, per review feedback. Many of them describe common concepts (BCL, VOX, squelch, timeout timer) that appear across a lot of drivers; marking a slightly different wording of each in every driver would generate a large amount of duplicated work for translators. These are better unified into a shared set of strings first, and translated from there. As a result this change no longer touches the locale catalogs at all.

Cross-checked the related Retevis H777 against its own manual. This surfaced one contradiction: the Alarm tooltip claimed the alarm is "receive-only" and "not transmitted", which is wrong for the Baofeng 1900 series (where Alarm is a Local/Remote list and Remote *does* transmit) and is unsupported by the H777 manual, whose Emergency Alert section only describes the side-key toggle. The tooltip is worded to cover both variants.

`tools/check_commit.sh` passes.

### DIFF
--- a/chirp/drivers/h777.py
+++ b/chirp/drivers/h777.py
@@ -449,6 +449,9 @@ class H777Radio(chirp_common.CloneModeRadio):
         mem.extra = RadioSettingGroup("Extra", "extra")
         rs = RadioSetting("bcl", "Busy Channel Lockout",
                           RadioSettingValueBoolean(not _mem.bcl))
+        rs.set_doc('Prevents transmitting on a channel that is already in '
+                   'use. Pressing PTT while the channel is busy emits an '
+                   'alert tone and inhibits the transmitter')
         mem.extra.append(rs)
         if self._has_scramble:
             rs = RadioSetting("beatshift", "Beat Shift(scramble)",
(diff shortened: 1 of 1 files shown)


chirp/drivers/h777.py
```diff
@@ -449,6 +449,9 @@ def get_memory(self, number):
         mem.extra = RadioSettingGroup("Extra", "extra")
         rs = RadioSetting("bcl", "Busy Channel Lockout",
                           RadioSettingValueBoolean(not _mem.bcl))
+        rs.set_doc(_('Prevents transmitting on a channel that is already in '
```
kk7ds: So I didn't comment on this on the other patches, but I'm not sure we really want to mark _all_ of these for translation. If we do and we have a bunch of slightly different strings for common things like this then we'll generate a huge amount of work for the translators.

I would suggest either (a) don't mark these or (b) start a new manifest of strings that describe common/commonly-used things in a separate file, mark those and use those here. BCL is used in a _lot_ of places, as are a lot of the other terms here.

jasiek: Hmmm yeah I think it'd make sense to unify them at some point and *then* use those unified strings as translation keys.

jasiek: Went with (a) for now: dropped the `_()` wrapping from all of these, so the change no longer touches the locale catalogs. Agreed that the right end state is a shared manifest of common strings (BCL, VOX, squelch, TOT, ...) that drivers reference — but that unification should come first, and then those unified strings become the translation keys. Happy to take a swing at that separately if you want.

Repushed as a single commit.

PR #1620, opened 2026-08-03
Add support for QYT KT-A16 airband radio

chrgro:
This is my first contribution to chirp, thanks for a great piece of software!

I have a QYT KT-A16 airband radio. It was a cheap radio, but it works reasonably ok. However programming it for 8.33KHz channel spacing for EU use has been challenging, as the OEM software does not support that, while the radio itself does. But the radio alone cannot do memory channels. Hence my motivation to contribute.

Using extensive amounts of elbow grease and Codex assistance I was able to decode the protocol using wireguard and usbpcap, and then map the memory back to its components using combinatorial trial and error. 

I have validated this works well on my own device. 

image

image

The OEM software and the radio exposes many options that either do little/nothing or should never be used on an airband radio. I have tried to mark these appropriately using help text. 

I am unfamiliar with what the expectation is for the test image you want to accompany this patch. Is that just factory reset the radio, read out everything with chirp, then save the image?

---
Commit message body:

Full support for all settings, including help text.

Supports some settings not exposed in the radio, only via the OEM software.

Also supports several options not relevant for an airband radio but that the radio still seems to expose, either on-device or via OEM software.

---
# CHIRP PR Guidelines


kk7ds:
> have a QYT KT-A16 airband radio. It was a cheap radio, but it works reasonably ok. However programming it for 8.33KHz channel spacing for EU use has been challenging, as the OEM software does not support that, while the radio itself does. But the radio alone cannot do memory channels. Hence my motivation to contribute.

Hmm, what does "cannot do memory channels" mean in this context? You're fetching and returning memories - maybe there's some context missing from the above?

> I am unfamiliar with what the expectation is for the test image you want to accompany this patch. Is that just factory reset the radio, read out everything with chirp, then save the image?

Doesn't need to be a factory-reset image, although that's potentially fine. Ideally it has at least one representative memory store in it that the tests will use for some things. Please do add that to the commit so tests can run before I dig into the code.

Thanks!

chrgro:
> Hmm, what does "cannot do memory channels" mean in this context? You're fetching and returning memories - maybe there's some context missing from the above?

Sorry, this was poorly phrased from my side. What I meant is that you can tune 8.33KHz separated channels using just the radio controls. However you cannot store memory channels using just the radio, you need the software and a programming cable to do that. And that programming software from QYT did not support 8.33KHz. Hence my motivation to contribute :)

I've attached the image now, and updated the unit tests to essentially decode it, make some changes and reencode. 

Should I also do the localization work, or is that something you do in bulk separately?

kk7ds:
Sorry, I missed this.. Yeah I can do the localization stuff for you, just ignore that failure. The driver tests should pass though.

chrgro:
Thanks!

I've been a bit unable to sort out the test failures, so I'd like some advice if you can. Neither me nor AI is making much headway (at least without invasive fixes to the base classes).

The problem is that the QYT KT-A16 derives mode from frequency. If your selected frequency is in the airband range, then you get AM, and if it is in the weather channel range, you get FM. The way I implemented this in the python was to set 
```
  features.has_mode = False
  features.valid_modes = []
```

That works nicely for the driver itself. Tests of the image I provided work fine.

However it fails in your test_edges.py: There you create a generic memory with a frequency just abow the low end of the frequency range, and that memory defaults to FM. And since valid_modes = [], it stays that way. When that generic memory is passed through the QYT driver I wrote, you get AM back (because this happens to be within the airband range, and I have this function: 

```
def _mode_for_frequency(frequency):
    if AIR_BAND[0] <= frequency < AIR_BAND[1]:
        return "AM"
    if WEATHER_BAND[0] <= frequency < WEATHER_BAND[1]:
        return "FM"
    return None
```

And thus the test fails, because the QYT driver has not decoded to the same thing you started out with.

Ideas for how to get around this?

kk7ds:
Top of the driver notes page, should clear it up:

https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Implied-modes

chrgro:
Ahh, thanks for the pointer! I've changed the PR accordingly, do have a look.

PR #1625, opened 2026-08-14
baofeng_5rh_gps: Add Baofeng 5RH Pro and UV-5RM Plus with GPS

steinbrueckri:
This builds on @deeja's draft PR #1589. The driver, the clone protocol and
the memory format are his work, and his commits are preserved here with his
authorship. What I add on top are fixes found while testing that driver
against my own radio, a test image, and a cleanup pass.

Related to #11834 (which also collects #12066, #11836 and #12065 as
duplicates).

### Fixes on top of #1589

* `valid_tuning_steps` was never set, so the `COMMON_TUNING_STEPS` default
  applied and any 6.25 kHz channel was rejected with "Unable to find a
  supported tuning step for 446.006250". The OEM CPS accepts PMR446 fine.
* `_freq_bcd_to_hz()` went through `float("512.00375") * 1e6` and truncated,
  decoding 7633 of 385600 tested frequencies 1 Hz too low, notably all of
  510-535 MHz. It now uses integer math, mirroring the encode direction.
* Bandwidth read bits 4 and 5 of `flags1` while only bit 5 carries it.
* "Scan Add" is a bitmap at 0x81A0 that was not read or written at all, so
  the setting was lost on upload.
* `sqtype` (`flags2[3:0]`) was never written. The CPS sets it whenever a
  receive tone is present, otherwise the squelch does not open on it.
* `offsetdir` (`flags1[3:2]`) was never written either.
* `valid_cross_modes` lacked `DTCS->`, so a TX-only DCS channel — which the
  CPS creates without complaint — could not be edited in chirp.
* `has_tuning_step` claimed a per-channel step that neither the channel
  struct nor the CPS channel editor has, and `can_odd_split` was left False
  although RX and TX are two independent fields.
* `set_memory()` zero-filled the whole 48 byte slot, silently resetting the
  per-channel settings this driver does not decode (talkaround, optional
  signaling, the DTMF/2-Tone/5-Tone/MDC indexes, emergency system). It now
  clears a slot only when it was previously unused.
* `valid_bands` did not match the hardware. The radio reports its own ranges
  in the image header, so `get_features()` reads them from there and keeps a
  hardcoded list only as a fallback.
* `_rebuild_zones()` read `_memobj` before `process_mmap()` had built it,
  turning the clone error paths into an `AttributeError` instead of a
  `RadioError`.

### Added

* **Airband receive and AM.** The image header lists only the four main
  bands, so reading it as complete coverage rejected airband memories. AM is
  not in `flags1` — a 133.450 memory entered on the radio itself keeps
  `flags1` 0xA0 like a plain wide FM channel and is marked by channel byte
  24 instead, which is 0x02 there and 0x00 on all regular VHF/UHF channels.
  A channel chirp writes for the airband now comes out byte-identical to one
  the radio wrote itself, and airband traffic is audible on the radio.
* **UV-5RM Plus GPS alias.** Same hardware and protocol, but it was not
  listed anywhere, so its owners had to guess that the "5RH Pro with GPS"
  entry is the right one — #11834 shows people working through several
  profiles without success.
* **Beta prompt** stating that only v2 firmware (`v2_0_09`) was tested, plus
  the usual download/upload instructions.
* **`valid_characters`** was left at the upper-case default, so a name like
  "Relais Ost" was rejected even though names are stored as GB2312. Now the
  same set as the non-GPS sibling in `baofeng_uv17Pro.py`.

### Deliberately not implemented

"Launch banned" (receive only) lives in bit 5 of channel byte 19 — comparing
a CPS image before and after enabling it shows that byte going 0x00 -> 0x20
on exactly the five banned channels. It is **not** exposed as duplex "off",
because the firmware does not act on it: a channel uploaded with the bit set
still keys up, before and after a power cycle, and the radio has no matching
entry in its channel menu. Where the bit lives is recorded in a comment so
the next person does not have to repeat the experiment.

### Testing

Hardware: Baofeng UV-5RM Plus GPS, software `v2_0_09`, hardware `v1_0_4`,
CH340 cable, on macOS. @deeja independently confirmed the branch on a 5RH
Pro with GPS, including a push/pull comparison against the test image.

* Decoding verified byte for byte against ten images — five written by the
  OEM CPS, five read from hardware — covering 6.25 kHz channels, both
  bandwidths, both power levels, CTCSS and DCS in either direction and with
  either polarity, split channels in both directions, airband/AM channels,
  names, gaps in the channel range, and scan flags. Every channel validates
  and a get/set roundtrip is byte-identical.
* Read, upload unchanged, read back: not a single byte outside the zone
  table changes.
* A set of channels programmed in chirp (16 PMR446 channels on the 6.25 kHz
  grid, a TSQL channel, deliberate gaps) came back from the OEM CPS exactly
  as sent, and the rebuilt zone table matches what the CPS reads back.
* Airband channels written by chirp receive AM correctly on the radio.
* Driver test suite passes (35 passed, 7 skipped), `tools/check_commit.sh`
  is clean.

**Not verified**: boot image upload and writing the general settings are
from the original PR and were not exercised. Only v2 firmware was tested;
older firmware uses a different clone protocol.

### Disclosure

I used an AI assistant (Claude) for the protocol analysis and to write these
changes. Every finding is backed by a byte-level comparison against images
from the OEM CPS, and the result was tested on my own radio, which reads,
writes and works. Reviews are welcome — I would rather have this checked
carefully than have it merged quickly.


deeja:
Tested your branch on my 5RH Pro with GPS and is all working well. The push pull comparison came out the same using the provided test img.

steinbrueckri:
I played around with the driver a bit more this morning and found an issue, which is fixed by cd8fbf4561db853a86f086f1ac884eeeb6cf0cc6.

What’s still missing is TX blocking. The driver is already working without it, but I’ll add that later as well. I’m still a noob when it comes to HAM radio, and since I don’t have a license, I’m not allowed to transmit. So having TX properly disabled is pretty important for me too 😄

steinbrueckri:
Update on TX blocking, which I said I would add later: it turns out this
firmware does not support it.

The CPS stores "Launch banned" in bit 5 of channel byte 19 — comparing an
image before and after enabling it shows that byte going 0x00 -> 0x20 on
exactly the five channels I banned and staying 0x00 on the other seventeen.
But the radio ignores it: a channel uploaded from chirp with the bit set
still keys up, before and after a power cycle, and there is no matching
entry in the radio's own channel menu.

So instead of exposing it as duplex "off" and promising something that does
not work, I documented where the bit lives in a comment. See the
"Deliberately not implemented" section in the description above.

Worth noting for anyone with the same use case: the airband (108-136 MHz)
is outside every TX range the radio reports, so it cannot transmit there
anyway, regardless of this setting.

Also added since the last update: airband receive with AM, an alias so the
UV-5RM Plus GPS is findable under its own name, a beta prompt, and lower
case in channel names. Everything is verified against ten images now, and
airband traffic is audible on the radio.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,1109 @@
...
+    ("busylock", "Busy channel lockout"),
+    ("beep", "Key beep"),
+    ("keylock", "Key lock"),
+    ("autokey", "Auto key lock"),
+    ("dispdir", "Display direction"),
+    ("enhance", "Enhanced function"),
+]
+
+
+def _freq_bcd_to_hz(raw):
```
kk7ds: As noted on the original, isn't this just `bbcd`?

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,1109 @@
...
+    dath = int(dath)
+    datl = int(datl)
+    if dath & 0x80:
+        # DCS
+        if dath == 0xFF:
+            return (None, None, None)
+        pol = "R" if (dath & 0xC0) == 0xC0 else "N"
+        val = ((dath & 0x07) << 8) | datl
+        code = int("%X" % val)  # hex digits as decimal: 0x023 -> 23
+        return ("DTCS", code, pol)
```
kk7ds: Again, as in the original, this appears to be implementing the kenwood-style tone encoding with bit math instead of proper types and re-use of existing routines. Please either use that or explain why you can't.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,1109 @@
...
+        return (raw >> 8) & 0xFF, raw & 0xFF
+    if mode == "DTCS":
+        raw = int("%d" % val, 16)  # 23 -> 0x023
+        high = (raw >> 8) & 0x07
+        high |= 0xC0 if pol == "R" else 0x80
+        return high, raw & 0xFF
+    return 0x00, 0x00
+
+
+def _handshake(radio, is_write=False):
```
kk7ds, line 239: As I mentioned before, I'm very sure all the special timing in this routine is not necessary and also very sure it will not work reliably across platforms. Please don't do it this way. This tends to come from people looking at serial syscall dumps and assume that mimicing all the same behavior is required but it's almost never for the better.

Also, please don't choose a random XOR key. I'm not sure if the OEMs are doing this for "security" or what but it does nothing other than make serial traces hard to read. Unless the radio requires it (I'm sure it doesn't) then use an XOR key of zero to make it transparent, or at the very least, a static value if it can't be zero.

kk7ds (changes requested):
Thanks for picking this up. It doesn't look to me like any of my feedback from the previous attempt was addressed. I've only read down as far as the lowest comment I made, but I figure it's best to get things addressed before I spend a lot more time on it.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
deeja: @kk7ds  - fair enough comments on the code, but if you're talking about the device to device message timing, it's important. I was having issues with the handshake and couldn't figure it out until I matched specific periods between calls.

kk7ds: I've been doing this for a very long time and I can tell you that your short sleeps followed by attempts to probe the incoming buffer before doing reads are not going to be stable across platforms and are wholly unnecessary. In all cases I've ever encountered, a longer timeout and known-length reads are *always* workable and far more repeatable on multiple platforms and hardware. I've fielded enough bugs from people reporting that drivers written exactly like this don't work on a new platform, or with real serial ports, or when the system is heavily loaded, etc, etc. All of those were synthesized from port captures that include all the serial line discipline that the OS's own timeout routines are doing which ended up informing the code to look just like this, and it always ends the same way.

I believe and appreciate that you may have struggled to get that working and I can provide some tips, but I'm not okay merging the code in this state.

kk7ds: For what it's worth, I just hacked this up as an example. This massively-simpler code does the "handshake" with my 5RM Plus GPS just fine with the rest of the code working as-is, with the serial line timeout set to 500ms. My radio doesn't appear to switch to 115200, so I'm not sure what that part of the code is about. If one of the other models does (or maybe a later firmware?) then it'd be easy to add back in here without much drama.
```python
def _handshake(radio, is_write=False):
    port = radio.pipe

    port.write(bytes(T_INFO))
    resp = port.read(1)
    if not resp:
        raise errors.RadioNoResponse('Radio did not respond')
    elif resp != b'A':
        raise errors.RadioError('Radio did not respond as expected')

    # The radio is designed to communicate with XOR-obfuscated messages.
    # This is not enforced, so using an XOR key of 0x00 makes this
    # transparent and thus makes serial traces and debugging much easier.
    # The last byte of HEADER_SYNC as sent to the radio is what will be used
    # as the XOR key going forward, so just leave it as 0x00.
    header = bytearray(HEADER_SYNC)
    port.write(header)
    resp = port.read(1)
    if resp != b'A':
        raise errors.RadioError('Radio did not agree to program mode')

    password = b'\xFF' * 8
    port.write(password)
    resp = port.read(1)
    if resp != b'A':
        raise errors.RadioError('Radio did not ACK password')

    port.write(HEADER_INFO)
    model_info = port.read(16)
    LOG.info(f"Radio model: {model_info}")

    # Send direction (0x52='R' for read, 0x57='W' for write)
    port.write(is_write and b'W' or b'R')
    resp = port.read(1)
    if resp != b'A':
        raise errors.RadioError('Radio refused clone direction')

    # This is the "seed" that other routines will use. Return this for now
    # so the rest of the code works.
    return 0
```
The rest of the driver code needs to be written like this ^. Also, note that using a static XOR key of 0x00 works just fine and makes the serial traces human-readable. So much easier.

deeja: All good. I can test that code out on mine

deeja: @kk7ds  @steinbrueckri Changing out that handshake didn't work it seems. 
image

Regarding the switch up to the 115kbps, the code around this mostly came from the decompiled .Net code and USB traces that I took from the [UV-5RM Plus GPS software](https://www.baofengradio.com/pages/download)  and a lot of trial and error. 

If you want I can post up the .Net code.

kk7ds: And you switched the radio.pipe timeout to 500ms? The 100ms in the code will be too short. It's 100% reliable for me. I've tested it on all three platforms now.

A screenshot is not useful for debugging. A debug log and serial trace as generated by CHIRP would be, ideally attached to the bug. I can push the whole change on top of this if you want to test exactly what I'm running.

deeja: Firmware v2.0.09
Hardware v1.0.4
[chirp-trace-28c_x506.txt](https://github.com/user-attachments/files/31162032/chirp-trace-28c_x506.txt)

```
# Serial trace SerialTrace<id=0x7b70e30ab2e0, open=True>(port='/dev/ttyUSB1', baudrate=19200, bytesize=8, parity='N', stopbits=1, timeout=0.25, xonxoff=False, rtscts=False, dsrdtr=False) started at 2026-08-18T14:03:05.115948
  0.001 W 000: 00 00 00 00 00 00 00 00 00 00 00 00 ff ff ff ff   ................
  0.102 R 000: fe                                                ................
# Trace ended at 2026-08-18T14:03:05.358903
```

Set to .5 timeout. The log suggests it's 0.25 so maybe it is. The reply from the radio comes back earlier than that at 0.102 with 0xFE. 


Here's the first part of the successful pull down using the code from this PR
```
# Serial trace SerialTrace<id=0x75f71dc1b460, open=True>(port='/dev/ttyUSB1', baudrate=115200, bytesize=8, parity='N', stopbits=1, timeout=0.1, xonxoff=False, rtscts=False, dsrdtr=False) started at 2026-08-18T14:13:55.472415
  0.000 W 000: 50 52 4f 47 52 41 4d 6e                           PROGRAMn........
  0.101 R 000: 2f                                                /...............
  0.102 W 000: 91 91 91 91 91 91 91 91                           ................
  0.112 R 000: 2f                                                /...............
  0.112 W 000: 27 20 28 21 3c 23 2f 3a 27 21 20                  '.(!<#/:'!......
  0.212 R 000: 2c 28 43 5b 3c 26 91 91 6e 91 91 91 91 91 91 91   ,(C[<&..n.......
  0.213 W 000: 3c                                                <...............
  0.274 R 000: 2f                                                /...............
  0.275 W 000: 3c 6e 6e 6e                                       <nnn............
  0.286 R 000: 39 6e 6e 6e 3d 37 5f 2b 2c 2b 6e 6e 6e 6e 6e 6e   9nnn=7_+,+nnnnnn
  0.286 R 016: 6e 6e                                             nn..............
  0.286 R 000: 6e 6e 6e 6e ee                                    nnnn............
  0.286 R 000: 6f 6e 6e                                          onn.............
  0.286 R 000: 6e 4e 6e                                          nNn.............
  0.286 R 000: 6e 6e                                             nn..............
  0.297 R 000: 4e 6e 6e 6e 48 6e 6e 6e 5b 6e 6e 6e 57 6e 6e 6e   NnnnHnnn[nnnWnnn
  0.297 R 016: 2e 6e 6e 6e 0e 6e 6e 0e 7d 6e 6e 2e 79 6e 6e 6e   .nnn.nn.}nn.ynnn
  0.297 R 032: 4c 6e 6e 6e 48 6e 6e 6e 5b 6e 6e 6e 57 6e 6e 6e   LnnnHnnn[nnnWnnn
  0.297 R 048: 2e 6e 6e 6e 26 18 5c 31 5e 31 5e 57 91 18 5f 31   .nnn&.\1^1^W.._1
  0.297 R 064: 5e 31 5a 91 91 5c 5e 5c 58 40 5e 58 40 5e 5d 4e   ^1Z..\^\X@^X@^]N
  0.297 R 080: 5f 58 54 5b 5b 6e 4e 6e da 96 91 91 6c 6e 6e 6e   _XT[[nNn....lnnn
  0.297 R 096: 6e 6e 6e 6e 6e 6e 7e 16 2d 6e                     nnnnnn..-n......
  0.297 R 000: fe                                                ................
```

@steinbrueckri  I'll leave this with you now. Good luck

kk7ds: Yeah, your "this PR" trace is incomplete because the code here closes and re-opens the port (which won't be allowed either) so it throws away the portion of the trace that happens before that. It does imply, however, that your radio is switching rates, which mine is not. The first part of your trace with the new code also shows that your radio is doing something that the existing code just ignores as long as it happens quick enough, which is also bad and non-deterministic.

Thanks for your help @deeja.

@steinbrueckri let me know if your radio seems to be switching rate and I'll post some code to try to handle that accordingly and log the unknown bit between the initial announce and the rate switch so we know what to expect there. However, I'd rather do log dumps via the bug tracker on https://chirpmyradio.com/issues/11834 for visibility to other testers and for posterity.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Yes. It's `lbcd` rather than `bbcd` here — byte 0 holds the least significant digits — but the point stands. Both the channel frequencies and the band table in the header are `lbcd` now, in units of 10 Hz, so the two conversion helpers are gone and it's just `int(_mem.rx_freq) * 10`.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: You're right, it is exactly the encoding `tdh8.py` already implements. The fields are `bbcd` now and use that form: CTCSS as freq*10, DCS normal as 8000+code, DCS inverted as 12000+code, 16665 for unset.

Confirmed against a downloaded image — the raw values read through `bbcd` come out as 885, 1000, 8023 and 12023 for 88.5, 100.0, D023N and D023R, so it lines up exactly. All 640 memories decode identically to before and a download/upload round trip is unchanged byte for byte.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: My radio does switch rate, but I think the switch is an artifact rather than part of the protocol.

Announce at 19200 on a 5RH Pro with GPS, firmware v2_0_09:

```
  0.004 W 000: 00 00 00 00 00 00 00 00 00 00 00 00 ff ff ff ff
  0.209 R 000: ff
```

I get `0xFF` where @deeja gets `0xFE`, and that difference is the interesting part — a real reply byte would be the same on both radios. An `A` sent at 115200 and sampled at 19200 decodes as `0xFF` or `0xFE` depending on where the sampling point falls: the whole byte lasts 87µs while one bit at 19200 is 52µs, so the receiver samples its first data bit when the line is already back to idle. Both values fit that, and each radio reproduces its own value on every run.

So there is no announce byte and no mid-session rate switch, just a port on the wrong rate. Opening at 115200 gives a clean handshake immediately:

```
  0.002 W  T_INFO
  0.105 R  A
  0.106 W  PROGRAM.
  0.208 R  A
  0.212 W  INFORMATION
  0.217 R  BF-5RH..........
  0.217 W  R
  0.272 R  A
```

I've applied your version of the handshake, plus:

- blocking reads of known length throughout, no `in_waiting` polling and no sleeps
- no `close()`/`open()`; `_announce()` tries 115200 and falls back to 19200, so both variants work with the same code and neither needs the reply byte interpreted
- XOR key of `0x00`, which is what makes the trace above readable in the first place

On my radio the twelve blocks now arrive in a steady 0.406–0.416s each, and the downloaded image is byte for byte identical to what the previous code produced. No timing tricks needed anywhere, exactly as you said.

The boot-image path got the same treatment for consistency, but I have no safe way to exercise it, so that part is structural only and untested.

Thanks @deeja for the testing. Full traces are on #11834, including the one that fails at 19200.

kk7ds: Nice @steinbrueckri that all makes a lot of sense, of course. I'll continue code discussions here, testing discussions on the bug (thanks for uploading the traces there).

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+    MODEL = "UV-5RM Plus GPS"
+
+
+@directory.register
+class BaofengUV5RHRadio(chirp_common.CloneModeRadio):
+    """Baofeng 5RH Pro / UV-5RM Plus with GPS, v2 firmware."""
+    VENDOR = "Baofeng"
+    MODEL = "5RH Pro with GPS (v2)"
+    BAUD_RATE = 115200
+    ALIASES = [UV5RMPlusGPSAlias]
```
kk7ds: This is deprecated:

https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Aliases

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+        except errors.RadioError:
+            raise
+        except Exception as e:
+            raise errors.RadioError("Failed to upload to radio: %s" % e)
+
+    def _rebuild_zones(self):
+        """Regenerate the zone tables so every in-use channel is reachable.
+
+        A channel is only reachable on the radio if its ID sits in a zone
+        (DataProtocol.cs:1828). Chirp has no zone concept, so channels are
```
kk7ds: That's sort of true, but it would be better to have this driver behave like the other zoned radios that CHIRP does support in the proper way. See `tk8180` or `tk280` for example. The former is highly dynamic (to an annoying degree) and the latter is a bit more static.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+T_INFO = bytearray(16)
+for _i in range(12, 16):
+    T_INFO[_i] = 0xFF
+
+DATA_LEN = 49152
+BLOCK_SIZE = 4096
+CHN_SIZE = 48
+CHN_MAX = 640
+
+# Zone layout (DataProtocol.cs ConvertZone/ZoneConvert). The radio navigates
```
kk7ds: What are these `.cs` references? I worry that they're referring to disassembled proprietary code. If so, they're both both helpful going forward and also possibly inviting concern from the owner of that code. Please remove.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+    """Same hardware and clone protocol, sold under a different name."""
+    VENDOR = "Baofeng"
+    MODEL = "UV-5RM Plus GPS"
+
+
+@directory.register
+class BaofengUV5RHRadio(chirp_common.CloneModeRadio):
+    """Baofeng 5RH Pro / UV-5RM Plus with GPS, v2 firmware."""
+    VENDOR = "Baofeng"
+    MODEL = "5RH Pro with GPS (v2)"
```
kk7ds: This is not really an appropriate model string. Is there a 5RH Pro without GPS? My "5RM Plus" shows just that on the badge, even though it has GPS. And if there's no v1 (especially if not supported in the tree) then we really don't need to put that all in there.

For model for both, "5RM Pro" and "UV-5RM Plus" are sufficient. If we really need to distinguish, then you can add stuff to `VARIANT` but please keep it relevant and short as possible.

chirp/drivers/baofeng_5rh_pro.py
```diff
@@ -0,0 +1,937 @@
...
+    VENDOR = "Baofeng"
+    MODEL = "UV-5RM Plus GPS"
+
+
+@directory.register
+class BaofengUV5RHRadio(chirp_common.CloneModeRadio):
+    """Baofeng 5RH Pro / UV-5RM Plus with GPS, v2 firmware."""
+    VENDOR = "Baofeng"
+    MODEL = "5RH Pro with GPS (v2)"
+    BAUD_RATE = 115200
```
kk7ds, line 600: Is it possible that the 5RM Plus uses 19200 and the 5RH Pro uses 115200? If so, it'd be better to just have a subclass that overrides the `BAUD_RATE` and that would also prevent one selection from working with the other radio.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+
+    # Both per-channel bitmaps (valid_flags at 0x7A20, scan_flags at 0x81A0)
+    # use a clear bit for "on", matching CPS ConvertChnValidFlg
+    # (DataProtocol.cs:883).
+    @staticmethod
+    def _get_flag(bitmap, index):
+        return ((int(bitmap[index // 8]) >> (index % 8)) & 1) == 0
+
+    @staticmethod
+    def _set_flag(bitmap, index, on):
```
kk7ds: Please use `lbit` or `bbit` accordingly in your memory format.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+        nb = name.encode('gb2312')
+    except UnicodeEncodeError:
+        nb = name.encode('ascii', errors='ignore')
+    return nb[:length].ljust(length, b'\x00')
+
+
+def _decode_tone(val):
+    """Decode a sub-audio field into a (mode, value, polarity) spec.
+
+    The field is the usual BCD encoding, with the top nibble flagging DCS:
```
kk7ds: Again, this looks like `kenwood_tone` exactly. Unless there is a reason, please use the utility class provided there and avoid doing your own math here.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+
+        # tx_tone is the encode (transmit) field, rx_tone the decode one
+        chirp_common.split_tone_decode(
+            mem,
+            _decode_tone(_mem.tx_tone),
+            _decode_tone(_mem.rx_tone),
+        )
+
+        # power[7:6]: 2 == High, 0 == Low; wideth[5]: set == 25K == FM
+        mem.power = POWER_LEVELS[1] if (int(_mem.flags1) >> 6) >= 2 \
```
kk7ds: Please enumerate these bitfields in the memory format and don't do your own math.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+
+        mem.name = _decode_name(_mem.get_raw(asbytes=True)[32:48]).rstrip()
+
+        return mem
+
+    def set_memory(self, mem):
+        _mem = self._memobj.memory[mem.number - 1]
+
+        if mem.empty:
+            _mem.set_raw(b'\xff' * CHN_SIZE)
```
kk7ds: `fill_raw()` will do this for you.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
```diff
@@ -0,0 +1,937 @@
...
+
+        dly = (int(_s.vox_dly) - 10) // 5
+        _list("vox_dly", "VOX delay (s)", VOXDLY_LIST, dly)
+
+        bt = int(_s.blight_time)
+        bt_idx = 0 if bt < 5 else min(bt - 4, len(BLIGHTTIME_LIST) - 1)
+        _list("blight_time", "Backlight time (s)", BLIGHTTIME_LIST, bt_idx)
+
+        for key, name in _BOOL_SETTINGS:
+            rs = RadioSetting(
```
kk7ds: Please use MemSetting whenever possible. It looks to me like you can probably use it for almost all of this (with RadioSettingValueMap) and probably have basically nothing in `set_settings()`.

kk7ds (commented):
Thanks for your work, getting closer!

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Removed, all seven of them. The comments now describe the behaviour instead of
pointing at where it was observed.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: It is exactly that, you were right. Now using `KenwoodToneModel` with
`dcs_base=0x8000, pol_mask=0x4000, tone_flag=0x0000` and base 16 for both,
which dropped the two conversion functions entirely.

The fields are `u16 rxtone`/`txtone` as the model expects. Verified against a
downloaded image: memories decode the same and a round trip is byte for byte
identical. One difference worth mentioning: a TSQL channel now reports `rtone`
as what the radio actually stores for transmit, where `split_tone_decode()`
used to leave it at the default.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Done. `power`, `wideth`, `offsetdir`, `sqtype` and friends are named fields in
the memory format now, so setting them is an assignment.

A side benefit I had not expected: `freqinvert` and `talkaround` used to
survive a write only because of a hand-written `& 0x03` mask. Now they are
just fields nobody touches.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Done.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Done, `lbit` for both per-channel bitmaps, which removed the two shift-and-mask
helpers. Their inverted sense is in the field names now (`chn_unused`,
`chn_unscanned`) rather than in a comment.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Done, and you were right that it takes almost all of it. `set_settings()` is
now a six line loop.

The three settings whose stored byte is not the option index use
`RadioSettingValueMap`. The only thing left in `set_settings()` is the radio
name, because its GB2312 encoding cannot be expressed as a MemSetting.

Verified against a downloaded image: all 25 settings read back the same,
writing them unchanged leaves the image byte for byte identical, and changing
vox level, vox delay and backlight time stores the same bytes as before.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Done. It is a base class with one registered subclass per badge now. See the
model naming thread for which names and why.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: You are right that the old string was wrong, but it turns out the answer is
not one name. These are all the same device:

| Source            | Name on it              |
|-------------------|-------------------------|
| my type plate     | UV-5RH Plus             |
| Amazon DE listing | UV-5RH Plus GPS         |
| @deeja's unit     | 5RH Pro with GPS        |
| your unit         | 5RM Plus                |
| Baofeng's site    | UV-5RH Pro GPS, UV-5RH  |

The clincher is a unit on Reddit labelled "UV-5RM Plus" whose FCC ID is
2AJGM-5RHPRO [1]. An FCC ID is per hardware, so the sticker is the only thing
that varies. My own unit has no FCC label at all, being an EU one.

So the driver now registers three: "5RH Pro", "UV-5RH Plus", and "UV-5RM Plus"
with VARIANT "v2", so owners can pick whatever is printed on their radio. The
VARIANT is there because baofeng_uv17Pro.py already registers "UV-5RM Plus" —
which also answers your question about whether a v1 exists: it does, in the
tree, and that is what it is.

GPS is deliberately gone from all of them. The driver exposes no GPS settings,
and the receiver does not appear to work on these units — mine does not get a
lock, and owners on Reddit report the same [1]. Promising it in the model list
seemed worse than leaving it out.

I should be clear that the naming still has unknowns, and I am new to this
corner of the hobby, so treat the three names as a proposal rather than a
conclusion:

- I do not know whether the list is complete. It is three tested radios plus
  what Baofeng's own site lists, and Baofeng may well ship a fourth sticker
  tomorrow.
- "UV-5RH" without the Plus is deliberately *not* registered, because the
  video in the BAUD_RATE thread shows one being programmed with the 5RM
  driver, i.e. on the older protocol. But I have no way to tell whether every
  "UV-5RH" is the older one, so someone with that badge and v2 firmware would
  currently not find an entry.
- Whether any of these ever shipped without GPS hardware, I cannot say.

You know this ecosystem far better than I do, so if you would rather have
fewer entries, different names, or a different split, just say which and I
will change it.

[1] https://www.reddit.com/r/Baofeng/comments/1j4pc87/uv5rm_plus_gps_help/

chirp/drivers/baofeng_5rh_pro.py
steinbrueckri: I do not think the rate follows the badge, and I would rather not tie it to
one until we know. Three data points so far:

  my UV-5RH Plus        115200
  @deeja's 5RH Pro      115200
  your 5RM Plus          19200

Given the FCC ID above says these are one device, the badge cannot be what
decides it. There is also a YouTube video of a UV-5RH being programmed with
the 5RM driver [2], i.e. the same badge on a radio speaking the *older*
protocol. That points at firmware rather than model.

Which is the thing I cannot check: what firmware does yours report? The driver
now logs it when an image is loaded, since the image carries the same strings
the CPS shows under Device Information:

  Radio firmware v2_0_09, hardware v1_0_4, programmed 2026.08.13 21:26

If it really is per model I will happily split it. But if it is per firmware,
a hardcoded BAUD_RATE would lock out everyone whose radio has the other one,
in both directions, and the fallback costs a second in the worst case. Erring
on that side seemed safer while it is still a guess.

[2] https://www.youtube.com/watch?v=6cRLyE02V4g

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
steinbrueckri: Done, zones are sub-devices now, and the zone table is described in the memory
format instead of hand-computed offsets.

One deliberate simplification versus tk280: zones own a fixed slice of the
channel list, so channel n of zone z is channel z*64+n. The radio's limits are
10 zones of 64 with a hard total of 640, which is exactly the channel list, so
a fixed mapping is lossless here and there is nothing to allocate or compact.
It also keeps channel numbers stable, so existing images stay valid.

The tradeoff is that a channel cannot appear in two zones. If you would rather
have the dynamic tk280 arrangement despite that, say so and I will do it.

Tested on the radio: a download fills the zone tabs, reorganising channels
across them works, and the result navigates correctly on the radio afterwards.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
kk7ds: Okay, so I really really try to avoid adding "v2" versions in a way that they will show up in the download box. I hate that these manufacturers can't be more organized with their rollouts and especially hate when they call a radio "foo V2" because "foo" is all over the google searches and they want to capitalize on the name recognition. That's a huge problem with the now-ubiquitous UV-5R, although appears to be not as bad here.

The way to resolve that is to have the v2 variant in the code, but use `@directory.detected_by` linkage from the original and let the code do the programming mode exchange. I haven't looked closely yet, but does that seem possible? Even if the parent tries to interrogate as a v1 and then returns v2 to let it try, it should work. It's massively better for the users than listing v2, v3, v4 and having them try each (since they generally have no idea what they have). Doing so also teaches them to try random other driver profiles (which too often partially work due to all the incestuous copying of firmware that happens between these).

So I have the problematic v2 and I also have some uv17pro-based things, so I could take a crack at the detection and probably have reasonable confidence of it working, if you need me to.

chirp/drivers/baofeng_5rh_pro.py
kk7ds: I'll try to get the firmware off mine soon.

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
kk7ds: Er, can the radio show one memory in two "zones"? If so, that behaves more like banks than zones. Perhaps I steered you incorrectly. The previous version of this radio had groups or banks or zones, but they were 100% static, "memory 50 was in bank 1 because it was >1 and < 10". People got annoyed because they assumed banks meant what they mean on nicer radios, which means you can put memories into banks at will.

If this radio can do that (either 1:1 or 1:many) then it should be using BankModel and not sub-devices for the reason you describe. If you're not familiar with this, try opening the `Yaesu_VX-7.img` image from `tests/` and see the "banks" tab. That implements a 1:many bank assignment. Is that how the radio behaves?

chirp/drivers/baofeng_5rh_pro_with_gps_v2.py
kk7ds: Not a side benefit, but one major reason for doing it this way. CHIRP used to be all hand-rolled bit math in every driver and it was full of bugs like that :)

machshev:
Thanks, I've just managed to read my "RD-55 Pro" with this driver, no change at all.
Not sure if the memory map is exactly the same or not, but it's either a pure rebadge or a variant.

**Update**: while not an exhaustive tests the channels do seem to have been programmed up correctly, and I'm able to get into the local repeaters. Locally I've added support for the APRS settings which I can upstream once this PR is in, if you don't get to it :)

machshev:
If it's useful, this seems to be the "RD-55 Pro" APRS memory map:

```
#seekto 0x9E00;
struct {
  char dest[6];        // 0
  u8 dest_ssid;        // 6
  u8 dest_flag;        // 7
  char callsign[6];    // 8   own callsign, space padded
  u8 ssid;             // 14
  u8 unknown_0f;       // 15
  u8 unknown_10[4];    // 16
  char symbol_table;   // 20  "/" for the primary APRS symbol table
  char symbol_code;    // 21
  u8 unknown_16[2];    // 22
  struct {
    char call[6];
    u8 ssid;
    u8 flag;
  } path[8];           // 24  digipeater path, blank slots are all spaces
} aprs;
```

It would be good to know if this matches the "5RH Pro" to see if "RD-55 Pro" is just an identical clone or will need to be a variant with differences.

steinbrueckri:
> If it's useful, this seems to be the "RD-55 Pro" APRS memory map:
> 
> ```
> #seekto 0x9E00;
> struct {
>   char dest[6];        // 0
>   u8 dest_ssid;        // 6
>   u8 dest_flag;        // 7
>   char callsign[6];    // 8   own callsign, space padded
>   u8 ssid;             // 14
>   u8 unknown_0f;       // 15
>   u8 unknown_10[4];    // 16
>   char symbol_table;   // 20  "/" for the primary APRS symbol table
>   char symbol_code;    // 21
>   u8 unknown_16[2];    // 22
>   struct {
>     char call[6];
>     u8 ssid;
>     u8 flag;
>   } path[8];           // 24  digipeater path, blank slots are all spaces
> } aprs;
> ```
> 
> It would be good to know if this matches the "5RH Pro" to see if "RD-55 Pro" is just an identical clone or will need to be a variant with differences.

Thanks, that's useful — the map matches the 5RH Pro exactly.

Decoding your struct against the test image in this PR
(`tests/images/Baofeng_5RH_Pro.img`) at 0x9E00 gives:

```
    dest        'APAT81'  ssid 7  flag 0xff
    callsign    'BG6LKK'  ssid 8
    symbol      '/' '['
    path[0]     'WIDE1'   ssid 1
    path[1..7]  all spaces
```

Every field lands on a sensible value at the offset you describe, including
the space-padded blank slots and the primary symbol table, so the APRS block
looks identical on both radios. The only part I'd still question is `dest`:
ssid 7 / flag 0xff doesn't obviously read as an SSID, so those two bytes may
well be something else.

A few things immediately after the structure, in case they help:

```
    0x9E58  1e 00 00 00 c8 00 00 00  — two 32-bit values, 30 and 200
    0x9EAC  seven copies of 00 85 cf 00 = 13600000 in units of 10 Hz =
            136.0 MHz, presumably APRS frequency slots left at the factory
            default. Note this reads as a plain little-endian integer, not
            the lbcd the rest of the memory uses for frequencies.
    0x9EC8  'APAT81' ssid 8 flag 1 followed by eight more blank path slots,
            i.e. what looks like a second copy of the same structure.
```

That only tells us the APRS block is the same in this one image, though — it
doesn't say whether the RD-55 Pro is a straight rebadge. The version strings
at 0x0050 and the band tables at 0x0010/0x0030 from an RD-55 Pro image would
be far more conclusive on that.

I'd rather keep APRS out of this PR, since it's already large and marked
beta, but this is a good starting point for a follow-up.

PR #1627, opened 2026-08-18
chirp_common: Change comparison operator from < to <= to allow inclus…

TrimbleSoftware:
chirp_common: Change comparison operator from < to <= to allow inclusive upper frequency range in Invalid Entry error messages.

Back in January 2026, (see PR https://github.com/kk7ds/chirp/pull/1469) the frequency input error message in chirp_common.py was modified to display the frequency range allowed by the driver when an out-of-bounds value was entered for a frequency.

The message would display the frequency range (or ranges)  numbers and imply that the **upper limit** was inclusive of the values shown. Because the comparison operator was **<** (less than) a frequency value exactly at the driver's upper limit could not be entered and the message would indicate that it should be allowed. By changing the comparison operator to **<=** (less than or equal to) for the upper limit an exact value for the drivers **upper limit** can be entered without causing the error message. This is  the exact behavior already used on the range's **lower limit** comparison.

The change allows the range to be checked for values that are **within** the range specified in the driver and eliminates confusing warning messages and allow input of the full frequencies specified in the driver code.

This clip shows the conflicting error message when inputting 520.0 Mhz as the frequency:

image

With this change that frequency would be allowed and no error would be displayed.


kk7ds:
I know this is probably what it should have been from the beginning, and also probably what most driver writers expect it to be, but this is kindof a big change that could affect a bunch of drivers. Here's one example I just found in a quick grep:

https://github.com/kk7ds/chirp/blob/master/chirp/drivers/retevis_ra25.py#L523

I'd want to survey the others more closely before we do this, and also probably add a unit test to assert the expected behavior for the future.

TrimbleSoftware:
> I know this is probably what it should have been from the beginning, and also probably what most driver writers expect it to be, but this is kindof a big change that could affect a bunch of drivers. Here's one example I just found in a quick grep:
> 
> https://github.com/kk7ds/chirp/blob/master/chirp/drivers/retevis_ra25.py#L523
> 
> I'd want to survey the others more closely before we do this, and also probably add a unit test to assert the expected behavior for the future.

There are a few drivers that use that _001_ work-around, but there are far more that use the actual frequency at _000_. When the allowable frequency range was not displayed in the error message it was not that noticeable, but now that it is displayed it makes you wonder why it reports an invalid value when it is within the ranger that is displayed.

kk7ds:
> There are a few drivers that use that 001 work-around, but there are far more that use the actual frequency at 000. When the allowable frequency range was not displayed in the error message it was not that noticeable, but now that it is displayed it makes you wonder why it reports an invalid value when it is within the ranger that is displayed.

Yeah, I understand the concern about the seeming discrepancy in the display. We could solve that by adjusting the end point of the range we display (end - 1) or the way we display it (low < high instead of low-high).

My concern is that there's 15 years of precedent with that being how the ranges are enforced that making a change like this subtly changes what the drivers can expect as input filtering. I think in almost all cases the driver authors intended what you expect (inclusive endpoint) it's just really hard to know that for sure in all the cases we'll be changing with one sweep of the pen. Any of them that actually intend not to allow 174.0 MHz when they list 170 as the upper limit would be wrong here. I'd just like to check a few representative samples of radios I care about to see what their actual upper dial limit is, as well as find others that may have compensated for it before we sweep that pen.

Here's another example:

> rf.valid_bands = [(108000000, 135997500)

That looks to me like the driver (kguv9dplus) author was intending to prevent actually dialing to 136MHz because the radio can't do it (and because airband tuning steps are strange). Other drivers that are similar (even one in the same file) do not make that distinction on the upper limit, which I assume is relevant since the author is the same person.

TrimbleSoftware:
> > There are a few drivers that use that 001 work-around, but there are far more that use the actual frequency at 000. When the allowable frequency range was not displayed in the error message it was not that noticeable, but now that it is displayed it makes you wonder why it reports an invalid value when it is within the ranger that is displayed.
> 
> Yeah, I understand the concern about the seeming discrepancy in the display. We could solve that by adjusting the end point of the range we display (end - 1) or the way we display it (low < high instead of low-high).
> 
> My concern is that there's 15 years of precedent with that being how the ranges are enforced that making a change like this subtly changes what the drivers can expect as input filtering. I think in almost all cases the driver authors intended what you expect (inclusive endpoint) it's just really hard to know that for sure in all the cases we'll be changing with one sweep of the pen. Any of them that actually intend not to allow 174.0 MHz when they list 170 as the upper limit would be wrong here. I'd just like to check a few representative samples of radios I care about to see what their actual upper dial limit is, as well as find others that may have compensated for it before we sweep that pen.
> 
> Here's another example:
> 
> > rf.valid_bands = [(108000000, 135997500)
> 
> That looks to me like the driver (kguv9dplus) author was intending to prevent actually dialing to 136MHz because the radio can't do it (and because airband tuning steps are strange). Other drivers that are similar (even one in the same file) do not make that distinction on the upper limit, which I assume is relevant since the author is the same person.

This is an interesting problem and I'll leave it up to you how to proceeded. At least now your aware of it...

Interesting enough, 12 yeas ago, the **low end** comparison operator was changed to be inclusive, but not the upper in this commit:

https://github.com/kk7ds/chirp/commit/fa2f735828d277223f37f216138f72912bb7e2d3

May have been the same dilemma then too?

kk7ds:
> Interesting enough, 12 yeas ago, the **low end** comparison operator was changed to be inclusive, but not the upper in this commit:
> 
> [fa2f735](https://github.com/kk7ds/chirp/commit/fa2f735828d277223f37f216138f72912bb7e2d3)
> 
> May have been the same dilemma then too?

We had much the same conversation at the time and settled on low<=freq<high as the proper thing. Testing some radios reminded me that I did the same thing back then.

So, all of the Kenwood and Yaesu radios I tested do not actually allow dialing past `high - $tuning_step`. Meaning, if 5kHz is the smallest tuning step you can get to, and the range stops at 174MHz, then the highest you can dial is 173.995MHz before it wraps. A random modern Baofeng radio I grabbed to test behaves the same. The icom radios I tested do allow you to dial all the way to the end, meaning they are fully inclusive of the edges of their declared limits. However, none of the radios will let me transmit on 148MHz, which is expected since a center frequency with 5kHz deviation would end up with up to 2.5kHz outside the allowed limits.

I think what this tells us is that probably a ton of the declared band limits in the drivers don't faithfully represent what the radios can actually do and so there is no neat solution here. If we allow setting inclusive of the upper limit and that causes a memory to be corrupted, not display, or not function properly, then that's also bad and likely to generate reports. That said, what we're currently displaying is very similar to what is in the spec sheets for these radios which don't actually dial to their stated end point.

I would propose maybe making this "error" dialog one of the ones that has a "more info" button which takes people to the website to explain why this is complicated and imperfect and hopefully that will be satisfactory for people. We could also change the display to "low~high" (which is how Kenwood lists it in their specs) or less-good "low-~high" or less-less good "low-<high" all of which implies some amount of variation (for the people that pay close enough attention).

We could also make each range show `high - min(rf.tuning_steps)` as the end which will not be right for some radios, but may be more likely satisfying for the user. Of course, someone will file a bug saying "but my radio can actually do 174.0MHz!". I guess in that case we could set the end range for that driver to 174.005MHz. We could also update every single driver and add `min(rf.tuning_steps)` to the end of each range and change the code to be inclusive as you have here.

We could also update the template driver, the driver notes, and maybe docs in RadioFeatures to try to encourage future driver authors to pick this value carefully.

PR #1628, opened 2026-08-18
Add Kenwood NX-800

kk7ds:
- **tk8180: Split dynamic zone behavior into mixin for reuse**
- **tk8180: Split OST settings behavior into mixin for reuse**
- **Store traces in ./traces if debugging and present**
- **Add verbose_repr() for bitwise structDE**
- **Fix browser edited notification getting eaten**
- **Add Kenwood NX-800 driver**


kk7ds:
Failure is false positive

PR #1632, opened 2026-08-29
wxui: Fix AttributeError opening Settings on wxPython >= 4.3

tpoppe:
## Summary
wxPython 4.3 replaced the `wx.propgrid.PG_ACTION_EDIT` and
`PG_ACTION_NEXT_PROPERTY` integer constants with the `PGKeyboardAction`
enum (`PGKeyboardAction.Edit` / `.NextProperty`). Code that still
referenced the old constants raised:

```
AttributeError: module 'wx.propgrid' has no attribute 'PG_ACTION_EDIT'
```

whenever the Settings tab (or a memory's property dialog) was opened.

## Fix
Detect which API is available at import time and use whichever
exists, so this works against both wxPython < 4.3 and >= 4.3.

## Testing
Reproduced on macOS (arm64) with wxPython 4.3.1, which is what pip
installs there since 4.2.0 (pinned in `requirements.txt`) has no
prebuilt wheel for that platform/Python combination. Confirmed the
Settings tab opens without error after the fix, using a Baofeng
UV-5G Plus (Radioddity UV-5G Plus driver) as the test radio.

### DIFF
--- a/chirp/wxui/common.py
+++ b/chirp/wxui/common.py
@@ -40,6 +40,26 @@ from chirp.wxui import radiothread
 LOG = logging.getLogger(__name__)
 CONF = config.get()
 
+
+def pg_action_edit():
+    # wxPython >= 4.3 replaced the PG_ACTION_* integer constants with
+    # the PGKeyboardAction enum. Look this up lazily (i.e. not at
+    # module import time) since by the time a settings grid is
+    # actually built, something will have already imported
+    # wx.propgrid for real; doing it ourselves at import time with an
+    # explicit `import wx.propgrid` breaks tests that stub out `wx`
+    # wholesale with a Mock.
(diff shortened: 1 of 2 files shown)


tpoppe:
Updated: the first version broke \`tests/unit/test_wxui_radiothread.py\`, which stubs \`wx\` out entirely with a \`Mock\` to run headless. My original fix added an eager, module-scope \`import wx.propgrid\` statement, which fails against that stub (a \`Mock\` isn't a real package). Reworked it to resolve the constant lazily via two small helper functions instead, matching how the rest of this code already relied on \`wx.propgrid\` being attached by the time it's actually used, rather than importing it directly.

Full local unit test suite now passes (682 passed, 12 skipped), and the Settings tab still opens correctly against wxPython 4.3.1.

PR #1637, opened 2026-09-09
Add ABBREE AR-7600 driver

gkrash:
Fixes #12525

## Summary

New driver for the ABBREE AR-7600 mobile radio, reverse-engineered from
scratch (no prior CHIRP support, no existing protocol documentation)
via decompiling the vendor CPS software and a live USB capture.

## What's supported

- Frequency, CTCSS/DCS tone (cross-mode included), name, power
  (High/Low), bandwidth (FM/NFM), skip
- A substantial set of vendor-specific per-channel extras exposed via
  the standard "Extra" tab: Busy Channel Lockout, CtDegree, VOX,
  Random Frequency (roam), Scramble, Compander, Hop/Jump Password,
  Signal Type, Dig Serve, PTT ID, Speaker Unmute, Scramble Frequency,
  Signal Dec/Enc
- The radio's two independent 250-channel groups (Group A / Group B),
  each with its own VFO memory, exposed as separate CHIRP sub-devices
  to match the radio's own layout (same pattern as `bj9900.py`'s
  Left/Right VFO)

Both download and upload are validated end-to-end against real
hardware, including surviving a power cycle and matching the vendor
CPS's own display after a round-trip.

## Testing

- `tests/images/Abbree_AR-7600.img` included, generated from a real
  radio via the driver's own `sync_in()`
- Full `tests/test_drivers.py` suite passes for both sub-devices
  (66 passed, 18 skipped for undeclared features, 4 non-strict xpass)
- `pep8`/`flake8`/`mypy` all clean

## Notes for reviewers

- General (radio-wide) settings are not yet exposed via
  `get_settings()` -- only per-channel extras are implemented so far.
  `has_settings = False` reflects this honestly; a follow-up PR may
  add these once mapped.
- A handful of vendor features are round-tripped but not decoded/
  exposed: a "Ctdcs Code" and "Spe Code" per-channel field, and a
  DMR-style contact/username table this analog radio's shared OEM
  platform still carries under the hood. Both are preserved as-is on
  write, just not surfaced in the UI.
- Happy to share more of the reverse-engineering process (protocol
  framing, byte-level citations into the decompiled vendor source) if
  useful for review -- just ask.


chirp/drivers/abbree_ar7600.py
```diff
@@ -0,0 +1,1119 @@
...
+
+NO_TONE = 0xFFFF
+
+_MEM_FORMAT = """
+struct {
+  lbcd rxfreq[4];
+  lbcd txfreq[4];
+  ul16 rxtone;
+  ul16 txtone;
+  u8   flags12;
```
kk7ds: These need to be broken out into bitfields. You're bypassing all the goodness of bitwise. I'm guessing because your LLM agent just brute-forced everything, but this needs to be done properly.

chirp/drivers/abbree_ar7600.py
```diff
@@ -0,0 +1,1119 @@
...
+    RadioSettingValueBoolean, RadioSettingValueList
+
+LOG = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Wire-scramble tables, extracted verbatim from the vendor CPS
+# (modPcCom.InitCodeListRandom, RwRadio.exe V1.1). Do not hand-edit.
+# ---------------------------------------------------------------------------
+
+SECRET_CODE_DATA = [
```
kk7ds, line 68: I am highly (highly) suspicious of this kind of thing. Are you sure this isn't programmatic (like the XOR stuff everyone copied from Kenwood)?

chirp/drivers/abbree_ar7600.py
```diff
@@ -0,0 +1,1119 @@
...
+    183, 219, 112, 176, 252, 14, 27, 109, 184, 148,
+    233, 0, 68, 159, 147, 140, 204, 225, 25, 85,
+    70, 155, 97, 154, 238, 55, 99, 146, 6, 126,
+    135, 165, 205, 153, 117, 58, 24, 73, 95, 128,
+    253, 118, 250, 104, 255, 40, 193, 35, 13, 229,
+    106, 217, 43, 181, 2, 29, 50, 10, 47, 239,
+    48, 169, 114, 44, 8, 49,
+]
+
+SECRET_RANDOM = [
```
kk7ds, line 107: This is clearly not random and looks a lot like brute forcing of something that is actually much simpler. I really don't want to know much about what you decompiled, but it's easy to be fooled into thinking this is really a lookup table when in fact it's something far simpler. I'm really loathe to merge something that looks (to me, no offense) like we really don't understand what's going on under the covers.

gkrash: No offense taken! Happy to dig into these more. Am no stranger to python, but the detail on how this works is definitely something I appreciate the pushback on. Will dig into your comments and amend.

That said - there’s absolutely a fair amount of brute forcing going on - I don’t have much background in this are and its more than a bit of fumbling around in a dark room with a crappy flashlight trying to get a useful CPS for this radio (the vendor one is "doesn't support csv import or copy/paste" levels of awful)

chirp/drivers/abbree_ar7600.py
gkrash: Done. Bytes 12–15 are now real bitwise bitfields (u8 vox:1, unknown13:1, power:1, wide:1, ...) and get_memory / set_memory read and write _mem.bcl, _mem.pttid, _mem.scan, etc. directly. All the FLAG*_MASK / FLAG*_SHIFT constants are gone. Bits no CPS control was found to touch are named unknownNN_* and preserved on write.

Verified against the previous version with a differential test on the real image: identical decode across 2,000 random flag-byte patterns and identical raw bytes across 1,000 random writes. Net ~70 lines smaller.

chirp/drivers/abbree_ar7600.py
gkrash: It doesn't appear to be - went back to the decompiled CPS and it's a hardcoded table, doesn't appear to be a generator. Full detail in the thread below (line 111), since that's the substantive one.

chirp/drivers/abbree_ar7600.py
gkrash: Hopefully this isn't too much, but the source routine in the CPS is real basic, I had it pasted in here prior, but it occurred to me that might be a problem so I edited.  It really is just defining 3 three array literals and returning true. No loop, no seed, no key schedule. "Random" in the name just means "looks random" - nothing is computed at runtime.

Happy to paste what's in there if you would like to see it, just trying to be safe about it.

also checked the numbers themselves for hidden structure:

- all three are permutations of 0–255
- gbytSecretCodeData and gbytSecretSeqList are exact inverse permutations of each other, an 8-bit S-box and its decode half
- no fit to an LCG orbit, affine-in-index (a*i+b) mod 256, the AES S-box / GF(2^8) inverse, RC4-KSA with trivial keys, bit-reversal, or any simple pairwise composition of the three

So the protocol layer really appears to be a plain table-driven byte substitution plus additive whitening (SECRET_RANDOM as a position-dependent mask), with vendor-chosen arbitrary tables. 

They appear to be permutations, and one of the three was redundant. Given the S-box/inverse relationship I've dropped the second block of 256 magic numbers: SECRET_SEQ_LIST is now derived as the inverse of SECRET_CODE_DATA at import, with an assert that doubles as a permutation sanity check. The remaining two literal tables now carry a comment stating what they are and citing the decompiled function above. Happy to restructure further if you'd still rather see it done differently. Also happy to keep digging if you feel like there's gold in here somewhere that I'm just missing because I'm green.

PR #1643, opened 2026-09-13
retevis_rt18: Experimental Retevis RT18 driver

RafaelGCPP:
## Summary
- Adds an experimental, reverse-engineered driver for the Retevis RT18.
- No vendor documentation or protocol reference was available, so the
  clone protocol and 16-channel memory layout were derived entirely
  from passive USB capture analysis (Wireshark + USBPcap) of the
  vendor Windows CPS talking to a real unit over its programming
  cable.
- Channel memory (frequency, tones, power, narrow/wide, scan add,
  busy lock, scramble, compander, spec code) is confirmed against
  captured traffic and against a real unit.
- Squelch tone/DCS code, wide/narrowband, and squelch level were
  reprogrammed via CHIRP and confirmed correct by reading the radio
  back with the vendor's own Windows CPS.
- The rest of the general settings screen (timeout timer, VOX, beep,
  scan mode, battery save, etc.) is inherited unchanged from the
  Radtel T18 family (`chirp/drivers/radtel_t18.py`) and has not been
  individually tested against a real RT18 yet -- flagged as such in
  the module docstring.
- A 6-digit programming password setting is included, confirmed
  against a real "set password" capture, and also tested end-to-end
  on a real unit (reprogrammed via CHIRP, read back and confirmed
  correct with the vendor CPS). It can be changed or cleared from
  CHIRP.

## Test plan
- [x] `tests/images/Retevis_RT18.img` added (synthetic, generated
      programmatically -- contains no real user data) so the driver
      is covered by the standard per-driver suite.
- [x] `pytest tests/test_drivers.py -k Retevis_RT18` passes (edges,
      brute force, clone-error-handling, settings, features, copy-all).
- [x] `pytest tests/unit` passes (705 passed, 14 skipped).
- [x] Full read/write clone cycle verified against a real RT18 unit
      over its serial programming cable: reprogrammed the password,
      squelch tone/DCS code, wide/narrowband, and squelch level from
      CHIRP, then read the radio back with the vendor's own Windows
      CPS and confirmed the changes took effect and matched what was
      set in CHIRP. The remaining general settings (timeout timer,
      VOX, beep, scan mode, battery save) were not individually
      exercised this way yet.


chirp/drivers/retevis_rt18.py
```diff
@@ -0,0 +1,248 @@
...
+
+    for addr in range(0, radio._memsize, radio.BLOCK_SIZE):
+        status.cur = addr + radio.BLOCK_SIZE
+        radio.status_fn(status)
+
+        block = t18._t18_read_block(radio, addr, radio.BLOCK_SIZE)
+        data += block
+
+        LOG.debug("Address: %04x" % addr)
+        LOG.debug(util.hexprint(block))
```
kk7ds: Please don't do this, use annotation in the serial trace instead:

https://chirpmyradio.com/projects/chirp/wiki/DevelopersDriverNotes#Communication-logging

chirp/drivers/retevis_rt18.py
```diff
@@ -0,0 +1,248 @@
...
+
+def _rt18_apply_password(setting, radio):
+    digits = [int(c) for c in str(setting.value).strip()]
+    raw = bytes(digits) + _PASSWORD_BLANK[len(digits):]
+    radio._mmap.set(_PASSWORD_ADDR, raw)
+
+
+def _rt18_enter_programming_mode(radio):
+    """Identical handshake to the Radtel T18 family (magic + ident +
+    ack), plus one extra step this RT18 firmware does before the normal
```
kk7ds: So are you overriding everything here when all you need to do is override this "enter programming mode" bit? Can we maybe move that to the class so you can override it in your subclass without having to do everything else?

chirp/drivers/retevis_rt18.py
```diff
@@ -0,0 +1,248 @@
...
+            status.cur = addr + radio.BLOCK_SIZE
+            radio.status_fn(status)
+            t18._t18_write_block(radio, addr, radio.BLOCK_SIZE)
+
+    t18._t18_exit_programming_mode(radio)
+
+
+@directory.register
+class RT18Radio(t18.T18Radio):
+    """Retevis RT18 (EXPERIMENTAL, reverse-engineered -- see module
```
kk7ds: We've had _some_ documentation for only a very small handful of radios. You don't need to disclaim that.. I'd be more interested in calling out if you _had_ documentation :)

kk7ds (commented):
Couple small things, otherwise looks okay, thanks.

chirp/drivers/retevis_rt18.py
RafaelGCPP: Ok. Claude added during one of our debug sessions (there was something funky in the comms) and left it there. I am adjusting this.

chirp/drivers/retevis_rt18.py
RafaelGCPP: It makes sense. I was afraid it would break something :) 

I tested a little more and it does everything the Retevis CPS does, only better!

chirp/drivers/retevis_rt18.py
RafaelGCPP: I was trying to avoid messing with Radtel T18 driver, but I found a minimal intrusion way to do so.

chirp/drivers/retevis_rt18.py
RafaelGCPP: Fixed — switched to `self.pipe.log()` to annotate the serial trace instead of dumping the raw bytes into the debug log. Thanks for the pointer to the wiki.

chirp/drivers/retevis_rt18.py
RafaelGCPP: Good call — added an overridable `_enter_programming_mode()` hook on `T18Radio` in radtel_t18.py, so `RT18Radio` now only overrides that one step instead of duplicating `do_download`/`do_upload`/`sync_in`/`sync_out`. Verified all 897 tests across the T18 family still pass.

chirp/drivers/retevis_rt18.py
RafaelGCPP: Removed that line, thanks.

chirp/drivers/retevis_rt18.py
RafaelGCPP: So I created an _enter_programming_mode(radio) method in Radtel's class so I can override it in Retevi's one. That way  I didn't had to overwrite everything. I also moved the password functions inside RT18Radio class.

chirp/drivers/retevis_rt18.py
RafaelGCPP: Removed the dumb text :-D

chirp/drivers/retevis_rt18.py
RafaelGCPP: Fixed. Now I use the serial trace for logging.

RafaelGCPP:
Let me know if it needs some more refinements. It is my first contribution to an open source project!

kk7ds:
(squashed the intermediary commit and rebased)

PR #1644, opened 2026-09-13
wxui: Fix broken keyboard focus under wxPython 4.3 / wxWidgets 3.3

RafaelGCPP:
## Summary
- Under wxPython 4.3+ (wxWidgets 3.3+) on Windows (MSW), keyboard
  focus in CHIRP's UI could silently break -- clicking/tabbing into
  property grid rows (e.g. the Settings tab) or newly-opened editor
  tabs would stop responding, with `'SetFocus' failed with error
  0x00000057 (The parameter is incorrect.)` printed to the console.
- Root cause: `wx.CallAfter()`'s semantics changed in wxPython 4.3 --
  it now runs "after the current and pending event handlers have
  completed" instead of getting its own turn of the event loop (as
  under 4.2/wxWidgets 3.2). Several places in the wx UI relied on two
  sibling `wx.CallAfter()` calls landing in *separate* turns (one
  tearing down a `wx.ProgressDialog` or modal dialog, the next moving
  keyboard focus elsewhere) -- under 4.3 they can now land in the same
  turn, so a `SetFocus()` gets issued while the dialog is still alive
  (just hidden, not yet `Destroy()`ed). This is a real, long-standing,
  never-fixed-upstream wxWidgets bug on MSW -- see
  https://github.com/wxWidgets/wxWidgets/issues/24290, whose title is
  literally "SetFocus when selecting a notebook page while Progress is
  active".
- Fixed at the three call sites this surfaced at:
  - `chirp/wxui/common.py`: `ChirpEditor.stop_wait_dialog()` now
    destroys its `wx.ProgressDialog` *synchronously* instead of via
    `wx.CallAfter()`, and takes an optional `then=` callback that is
    only scheduled via `wx.CallLater()` (a real timer tick, immune to
    the semantics change) once the dialog is actually gone.
    `start_wait_dialog()` was adjusted to tolerate the dialog being
    destroyed before its own queued `Show()` runs.
  - `chirp/wxui/settingsedit.py`: closes the wait dialog and
    `wx.SafeYield()`s *before* building the settings `Treebook`'s
    pages, since `wx.Treebook.AddPage()` implicitly focuses the first
    page it adds (like `wx.Notebook` does) and was racing the
    still-alive dialog.
  - `chirp/wxui/main.py`: `_menu_download()` and `_do_network_query()`
    now let their modal clone/query dialog actually `Destroy()` (by
    exiting its `with` block) and `wx.SafeYield()` before building the
    new `ChirpEditorSet`, which immediately focuses its first memory
    grid.

## Test plan
- [x] `pytest tests/unit` passes (705 passed, 14 skipped).
- [x] `tools/cpep8.py` and `mypy` clean on the changed files.
- [x] Manually verified on a real Windows machine running Python
      3.14 / wxPython 4.3.1 (wxWidgets 3.3.3): before this change,
      opening a radio's Settings tab reliably broke keyboard/mouse
      focus in the property grid, logging the `SetFocus` error to the
      console; after this change it no longer reproduces across
      repeated download -> Settings tab cycles.


kk7ds:
Where is the commit that says this changed in wxpython 4.3? The referenced bug is reported on wxpython 4.2. I ask because the settings loading dialog has always been a bit glitchy on Windows, so it may have always been this way.

Can I ask you to please file bugs on the chirp site for each of the unique issues here? Specifically capturing the fatal error message you say is possible for posterity, but also for detailed reproduction steps for the focus issues?

https://chirpmyradio.com/projects/chirp/issues/new

Then reference those in the commit here. I'll try to repro and test this on a windows machine, which isn't on 4.3.x. Of course, any fix we merge here has to work on the actual version of wx that gets bundled in the build, as well as all the other platforms...

RafaelGCPP:
Created Bug #12594 in issues.

I was eager to make chirp work for my radio, as I had set a password and needed to clear it. The reverse engineering showed it to be possible, but when everything was working, and I could read the password from the radio, I found the settings grid was not enabling for editing. Bummer... 

I could do it by downgrading to Python 3.10, but since it is on the verge of being phased  out I decided it was worth trying fixing it for Python 3.14. This is the reason for this PR: allowing chirp to run on a newer Python!

kk7ds:
Just an update here, on Linux, wxpython 4.2.1, wx 3.2.4, python 3.12 this causes the settings pane to show up empty until you force a window refresh by navigating to another app and back. Still digging into why...

RafaelGCPP:
I'll take a look

kk7ds:
Okay it only happens in wayland-native mode, which has other quirks. I recently flipped compat mode off by default and have been thinking about putting it back. So I'll do that and then we can merge this.

Also tested on MSW py3.10 with wxpython 4.2.x which is what the formal builds run.

RafaelGCPP:
Curiously, I couldn’t reproduce the empty Settings pane on Linux Mint 22.3 / Cinnamon Wayland / Python 3.12.3, using wxPython 4.2.1 built locally against the system wxWidgets 3.2.4. The Settings Treebook populated immediately after downloading from a UV-5R. I also tested the X11 compatibility backend, which behaved normally.

Under Wayland I do see GTK popup warnings when opening dropdown controls, but they do not prevent the Settings pane from rendering. Everything looks absolutely normal.

RafaelGCPP:
Oh, wait! I forgot to apply my  own patch... (Facepalm)

RafaelGCPP:
Nevermind... it still works fine.

kk7ds:
It only manifested for me on radios with many subdevices (where the subdevices, settings, etc) are options on the right hand side treeview, not along the top. But as I mentioned I'll just merge this once my patch to default to compat mode lands (#1647). Also, we need to reference the bug you opened in your commit message per the guidelines. I can do that for you this time though.

PR #1646, opened 2026-09-17
ftm300d: more settings

cqcallaw:
Display and TX/RX menus fully mapped. Config menu mostly mapped.

Remaining Config menu options seem low value (set date/time) or tedious to map (mic button programming).

I made some organizational and stylist choices that I'm happy to discuss or debate:

1. Some settings apply to the selected/active band on the radio, which doesn't map well to a PC programming interface. Such settings have subgroups where it seemed reasonable, and label differentiation where a subgroup seemed cluttered.
2. Mixed case for settings groups but all capital letters for setting values. A middle ground between screaming caps everywhere and mixed case everywhere. A bit shouty but hopefully still legible.

I wasn't able to test round-trip on every one of these settings, but the ones I tested work well on a live radio. As before, ably assisted by Grok Build.


chirp/drivers/ftm300d.py
```diff
@@ -489,6 +859,70 @@ def get_prompts(cls):
             "press the DIAL knob.\n")
         return rp
 
+    def _append_mode(self, parent, suffix, title):
+        group = RadioSettingGroup("band_%s" % suffix, title)
+        parent.append(group)
+        bw = int(getattr(self._memobj, "fm_bandwidth_%s" % suffix))
+        if bw not in (0, 1):
```
kk7ds: Why are you doing all these validations? When properly done (with current_index as you are) the settings will show invalid out of range settings in yellow and let the user pick the right value to correct. I wouldn't expect that to be a big problem in a yaesu radio, but it's common for chinese junk to have corrupted values straight from the factory. If you need to do a lot of these it makes me worry there's something else going on.

cqcallaw: 100% clanker-generated defensive coding. I wasn't aware of the UI-level validation.

I had Grok strip out all the value clamps and was able to load, download, and upload captures without issue, so I'd say the clamps can be safely dropped. Pushed an update to this PR.

tests/unit/test_yaesu_clone.py
```diff
@@ -458,26 +464,179 @@ def test_settings_defaults_and_roundtrip(self):
...
+
+        settings = src.get_settings()
+        for setting in _iter_settings(settings):
+            if setting.get_name() == "aprs_ssid":
+                setting.value = ""
+        src.set_settings(settings)
+        packed = src.get_mmap().get_packed()
+        self.assertEqual(0xCA, packed[0x50E])
+        values = _setting_map(src.get_settings())
+        self.assertEqual("", values["aprs_ssid"])
```
kk7ds, line 639: I'm not quite sure what all is going on here.. there seems to be a lot of forced setting of values in settings and then asserting the low-level bytes...I guess? Seems like this is unmaintainable without automation and maybe not that useful really. Is this maybe agent slop to "verify" stuff but doesn't really add much value?

tests/unit/test_yaesu_clone.py
```diff
@@ -514,6 +692,206 @@ def test_settings_from_live_dumps(self):
...
+            ('Yaesu_FTM-300DR_20260916_apo=1.5h.img',
+             {'apo': '1.5 HOUR'}),
+            ('Yaesu_FTM-300DR_20260916_apo=2.0h.img',
+             {'apo': '2.0 HOUR'}),
+            ('Yaesu_FTM-300DR_20260916_apo=3.0h.img',
+             {'apo': '3.0 HOUR'}),
+            ('Yaesu_FTM-300DR_20260916_apo=4.0h.img',
+             {'apo': '4.0 HOUR'}),
+            ('Yaesu_FTM-300DR_20260916_apo=12.0h.img',
+             {'apo': '12.0 HOUR'}),
```
kk7ds: Also this.. what are these weird image names? I didn't ask in the previous round, but lots more added here makes me wonder.

chirp/drivers/ftm300d.py
```diff
@@ -574,24 +1129,61 @@ def get_settings(self):
...
                 west, mag = timezone_to_fields(half)
                 self._memobj.tz_west = west
                 self._memobj.tz_mag = mag
-            elif name == "lcd_brightness":
-                self._memobj.lcd_brightness = BRIGHTNESS_TO_RADIO[value]
+            elif name == "beep":
+                on, high = beep_to_fields(value)
+                self._memobj.beep_on = on
+                if high is not None:
+                    self._memobj.beep_high = high
```
kk7ds, line 1157: Feels like RadioSettingValueMap should be workable here, but I guess the reason we need all this bespoke handling is because these values are split into two fields (enable and magnitude)? I was going to say, just mapping them to values with the right value should work, but they're far away in memory so, maybe not so much.

kk7ds (commented):
couple more questions

tests/unit/test_yaesu_clone.py
cqcallaw: This is indeed auto-generated, but seemed useful to me. It verifies that a given setting value maps to the expected byte value(s). I'm not sure what automation would be required since it's reading a local image; this test seems to run fine locally and in CI.

If it seems low value, just say the word and I'll excise it.

tests/unit/test_yaesu_clone.py
cqcallaw: These are the names of image captures I made to map out various settings. Change one setting to a value, capture, change to the next, and so on. Grok added tests that read in the image and verifies the mapped value matches expectations. They are skipped if the images don't exist, which would be anywhere that isn't my dev machine.

Committing those images in the master branch doesn't make sense, especially since they carry memories are only useful in my geo, but they're a nice confidence builder for local testing. Shall I strip them out?

chirp/drivers/ftm300d.py
cqcallaw: Yeah, beep is laid out weirdly, but I did double-check it with live uploads and the mapping is correct. I don't see a way to simplify this code, but I'm open to suggestions.

tests/unit/test_yaesu_clone.py
kk7ds: Well, what I mean is that the assertions are all just pretty opaque and it seems unlikely that someone would actually make a change to one side and not just do the "run the test, see which assertion fails, change it to whatever the new value is".

Anyway, fair enough, all good.

tests/unit/test_yaesu_clone.py
kk7ds: Okay I figured. Yeah, I think strip these out. I do plenty of this sort of thing myself during dev but don't persist it.

kk7ds:
If you can strip out the image stuff and then squash this down to a single commit I'll merge for tomorrow. Thanks!

tests/unit/test_yaesu_clone.py
cqcallaw: Done

cqcallaw:
> If you can strip out the image stuff and then squash this down to a single commit I'll merge for tomorrow. Thanks!

Excellent, thanks! Squash, rebased, and pushed.

PR #1650, opened 2026-09-19
retevis_rt18: Retevis RT18 driver refactoring to avoid code replication.

RafaelGCPP:
As suggested, I am bringing some changes to the RadtelT18 class, so we avoid code replication in Retevis RT18 code.


RafaelGCPP:
I'll rearrange this PR.

